PR #186, opened 2023-01-13
Conditional running of transactions and methods

tilk:
This PR adds a new functionality to the transaction subsystem. It allows execution of methods/transactions to be dependent on the execution of other methods/transactions, and/or additional conditions. This extension allows to implement two new library modules, both with FIFO-like interfaces:

* `Connect` exports two methods `read` and `write` which can only be run together in the same clock cycle. Data is passed from one to the other combinationally, without any buffering.
* `Buffer` contains a single cell. Unlike a FIFO with depth 1, it's possible to do `read` and `write` in the same clock cycle.

The implementation may cause more locking than needed in some conditions, as dependencies are resolved after transaction scheduling. In certain situations, some dependencies could be handled by the scheduler, but this can be left for later.

The main problem with this PR is that I have no idea if it is really correct. For simple use cases the tests I wrote find no issues, but this is not a proof that no issues exist.

Tasks left to do:

- [ ] Don't generate condition formulas when there is no conditions for the transaction.
- [ ] Verify that the implementation is actually correct (?!?)
- [ ] Write docstrings for the new functionality.


lekcyjna123:
@tilk I suppose that we can close this PR now as #347 have been merged.

tilk:
> @tilk I suppose that we can close this PR now as #347 have been merged.

Archived on my own fork, closing.

PR #189, opened 2023-01-15
TestbenchIO: new syntax suggestion for calls

Arusekk:
While reviewing #187 I found that it could be simplified if we used a slightly different syntax for the TestbenchIO calls, leveraging Python keyword arguments.
While at it, this adds type annotations to some related values previously being Any and therefore not checked (results of [`getattr()`](https://docs.python.org/dev/library/functions.html#getattr)).

In order to ease the hassle for PRs currently in progress utilizing the old syntax, I prepared a set of `sed` commands that should do the job:
```sh
sed -i -e  '/call\(_init\|_try\)\?({/s/"\(\w*\)": /\1=/g' \
       -e 's/call\(_init\|_try\)\?({\(.*\)})/call\1(\2)/' \
       $(grep -rl '\.call\(_init\|_try\)\?({')

sed -i -e 's/call\(_init\|_try\)\?(\(\w\+\))/call\1(**\2)/' \
       $(grep -rl '\.call\(_init\|_try\)\?(\w\+)')
```

The first commit is the result of running them, and the second is some actual changes.  Because of how I did it, it does not stand on itself and needs the other commit for any tests to work at all.  We might want to squash them if we want bisecting to be stable, but let's keep that for now to help reviewers. :-)


tilk (approved):
Looks good. Complementary change to mock syntax would be nice, but it's better to make the change in a separate PR.

tilk:
#217 introduced a new test.

lekcyjna123 (approved):

Issue #107, opened 2022-10-25
Functional unit for division (M extension)


tilk:
To support the M extension, we would need to have a functional unit that processes divisions. Division implementation can be multicycle non-pipelined.

PR #196, opened 2023-01-17
[#107] Division unit

wkkuna:
Currently: Very silly division based on repeated subtractions (just so I could both experience amaranth & coreblocks structure myself).

Remaining: Implementation of a sensible division algorithm (the structure is mostly there I guess)

Some resources I found and will most likely use:
https://www.controlpaths.com/2022/06/20/division-algorithms-in-fpga/
(mentioned in issue) https://yuhei1-horibe.medium.com/designing-divider-213fbd32beb2
(sounds interesting but unfortunately it's not available on sci hub) https://dl.acm.org/doi/10.1145/3502492

<sub> Please don't be harsh on me I'm very self conscious of my coding skills </sub>

### DIFF
--- a/coreblocks/fu/div_unit.py
+++ b/coreblocks/fu/div_unit.py
@@ -1,0 +1,110 @@
+from amaranth import *
+from enum import unique
+from coreblocks.params import GenParams, CommonLayouts, Funct3, FuncUnitLayouts
+from coreblocks.transactions.core import def_method
+from coreblocks.transactions.lib import FIFO
+from coreblocks.fu.division.long import LongDivider
+from coreblocks.utils import OneHotSwitch
+
+__all__ = ["DivUnit", "DivFn"]
+
+
+class DivFn(Signal):
+    @unique
+    class Fn(Signal):
+        DIV = 1 << 0
+        DIVU = 1 << 1
+        REM = 1 << 2
+        REMU = 1 << 3
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(MulFn.Fn, *args, **kwargs)
+
+
+class DivFnDecoder(Elaboratable):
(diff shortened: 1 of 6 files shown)


coreblocks/fu/div_unit.py
```diff
@@ -76,22 +80,36 @@ def _(arg):
...
 
             arg1, arg2 = arg.s1_val, Mux(arg.imm, arg.imm, arg.s2_val)
+            sgn1, sgn2 = arg1[sign_pos], arg2[sign_pos]
+            result_sign = sgn1 ^ sgn2
 
             dividend = Signal(self.gen.isa.xlen)
             divisor = Signal(self.gen.isa.xlen)
+            sign = Signal()
+
+            uarg1 = Mux(arg1[sign_pos], arg1 & (2**xlen - 1), arg1)
```
Kristopher38, line 90: `arg1 & (2**(xlen - 1) - 1)`

Kristopher38: actually after some further thought `(arg1 & (2**(xlen - 1) - 1)) + 1` which should be equivalent to `-arg1` (why doesn't the latter work?)

wkkuna: It does work... It's the test I've written - I've tried to pass signed arguments to the unit that can handle only unsigned ones (without removing the sign first - and this is what I was looking at within the traces)...

Thanks for help!
(Here's the line I'm talking about : https://github.com/kuznia-rdzeni/coreblocks/pull/196/files#diff-e17bd7a09a456a593bc0ec14b546af0f56ef770db6a14527b2180675b08ae90dR74-R75 - I've used the (min_int, max_int) range at that time) 
I'll fix it and upload it later today.

PR #201, opened 2023-01-18
Fixes and improvements to Amaranth type stubs

tilk:
Type stubs for `Elaboratable` had a problem which prevented some type checking and code completion from working. Fixing this required some further changes to make the codebase pass typing checks again.


stubs/amaranth/hdl/dsl.pyi
```diff
@@ -61,7 +63,7 @@ class _ModuleBuilderRoot:
     def __init__(self, builder, depth) -> None:
         ...
     
-    def __getattr__(self, name):
+    def __getattr__(self, name) -> NoReturn:
```
Arusekk, line 66: Is this right?

Arusekk (approved):
Looks good overall.

stubs/amaranth/hdl/dsl.pyi
tilk: Yes! The only function of `__getattr__` in `_ModuleBuilderRoot` is raising exceptions when the developer makes a mistake.

PR #209, opened 2023-01-24
Configurable FU used by `GenParams`

speederking07:
Goal of this PR is to create mechanism for easy configuration of FU used in core.

- [x] new field with FU configuration in `GenParams`
- [x] add core generation based on FU configuration
- [x] implement dependency injection (kinda implemented)
- [x] find new naming scheme for FU and Blocks
- [x] test commit method
- [x] add cross verification for isa and FU configuration


coreblocks/core.py
```diff
@@ -80,23 +78,84 @@ def elaborate(self, platform):
...
             free_rf_put=free_rf_fifo.write,
             rf_free=rf.free,
-            lsu_commit=self.lsu_unit.commit,
+            lsu_commit=self.func_blocks_unifier.commit,
         )
 
         return m
+
+
+# TODO: Move somewhere else
```
tilk, line 100: Maybe it's time to move it somewhere else?

coreblocks/core.py
```diff
@@ -80,23 +78,84 @@ def elaborate(self, platform):
...
+
+        self.branch_result_collector = None
+        match branch_result_methods:
+            case []:
+                raise Exception("CPU without jumps")
+            case [method]:
+                self.branch_result = method
+            case [*methods]:
+                self.branch_result_collector = br_collector = Collector(methods)
+                self.branch_result = br_collector.get_single
```
tilk, line 132: This looks like something to be generalized, as there is an almost identical block just a few lines below.

Also, the special case for length one makes me feel uneasy, but I understand it will simplify the connection diagram. Maybe some general way of doing this should be created, but this is not a priority.

coreblocks/fu/jumpbranch.py
```diff
@@ -9,7 +9,9 @@
 from coreblocks.params import *
 from coreblocks.utils import OneHotSwitch
 
-__all__ = ["JumpBranchFuncUnit"]
+__all__ = ["JumpBranchFuncUnit", "JumpComponent"]
+
+from coreblocks.utils.protocols import FuncUnit
```
tilk, line 14: Weird place for an import.

coreblocks/lsu/dummyLsu.py
```diff
@@ -1,10 +1,13 @@
...
+
 from coreblocks.transactions import Method, def_method, Transaction
 from coreblocks.params import *
 from coreblocks.peripherals.wishbone import WishboneMaster
 from coreblocks.utils import assign
 
-__all__ = ["LSUDummy"]
+__all__ = ["LSUDummy", "LSUBlockComponent"]
+
+from coreblocks.utils.protocols import FuncBlock
```
tilk, line 10: Weird place for an import.

coreblocks/lsu/dummyLsu.py
```diff
@@ -283,3 +286,13 @@ def _(arg):
             m.d.sync += reserved.eq(0)
 
         return m
+
+
+class LSUBlockComponent(BlockComponentParams):
+    def get_module(self, gen_params: GenParams, dependencies: ComponentDependencies) -> FuncBlock:
+        if dependencies.wishbone_bus is None:
+            raise Exception("LSU requires wishbone bus, but it was not provided as dependency")
```
tilk, line 294: There should be a way to avoid writing this. Maybe raise the exception from `ComponentDependencies`?

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,46 @@
...
+# extra constructor parameters of FuncBlock
+class ComponentDependencies:
+    def __init__(self, **dependencies):
+        self.dependencies = dependencies
+
+    def __getattr__(self, item):
+        if item in self.dependencies:
+            return self.dependencies[item]
+        else:
+            return None
```
tilk, line 26: This is kinda like `defaultdict`.

Also, `None` is in many (all?) places not really useful - if there is a dependency, you can't do anything with `None` other than raising an exception.

Also, type annotations.

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,46 @@
...
+    "BlockComponentParams",
+    "FunctionalComponentParams",
+    "optypes_supported",
+]
+
+
+# extra constructor parameters of FuncBlock
+class ComponentDependencies:
+    def __init__(self, **dependencies):
+        self.dependencies = dependencies
```
tilk, line 20: Is there a way to make dependencies typed? Maybe the keys should not be strings, but carry some type information with them?

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,46 @@
...
+        else:
+            return None
+
+
+class BlockComponentParams:
+    def get_module(self, gen_params: gp.GenParams, dependencies: ComponentDependencies) -> FuncBlock:
+        ...
+
+    def get_optypes(self) -> set[optypes.OpType]:
+        ...
```
tilk, line 34: This should be an ABC.

tilk:
I think we should focus on merging this, and leave improvements to the DI stuff for later PRs, especially if they would delay you.

speederking07:
I agree we need to generalize unifying methods for `RSBlocks` , however this commit is already large enough. So I will leave this improvement as well as adding comments to next PRs.

tilk:
> I agree we need to generalize unifying methods for `RSBlocks` , however this commit is already large enough. So I will leave this improvement as well as adding comments to next PRs.

Oh pleeeease, that's not a big deal. After merging #227, which makes `Collector` conform to the method combinator interface (which it was supposed to anyway), the required change will be trivial.

coreblocks/params/fu_params.py
tilk: This wasn't implemented.

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class FunctionalComponentParams:
+    @abstractmethod
+    def get_module(self, gen_params: gp.GenParams, dependencies: ComponentDependencies) -> FuncUnit:
+        ...
+
+    @abstractmethod
+    def get_optypes(self) -> set[optypes.OpType]:
+        ...
```
tilk, line 53: Should be ABCs, you forgot to do that.

coreblocks/stages/func_blocks_unifier.py
```diff
@@ -0,0 +1,73 @@
...
+
+        self.commit_product = None
+        match commit_methods:
+            case []:
+                raise Exception("CPU without side effects")
+            case [method]:
+                self.commit = method
+            case [*methods]:
+                self.commit_product = commit_product = MethodProduct(methods)
+                self.commit = commit_product.method
```
tilk, line 56: This copy-paste should be eliminated. Common code should be abstracted in a function.

coreblocks/lsu/dummyLsu.py
```diff
@@ -283,3 +287,11 @@ def _(arg):
             m.d.sync += reserved.eq(0)
 
         return m
+
+
+class LSUBlockComponent(BlockComponentParams):
+    def get_module(self, gen_params: GenParams, dependencies: ComponentDependencies) -> FuncBlock:
+        return LSUDummy(gen_params, dependencies.get(Key[WishboneMaster]("wishbone_bus")))
```
tilk, line 294: Construct the keys globally, and use their identity for indexing - instead you are basically losing the type safety.

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class Key(Generic[T]):
+    def __init__(self, name: str):
+        self.name = name
+
+
+# extra constructor parameters of FuncBlock
+class ComponentDependencies:
+    def __init__(self, **dependencies):
```
tilk, line 27: Kwargs are not appropriate for this use, as you lack the required type information there.

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,57 @@
...
+        self.dependencies = dependencies
+
+    def get(self, key: Key[T]) -> T:
+        if key.name not in self.dependencies:
```
tilk, line 31: `dependencies` should be indexed by keys, not their names.

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,57 @@
...
+    "FunctionalComponentParams",
+    "optypes_supported",
+]
+
+T = TypeVar("T")
+
+
+class Key(Generic[T]):
+    def __init__(self, name: str):
+        self.name = name
```
tilk, line 22: Because keys should be constructed once, you could get the names automatically from the definition. See, for example, how `Method` does this.

speederking07:
@tilk now it should look like your idea from last class. I am not the biggest fan of mutable argument for collecting outputs, but it look nice so I can accept this version.

coreblocks/params/fu_params.py
tilk: No, this is not enough. Please read the fine documentation.
https://docs.python.org/3/library/abc.html

coreblocks/params/genparams.py
```diff
@@ -22,21 +25,31 @@ class GenParams(DependentCache):
...
         self.isa = ISA(isa_str)
+        self.func_units_config = func_units_config
+
+        # Verification temporally disabled
+        # if not optypes_required_by_extensions(self.isa.extensions) <= optypes_supported(func_units_config):
+        #     raise Exception(f"Functional unit configuration fo not support all extension required by{isa_str}")
+
+        self.rs_entries = 1
+        for block in self.func_units_config:
+            import coreblocks.stages.rs_func_block as rs_block
```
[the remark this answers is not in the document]
speederking07: I agree it is ugly. but this import cannot be moved, because is causes circular import. I can try to replace this import with some `Protocol`.

coreblocks/stages/func_blocks_unifier.py
```diff
@@ -0,0 +1,54 @@
...
+            self.rs_blocks.append(block.get_module(gen_params=gen_params, connections=connections))
+
+        self.result_collector = Collector([block.get_result for block in self.rs_blocks])
+        self.get_result = self.result_collector.method
+
+        self.update_combiner = MethodProduct([block.update for block in self.rs_blocks])
+        self.update = self.update_combiner.method
+
+        self.unifiers = {}
+        self.extra_outputs = {}
```
tilk, line 28: Types are missing.

tilk:
I thought we agreed to leave complex changes for later. Now this PR will be stuck for a next week or two in CR 🤣

speederking07:
I also thought we will keep it simple, but since:
> Oh pleeeease, that's not a big deal. After merging https://github.com/kuznia-rdzeni/coreblocks/pull/227, which makes Collector conform to the method combinator interface (which it was supposed to anyway), the required change will be trivial.

Plus you wanted to have type safe dependencies, removed duplicated code fragment with method unifying, in that case it was as easy to fix every issue own as to create this more universal approach. All this requirements were screaming to me that I need to add this final piece to this code.

tilk:
I wanted to have the code duplication removed, that's all. For other comments I marked which ones are to be postponed. The design you proposed now is by no means perfect, and will need to go through another round or two of CR, while the previous one was satisfactory for now. Other team members need to take a look, too - but exams might delay that.

speederking07:
I would leave it as it currently is, I don't think that changes are that fundamental (it took me an hour to implement). But if you want I can go back to protocols, as I think type safe dependencies should stay. But I really see no point in abstracting function which will need to take a lot of parameters, only to be replaced with next PR one week later.

tilk:
I think that I'll let others decide on this one. We can always change this later, and we certainly don't want this change to land into pull request limbo.

test/fu/test_mul_unit.py
```diff
@@ -38,30 +38,40 @@ def compute_result(i1: int, i2: int, i_imm: int, pc: int, fn: MulFn.Fn, xlen: in
 }
 
 
+def gen_test_params(param):
```
lekcyjna123, line 41: Why you added that?

lekcyjna123 (changes requested):
This PR looks great! I am for merging it with the current state (so with dependency injection), because in my opinion it is very readable.

There are two points which should be fixed before merge:
- you changed GenParams interface, but didn't change benchmark scripts, so benchmark will fail on this PR
- now there are two definitions of OpType class, one of them should be removed

On the last meeting you have asked how we can connect (make an unifier) commit. I would propose two change commit interface so that it will return a bit 0 or 1. This will mean that commit ended with success or not. So an unifier will call all commit methods and will wait for the answer. If all will return OK, than unifier will return OK also.

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,89 @@
...
+class OutputKey(ABC):
+    @classmethod
+    @abstractmethod
+    def unifier(cls) -> type[Unifier] | None:
+        return None
+
+    @classmethod
+    @abstractmethod
+    def method_name(cls) -> str:
+        ...
```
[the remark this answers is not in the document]
speederking07: Hmm maybe that is a good idea. I will change it

coreblocks/params/genparams.py
```diff
@@ -22,21 +25,34 @@ class GenParams(DependentCache):
     def __init__(
         self,
         isa_str: str,
+        func_units_config: list[BlockComponentParams],
```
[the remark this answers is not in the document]
speederking07: Thx, I missed that

coreblocks/params/isa.py
```diff
@@ -7,7 +7,6 @@
     "Funct3",
     "Funct7",
     "Funct12",
-    "OpType",
```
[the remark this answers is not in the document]
speederking07: Thx, removed

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,89 @@
...
+class ComponentConnections:
+    def __init__(self):
+        self.dependencies = {}
+        self.outputs = {}
+
+    def set_dependency(self, key: type[DependencyKey[T]], dependency: T) -> ComponentConnections:
+        self.dependencies[key] = dependency
+        return self
+
+    def set_output(self, key: type[OutputKey], output: Method) -> ComponentConnections:
```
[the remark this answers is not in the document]
speederking07: I changed it to `register_method` but still not sure if it is a good name.

coreblocks/stages/func_blocks_unifier.py
```diff
@@ -0,0 +1,57 @@
...
+
+        self.unifiers: dict[str, Unifier] = {}
+        self.extra_outputs: dict[str, Method] = {}
+
+        for (key, output_methods) in connections.get_outputs().items():
+            unifier_type = key.unifier()
+            if unifier_type is not None and len(output_methods) > 1:
+                unifier = unifier_type(output_methods)
+                self.unifiers[key.method_name() + "_unifier"] = unifier
+                self.extra_outputs[key.method_name()] = unifier.method
```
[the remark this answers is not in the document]
speederking07: The whole idea is that method_name needs to be unique in another case we will have problems with method name shadowing. Key names are unique but then there will be an issue with ugly method names. If some knows how to enforce uniqueness here I would be grateful

test/fu/test_mul_unit.py
speederking07: Because in most tests we use `GenParams` only for constans like number of RS, but since I changed `GenParams` they no longer gets number of RS as parameter and it would be tedious to create mockup RS config for every test only to set number of RSes.

speederking07:
I will create this waiting for success unifier tomorrow.

tilk:
> I will create this waiting for success unifier tomorrow.

But do it in another PR, pretty please?

coreblocks/stages/func_blocks_unifier.py
```diff
@@ -0,0 +1,57 @@
...
+            elif len(output_methods) == 1:
+                self.extra_outputs[key.method_name()] = output_methods[0]
+
+    def __getattr__(self, item):
+        return self.extra_outputs[item]
```
tilk, line 43: I don't like this at all! The `Core` class has a very specific requirements (`get_result` and `update`), and references them directly. Here, everything is processed implicitly (even things that might not be required from the core, but were added in future development. This hurts clarity too. I would just handle `get_result` and `update` explicitly. The unifier construction code should be abstracted, so that it could be used in other places.

These changes would also fix the problem described by @lekcyjna123, as key manipulation magic would disappear.

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,89 @@
...
+            self.outputs[key] = [output]
+        return self
+
+    def get_dependency(self, key: type[DependencyKey[T]]) -> T:
+        if key not in self.dependencies:
+            raise Exception(f"Dependency {key} not provided")
+        return self.dependencies[key]
+
+    def get_methods(self) -> dict[type[MethodKey], list[Method]]:
+        return self.outputs
```
tilk, line 65: Smelly, and unnecessary, see comment on `FuncBlockUnifier`.

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,89 @@
...
+class ComponentConnections:
+    def __init__(self):
+        self.dependencies = {}
+        self.outputs = {}
+
+    def set_dependency(self, key: type[DependencyKey[T]], dependency: T) -> ComponentConnections:
+        self.dependencies[key] = dependency
+        return self
+
+    def register_method(self, key: type[MethodKey], output: Method) -> ComponentConnections:
```
tilk, line 49: It's ugly that `MethodKey` and `DependencyKey` are two different things. It would be best if the same logic could handle both.

Actually, there is a slight variation in logic which is still needed - some keys are associated with exactly one value, and some collect multiple values from different components. Maybe there should be two distinguished kinds of keys - singleton keys and multi-keys?

speederking07:
> But do it in another PR, pretty please?

Good point. Maybe it should go with changes to LSU commit? @lekcyjna123 what do you think?

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,89 @@
...
+    "optypes_supported",
+    "DependencyKey",
+    "MethodKey",
+]
+
+T = TypeVar("T")
+
+
+class DependencyKey(Generic[T], ABC):
+    pass
```
tilk, line 24: Consider the following changes:

* Make keys [dataclasses](https://docs.python.org/3/library/dataclasses.html) (with `frozen=True` to enable hashing).
* Use instances of keys in `ComponentConnections`, not key types.

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,89 @@
...
+class MethodKey(ABC):
+    @classmethod
+    @abstractmethod
+    def unifier(cls) -> type[Unifier] | None:
+        return None
+
+    @classmethod
+    @abstractmethod
+    def method_name(cls) -> str:
+        raise NotImplementedError()
```
tilk, line 36: I currently see two possible ways of handling the awkwardness of `MethodKey` (strange implicit dependencies, name collision issues, etc.):

* Get rid of it entirely, make a generic way of associating multiple values with one key instead, handle each method explicitly. (This solution has the upside of being simple, but it misses an abstraction opportunity.)
* Find a way to move the unifier construction logic closer to `MethodKey`, while somehow keeping the genericity of keys.

Because this PR lingers for so long, I currently prefer the option A. Otherwise, I feel this PR will remain in limbo even longer.

tilk (changes requested):

coreblocks/params/fu_params.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+# extra constructor parameters of FuncBlock
+class ComponentConnections:
+    def __init__(self):
+        self.dependencies = {}
+        self.outputs = {}
+
+    def set_dependency(self, key: type[DependencyKey[T]], dependency: T) -> ComponentConnections:
```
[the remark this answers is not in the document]
speederking07: In my opinion is the bast way to ensure type safety. Using types as keys ensures their global uniqueness, and makes sure no one will creates his own instance with incompatible type. And the only drawback is that it might take a bit longer time to create new key.

coreblocks/stages/func_blocks_unifier.py
speederking07: I think I don't get it. `get_result` and `update` are explicit in lines 24-28. I see a problem with that it is not obvious which method are available, but since they depends of `blocks`, I see no way how to fix it. Please, elaborate how I can make it more abstract.

coreblocks/params/fu_params.py
speederking07: But they have really different responsibility. Dependency can be anything, and I assume that they can be used multiple times without an issue. And MethodKeys, and handle only extra methods, and need to unify its interface. For me this is strong enough distinction.

coreblocks/params/fu_params.py
speederking07: I will get git of them entirely. Can I assume that every method of RS block and FU units with this same name has this same interface, and basically this same responsibility. If so, I could create a dictionary, with names as keys, and unifiers as values in order to handle it.

coreblocks/params/fu_params.py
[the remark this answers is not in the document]
tilk: > And the only drawback is that it might take a bit longer time to create new key.

The other drawback is, as I mentioned, that value-parametrized (in other words, dynamically created) keys are impossible. There is also the philosophical opposition: keys are classes without usable instances.

coreblocks/stages/func_blocks_unifier.py
tilk: Forgot to mention `branch_result` and `commit`.

Again, the problem is that `core` knows exactly what it wants, and can grab it by name. While here you have an indirection which might seem beneficial, but is actually useless: you have to handle these and only these methods that are needed in `core`. The dynamic `for` is actually harmful: if the required keys are missing, you will get an error in `core` (and not here!), and if there are extra keys (possibly needed elsewhere), they will be handled here regardless. You implicitly entangle `core`, this module, and method keys.

> Please, elaborate how I can make it more abstract.

"Abstracting code" is an elaborate way to say "make it a separate function or method somewhere".

coreblocks/params/fu_params.py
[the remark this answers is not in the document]
speederking07: I agree to that philosophical opposition it is kinda wierd to use classes this way. But I see no reason to dynamically create key. I thought that whole idea is that keys should be global, otherwise we will end up with `DependencyKey[A]('dependency1')` and `DependencyKey[B]('dependency1')` for this same dependency. Putting them as classes makes sure this same key will always have this same underlying type.

speederking07:
@tilk Could you take a look if new version is at least step in good direction?

test/fu/test_mul_unit.py
lekcyjna123: But this function has local scope and empty implementation and isn't used anywhere.

coreblocks/params/fu_params.py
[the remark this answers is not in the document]
lekcyjna123: > The other drawback is, as I mentioned, that value-parametrized (in other words, dynamically created) keys are impossible.

Value-parametrized keys in such scenario wouldn't be impossible. In Python we have functions `compile` and `exec` which allow us to add new code in runtime. So in case of keys defined by classes, this would allow us to define new class as a string of code and next execute it in our context to create new class. Ugly, but possible.

test/fu/test_mul_unit.py
speederking07: Oh, you're right. I must have created it by mistake.

lekcyjna123:
I have prepared my own PoC of keys system based on speederking07 solution. Please see to #235 .

tilk:
#235 now contains combined changes of @lekcyjna123 and mine. Please check them out.

tilk (approved):
Let's merge this!

lekcyjna123 (approved):

PR #217, opened 2023-01-28
Zicsr extension - Unit and Register

piotro888:
This PR provides functional part of Zicsr extension.
It creates `CSRUnit` - unit for processing CSR instructions with interface like `DummyLSU` - RS Interface on input and FU on output.
Second class is `CSRRegister` - for declaring and defining behavior of individual `CSR`s, with transactional interface. `CSRRegister`s are registered to `CSRUnit`.

Both working and tested. For more details see docstrings in code and comments on this PR.

Ready for review and merge ~~WIP CSR implementation~~
Task list:
- [x] Base CSR implementation + tests
- [x] Immediate support
- [x] Transactional interface for registering and modifying a CSR (read/write)

Rest of the tasks will be implemented in separate PR
- [ ] User Specification counter registers and Priv Specification register map implementation
- [ ] Core intergration (modifies Fetch, ROB)
- [ ] Optional: FENCE.I


piotro888:
In last commit I added `CSRRegister` class, with transactional interface for both `CSRUnit` and external CSR operations (not from instructions), as proposed and discussed on the last meeting.

1. It would be nice if someone present on that meeting could review general idea of this part of implementation 

2. I have a implementation question regarding registering of `CSRRegister`s to `CSRUnit`. Currently it is done using `.register(reg)` method of `CSRUnit`. When registers would be placed in various place in code, what would be the best method to pass them to singe CSRUnit from other modules? Passing a referene to unit or appending to some reg list in gen params doesn't seem to be good idea. Usage of annotations was propsoned on the last meeting, but I think that this problem still persists (unless CSRUnit would parse all of files in project and look for annotations in constructor)

piotro888:
RS part of last commit could be reverted, waiting for feedback on #228

tilk:
Ad 2, did you look at #209? Maybe the stuff or ideas developed there could be useful here also?

coreblocks/structs_common/csr.py
```diff
@@ -0,0 +1,187 @@
...
+    def elaborate(self, platform):
+        m = Module()
+
+        internal_method_layout = {("data", self.gen_params.isa.xlen), ("active", 1)}
+        write_internal = Record(internal_method_layout)
+        fu_write_internal = Record(internal_method_layout)
+
+        m.d.sync += self.side_effects.eq(0)
+
+        @def_method(m, self.write)
```
lekcyjna123, line 42: Just to be sure. The check if the CSR instruction should cause side effect or not will be done in a code calling this methods?

coreblocks/structs_common/csr.py
```diff
@@ -0,0 +1,187 @@
...
+
+        # Standard RS interface
+        self.rs_layouts = gen_params.get(RSLayouts)
+        self.fu_layouts = gen_params.get(FuncUnitLayouts)
+        self.select = Method(o=self.rs_layouts.select_out)
+        self.insert = Method(i=self.rs_layouts.insert_in)
+        self.update = Method(i=self.rs_layouts.update_in)
+        self.accept = Method(o=self.fu_layouts.accept)
+
+        self.regfile: dict[int, tuple[Method, Method]] = {}
```
lekcyjna123, line 88: Taking into account the fact that, there will be probably dependency resolution on CSRs, I am not sure if this regfile shouldn't be created as a separate struct.

coreblocks/structs_common/csr.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def register(self, csr: CSRRegister):
+        if csr.csr_number in self.regfile:
+            raise RuntimeError(f"CSR number {csr.csr_number} already registered")
+        self.regfile[csr.csr_number] = (csr._fu_read, csr._fu_write)
```
lekcyjna123, line 93: You use here `_fu_read` and `_fu_write`, so I am not sure why we have in `CSRRegister` also `write` and `read` methods. My intuition reading this code was that `_fu_read` will be called by FU which use CSR and `read` will be called when an program instruction accessing CSR will be given by user.

coreblocks/structs_common/csr.py
```diff
@@ -0,0 +1,187 @@
...
+                        with m.If(should_write_csr & ~done):
+                            write_val = Signal(self.gen_params.isa.xlen)
+                            with m.If((instr.exec_fn.funct3 == Funct3.CSRRW) | (instr.exec_fn.funct3 == Funct3.CSRRWI)):
+                                m.d.comb += write_val.eq(instr.s1_val)
+                            with m.If((instr.exec_fn.funct3 == Funct3.CSRRS) | (instr.exec_fn.funct3 == Funct3.CSRRSI)):
+                                m.d.comb += write_val.eq(read_val | instr.s1_val)
+                            with m.If((instr.exec_fn.funct3 == Funct3.CSRRC) | (instr.exec_fn.funct3 == Funct3.CSRRCI)):
+                                m.d.comb += write_val.eq(read_val & (~instr.s1_val))
+
+                            write(m, write_val)
```
lekcyjna123, line 149: Modification of CSR should be deferred till `commit` to avoid problems when we change csr of an FU which have a big backlog of old instructions (e.g FPU)

test/structs_common/test_csr.py
```diff
@@ -0,0 +1,198 @@
...
+
+            yield self.dut.rob_single_insn.eq(0)
+
+            op = yield from self.generate_instruction()
+
+            yield from self.dut.select.call()
+
+            yield from self.dut.insert.call({"rs_data": op["instr"]})
+
+            while random.random() < 0.5:
```
lekcyjna123, line 101: Move waiting to separate function.

test/structs_common/test_csr.py
```diff
@@ -0,0 +1,198 @@
...
+                yield
+
+            yield self.dut.rob_single_insn.eq(1)
+
+            while random.random() < 0.5:
+                yield
+
+            res = yield from self.dut.accept.call()
```
lekcyjna123, line 114: In my opinion it would be better to have accept in different test process, so that this will better test parallelism.

test/structs_common/test_csr.py
```diff
@@ -0,0 +1,198 @@
...
+            op = yield from self.generate_instruction()
+
+            yield from self.dut.select.call()
+
+            yield from self.dut.insert.call({"rs_data": op["instr"]})
+
+            while random.random() < 0.5:
+                yield
+
+            yield from self.dut.update.call({"tag": op["exp"]["rs1"]["rp_s1"], "value": op["exp"]["rs1"]["value"]})
```
lekcyjna123, line 104: If possible `update` should be in my opinion in different test process to better test parallelism.

test/structs_common/test_csr.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def generate_instruction(self):
+        ops = [
+            Funct3.CSRRW,
+            Funct3.CSRRC,
+            Funct3.CSRRS,
+        ]
+
+        op = random.choice(ops)
+        rd = random.randint(0, 2**self.gp.phys_regs_bits - 1)
```
lekcyjna123, line 71: Please change drawing distribution to often get r0 or please reduce number of `phys_regs_bits`

lekcyjna123 (commented):
This is already a big commit, so I propose to think if there is a possibility to deliver further changes in separate PRs.

coreblocks/structs_common/csr.py
```diff
@@ -0,0 +1,187 @@
+from amaranth import *
```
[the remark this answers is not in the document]
tilk: I would look for ways to minimize the problem of CSR dependencies. Even disregarding our "It's an optimization" slogan, CSRs are not supposed to be efficient to access anyway. We also need to watch out for side effects - so it's probably best to plan on CSR accesses happening only on the non-speculative path.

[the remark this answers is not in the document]
lekcyjna123: But FPU should be in speculative path and each FPU operation is expected to read `fcsr` (e.g. to check rounding mode) and potentially can modify `fcsr`, so I think that CSR dependencies will be needed if we don't want to serialise all FPU operations.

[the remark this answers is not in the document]
tilk: Ah, the pesky FPU. Maybe FPU instructions don't really need to access the CSR directly, only update some internal flag information, which gets to the actual CSR only on commit? The flag setting can be reordered anyway (otherwise FPU instructions couldn't be reordered), so the only order that matters is that the CSR read sees all the flags set by FPU instructions before.

test/structs_common/test_csr.py
piotro888: Yes, these tests are incorrect and unit is buggy in case of `rp_s1` = 0.
It was not testing the case when `rp_s1` = 0 and value is passed immediately in s1_val.
Sorry for not mentioning that, I updated the tests earlier, but not pushed that, because it required additional fields from scheduler, so I created #228 first. I will merge that change to this PR and push updated tests and fixed unit.

coreblocks/structs_common/csr.py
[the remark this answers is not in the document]
piotro888: The implementation proposition mentioned on earlier meetings handled that by stalling the core in Fetch stage on CSR instruction (just like it is done in current jump implementation), because CSR accesses doesn't need to be efficient.
I'm writing this implementation according to this assumption (and that's why there is`fetch_continue` method and `rob_empty` signal in `CSRUnit`). 
I also delay execution of CSR instruction in `CSRUnit`, until all currently executed instructions are finished, because these can trigger CSR changes.
I think that this is good enough method of keeping everything in sync, if we don't matter about performance of CSR instructions with great improvement in simplicity.

Regarding FPU, I agree with @tilk propositon to reoder flag updates in FPU unit and write them to CSRs on commit. CSR read would see flags in order, because it will wait for ROB to be empty (all instruction commited)

coreblocks/structs_common/csr.py
piotro888: See [comment](https://github.com/kuznia-rdzeni/coreblocks/pull/217#discussion_r1118910856)

coreblocks/structs_common/csr.py
piotro888: Yes, this is checked in `CSRUnit` by calling `_fu_read` and `_fu_write` methods under `should_read` and `should_write` `If`s. [code link](https://github.com/kuznia-rdzeni/coreblocks/pull/217/files#diff-61a9003bae2910c5aa3221d7aead9f1f3bc541a73dc6201a42a766a8bd01fae5R110)

test/structs_common/test_csr.py
piotro888: I coded tests in that way to handle fetch stall/unstall checking and simulate real instruction flow.
With assumptions about this CSR implementation, only one CSR instruction could be present at the time, so this test represents next steps that cannot happen in any other order in real scenario. 
Maybe it could be possibly split, but with this assumptions parallel test process would execute exactly in the same way and would not improve coverage, but only complicate the test.

piotro888:
I added missing docstrings and as @lekcyjna123 suggested I think it is ready for merging now.
Integration to Core, better interface for registering `CSRRegisters` and implementing basic CSRs from specification could be done in next PR. For now this PR specifies complete and tested unit for CSR operations and CSR Register class.

Note: Depends on #228, but it should be merged soon.

coreblocks/structs_common/csr.py
```diff
@@ -0,0 +1,278 @@
...
+
+from coreblocks.transactions import Method, def_method, Transaction
+from coreblocks.utils import assign
+from coreblocks.params.genparams import GenParams
+from coreblocks.params.layouts import FuncUnitLayouts, CSRLayouts
+from coreblocks.params.isa import Funct3
+
+
+class CSRRegister(Elaboratable):
```
tilk, line 10: The current design of `CSRRegister` basically decouples CSR reads/writes from the modules handling them, and I don't know if that is a good idea. In particular, I believe it's now impossible for CSR side effects to have latency. The `_fu_read` and `_fu_write` methods are always ready, and there is no way to make the CSR unit wait if the operation cannot be performed in a single clock cycle. Also, in the example code in the docstring below, if there is some latency in the transaction, information about side effects will be silently lost. In other words, this design is not latency-insensitive.

coreblocks/structs_common/csr.py
```diff
@@ -0,0 +1,278 @@
...
+            return {"data": self.value, "read": self.side_effects.read, "written": self.side_effects.write}
+
+        @def_method(m, self._fu_read)
+        def _():
+            m.d.sync += self.side_effects.read.eq(1)
+            return self.value
+
+        # Writes from instructions have priority
+        with m.If(fu_write_internal.active & write_internal.active):
+            m.d.sync += self.value.eq((fu_write_internal.data & ~self.ro_bits) | (write_internal.data & self.ro_bits))
```
lekcyjna123, line 105: What happen when:
- in csr we have bit ro (let it be `a`-th bit)
- `a`-th bit store value `1`
- in current cycle we change bit `b`-th in this CSR (`a!=b`) which is rw
- there is no `write` in this cycle (so `write_internal.data` is 0)

What will be the value of `a`-th bit in the next cycle? If I correctly see it will be 0 instead of 1.

tilk (approved):
Let's merge now and worry later.

tilk:
#241 did its job :)

PR #227, opened 2023-02-08
Make Collector a method combinator

tilk:
Collector was supposed to be in the combinator section of the lib, and conform to the combinator interface. Somehow, this wasn't enforced on code review. This PR fixes that.

### DIFF
--- a/coreblocks/core.py
+++ b/coreblocks/core.py
@@ -50,7 +50,7 @@ class Core(Elaboratable):
 
         self.announcement = ResultAnnouncement(
             gen=self.gen_params,
-            get_result=self.result_collector.get_single,
+            get_result=self.result_collector.method,
             rob_mark_done=self.ROB.mark_done,
             rs_write_val=self.update_combiner.method,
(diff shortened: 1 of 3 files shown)


lekcyjna123:
Could you please point what was changed in code? I see that only variable names have been changed. Or I have missed something?
obraz

tilk:
> Could you please point what was changed in code? I see that only variable names have been changed. Or I have missed something? obraz

That's exactly what was changed.

lekcyjna123:
Thanks. I understand it now. I have though, that it wasn't conform to the combinator interface in implementation, not in way of variable naming.

tilk:
> Thanks. I understand it now. I have though, that it wasn't conform to the combinator interface in implementation, not in way of variable naming.

This might be formalized later using protocols or ABCs, but first #201 should be merged. Also, let's see what happens in #209, where this PR should get used.

speederking07 (approved):
I have not seen any bugs

PR #228, opened 2023-02-18
Allow different layouts for RS units

piotro888:
Creates `RSInterfaceLayouts` class that specifies `Scheduler` -> `RS` interface with all of available signals.
Specific RSes in their own layouts are allowed to implement subset of these fields, including our generic integer RS.

Currently there are units like LSU and CSR (WIP), which doesn't use our RS implementation, but are using RS interface to connect to `Scheduler`

It will be very useful and more optimal for #217 (CSR), because it uses some fields that are exclusive to it (like `csr` and unchanged register number) that would waste space inside other RSes. Without this change we could ignore these fileds inside CSR, but that would mean adding list of ignored fields to each RS or depend on optimizer, but WakeupSelect layouts would then differ and tests would need to be modifed. 
All of the decoded fields are needed to be passed to the end of the `Scheduler`, but could be separated to only necessary fields for each RS (or special units connected like RS - LSU, CSR)

It would be also useful for `FENCE` instructions handling, which produces additional three fields in decoder, and maybe for later FPU (?).

~~One question: If LSU uses the same layouts and fields as our RS, should it specify its own interface in `LSULayouts` or simply use ones from `RSLayouts`?~~

~~Depends on #222~~


coreblocks/utils/_typing.py
```diff
@@ -10,3 +10,4 @@
 StatementLike: TypeAlias = Statement | Iterable["StatementLike"]
 LayoutLike = Layout | Sequence[Tuple[str, ShapeLike | "LayoutLike"] | Tuple[str, ShapeLike | "LayoutLike", Direction]]
 DebugSignals: TypeAlias = Signal | Record | Iterable["DebugSignals"] | Mapping[str, "DebugSignals"]
+LayoutList = list[Tuple[str, ShapeLike | "LayoutList"]]
```
tilk, line 13: What's the point of `LayoutList`? What's wrong with `LayoutLike`?

coreblocks/utils/utils.py
```diff
@@ -205,6 +205,7 @@ def assign(lhs: AssignLHS, rhs: AssignRHS, *, fields: AssignFields = AssignType.
     else:
         if not isinstance(fields, AssignType):
             raise ValueError("Fields on assigning non-records")
+
```
tilk, line 208: Irrelevant change for this PR.

coreblocks/utils/utils.py
```diff
@@ -219,3 +220,11 @@ def has_explicit_shape(val: ValueLike):
...
                 )
         yield lhs.eq(rhs_val)
+
+
+def layout_subset(layout: LayoutList, *, fields: set[str]) -> LayoutList:
+    result = []
+    for field in layout:
+        if field[0] in fields:
+            result.append(field)
+    return result
```
tilk, line 230: Why not use a list comprehension?
```suggestion
    return [item for item in layout if item[0] in fields]
```

coreblocks/params/layouts.py
```diff
@@ -137,12 +139,14 @@ def __init__(self, gen_params: GenParams):
         self.retire_layout = [("rob_data", self.data_layout), ("rob_id", gen_params.rob_entries_bits)]
 
 
-class RSLayouts:
+class RSInterfaceLayouts:
```
tilk, line 142: What's the point of `RSInterfaceLayouts`? You defined it for some reason, why isn't it used anywhere (other than `RSLayouts`)? (I imagined it would be used by other RS-like modules?)

coreblocks/utils/_typing.py
piotro888: `LayoutLike` unfortunately doesn't work because of `current_instr = Record(self.rs_layouts.data_layout + [("valid", 1)])` [line](https://github.com/kuznia-rdzeni/coreblocks/blob/f5216a26ceaee564b856d59f75e4d37313372a2e/coreblocks/lsu/dummyLsu.py#L241) in LSU. Typing checker returns error that `+` operator is not supported for Layouts, Sequences and Tuples, as it depends on specific `List` functionality. 
I created that function to be used specifically for in Layouts classes, which uses more 'strict' `list` format.

coreblocks/utils/utils.py
piotro888: Must be artifact from rebase, will correct that in a moment

coreblocks/params/layouts.py
piotro888: > I imagined it would be used by other RS-like modules?
That is the use case for it is that it specifies all of the available signals, from where RS-like modules take their subsets.
It will be used for example in CSRUnit, with different subset of signals.
It should probably also be present is `LSULayouts`, but it will be identical to this one in `RSLayouts` (for now LSU uses layouts from `RSLayouts` but now I think it should definitely be copied to `LSULayouts`)

coreblocks/utils/utils.py
piotro888: I guess I haven't written enough Python code recently :)

coreblocks/utils/_typing.py
tilk: I understand now - indeed, `Sequence` does not define the addition operator.

I think it would be warranted to add at least a newline before `DebugSignals`. The previous types basically name the concepts present in Amaranth, but `DebugSignals` and `LayoutList` are types invented for Coreblocks use.

coreblocks/params/layouts.py
piotro888: Used in LSU in last commit https://github.com/kuznia-rdzeni/coreblocks/pull/228/commits/040b9244d9b1ac19d0a15938da3be5e6036ab864

coreblocks/utils/_typing.py
piotro888: Would comments like `# Types representing Amaranth concepts` and `# Internal Coreblocks types`  be helpful and are ok?

tilk: Yes.

tilk (approved):

coreblocks/scheduler/scheduler.py
```diff
@@ -309,6 +309,8 @@ def elaborate(self, platform):
                 "rs_data": {
                     "rp_s1": Mux(source1.valid, 0, instr.regs_p.rp_s1),
                     "rp_s2": Mux(source2.valid, 0, instr.regs_p.rp_s2),
+                    "rp_s1_reg": instr.regs_p.rp_s1,
```
lekcyjna123, line 312: Here you have added "rp_sN_reg" fields, but in layout you have defined "rp_sN_val" field, so this fields will never be matched.

lekcyjna123 (approved):

PR #235, opened 2023-03-05
Yet another generic RS configuration

lekcyjna123:
As I wasn't happy with the last changes from #209, so I decided today to make exploration by coding and I have prepared my own PoC with ideas presented by @speederking07 in one of his previous versions. My commit reintroduce keys to dependency network but with following changes:
- we have only one base class for all keys (`DependencyKey`) - no longer separate `MethodKey` 
- there are two ways of creating new keys:
  - one by inheritance and new class creation (for "standard dependencies e.g. `commit`)
  - second by name using generic base class `DependencyKey`
- removed connecting methods by names - now the key name is used only to generate hash (to hash generation we can add also class name to make sure that derived class name will be unique without need to take care about `name` field for them
- way of unifying method is now part of the key - no module need to know how to make unification, it can simply call unification method from key
- each key instance if equal to each other instance of the same key class

Point to discuss/future improvements:
- Do `ComponentConnections` class more generic (as for now it is releated to FU) (maybe add as a part of amaranth platform?)\
- add implementation of one-to-one dependecy by setting unifier to `None`

This is proof of concept, so there is a lot of ugly things in code, but they are easily fixable. Please feel free to comment.

### DIFF
--- a/coreblocks/core.py
+++ b/coreblocks/core.py
@@ -16,6 +16,7 @@ from coreblocks.stages.retirement import Retirement
 from coreblocks.peripherals.wishbone import WishboneMaster
 from coreblocks.frontend.fetch import Fetch
 from coreblocks.utils.fifo import BasicFifo
+from coreblocks.transactions.lib import MethodProduct
 
 __all__ = ["Core"]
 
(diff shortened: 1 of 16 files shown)


tilk:
I made some changes to your proposal in #236.

speederking07 (commented):
It looks fine for me, it is slightly different approach to keys, and now they looks better than ib my version

PR #236, opened 2023-03-06
RS configuration: proposed changes to @lekcyjna123's proposed changes

tilk:
Instead of reviewing, I proposed some changes myself. Have fun! Changes:

* Removed `name` from keys. Subclassing is an approach which is more descriptive and less prone to typos, so should be the default. Keys which have parameters are still possible, but should be created explicitly.
* Removed `dep_type` too. As the key type implies the dependency type (in the recommended use case), it's not needed.
* Created a hierarchy of abstract key classes:
  - `DependencyKey` is the base one. It has two type parameters: one for values related to the key, and the second for the values returned from the key.
  - `SimpleKey` is for keys which relate to a single value.
  - `UnifierKey` is for keys relating to potentially multiple methods, combined by an `Unifier`.
  These abstract key classes should never be instanced, they are for subclassing.
* Modified `ComponentConnections` so that all kinds of keys are treated the same.

Edit: More changes:

* Renamed `ComponentConnections` to `DependencyManager`, as that's what it does. If we use the dependency framework for something else later, it won't look weird.
* Removed passing `DependencyManager` around using a parameter. We already have something passed everywhere as a parameter, it's `GenParams`, and we can leverage it for this purpose, too.
* Added a safeguard to prevent instantiation of abstract key classes.
* Removed method chaining from adding dependencies. Method chaining is not idiomatic in Python, this is the only place with method chaining in Coreblocks, and it doesn't give a lot of benefit too, as adding dependencies is rarely done.

### DIFF
--- a/coreblocks/core.py
+++ b/coreblocks/core.py
@@ -42,7 +42,7 @@ class Core(Elaboratable):
         self.func_blocks_unifier = FuncBlocksUnifier(
             gen_params=gen_params,
             blocks=gen_params.func_units_config,
-            connections=ComponentConnections().set_dependency(WishboneDataKey(), wb_master_data),
+            connections=ComponentConnections().with_dependency(WishboneDataKey(), wb_master_data),
             extra_methods_required=[InstructionCommitKey(), BranchResolvedKey()],
         )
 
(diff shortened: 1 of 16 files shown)


lekcyjna123:
Removing `name` from key classes is good idea, but how you want to realise value-parametrised keys? By creating a separate class which can get value as parameter during construction?

tilk:
> Removing `name` from key classes is good idea, but how you want to realise value-parametrised keys? By creating a separate class which can get value as parameter during construction?

Exactly, and thanks to the magic of dataclasses, everything will work just fine.

I think that this subsystem will require adequate documentation, but let's agree on an implementation first.

tilk:
Made some changes to make this more polished. If this looks fine for you, let's start the merging party.

Arusekk (approved):
This smells a bit of Java Corporate Beans Dependency Injection Design Pattern, but I like it.  Less code ≈ less bugs.

lekcyjna123 (approved):

PR #244, opened 2023-03-13
Added overview of publications about exceptions

lekcyjna123:
Hi,

Similarly as in previous semesters I have made an overview of different publication before implementation starts. This time it is related with exceptions/interrupts/speculation. Previously I have created pdf which I presented on our meetings. Now we have a wiki with documentation, so I have changed the convention and prepared summary as markdown which can be merged into wiki. What do you think about that idea? Want we push such work to repository, or we stay by the previous solution?


lekcyjna123:
As agreed I checked grammar with NN model, to do that I used: https://huggingface.co/pszemraj/flan-t5-large-grammar-synthesis

Few comments:
- it looks like this model wasn't learnt on codding related data e.g. "Linux" was changed to "Linguistics"
- it changed one of the number - in original text there was 32 and in "corrected" 16
- it don't know about "markdown" concept and syntax - so if we would like to use it more, then we should write or find a script which will extract plain text from markdown, correct it with model and next put it back to markdown

tilk (approved):

PR #245, opened 2023-03-14
Fix the core_graph script after merging #209, and other cleanups

tilk:
~~This PR quick-and-dirty-fixes the core_graph script, as it still uses an old GenParams construction.~~ (To catch errors like this, implement #232.) This PR cleans up the default configuration of the core, so that it's the same in core_graph, synthesis and test_core. Also, the data structure sizes are now the same as they were before in synthesis.


lekcyjna123 (approved):
So the riddle of additional two bits have been solved. Good to hear that!

Arusekk (approved):
Looks good

PR #246, opened 2023-03-14
ZBA extension built into ALU

wojpok:


coreblocks/params/optypes.py
```diff
@@ -54,6 +55,7 @@ class OpType(IntEnum):
         OpType.EBREAK,
         OpType.MRET,
         OpType.WFI,
+        OpType.ADRESS_GENERATION,
```
tilk, line 58:
```suggestion
        OpType.ADDRESS_GENERATION,
```

coreblocks/params/optypes.py
```diff
@@ -31,6 +31,7 @@ class OpType(IntEnum):
     MUL = auto()
     DIV_REM = auto()
     SINGLE_BIT_MANIPULATION = auto()
+    ADRESS_GENERATION = auto()
```
tilk, line 34:
```suggestion
    ADDRESS_GENERATION = auto()
```

coreblocks/fu/alu.py
```diff
@@ -110,7 +123,7 @@ def elaborate(self, platform):
 
 
 class AluFuncUnit(Elaboratable):
-    optypes = {OpType.ARITHMETIC, OpType.COMPARE, OpType.LOGIC, OpType.SHIFT}
+    optypes = {OpType.ARITHMETIC, OpType.COMPARE, OpType.LOGIC, OpType.SHIFT, OpType.ADRESS_GENERATION}
```
tilk, line 126:
```suggestion
    optypes = {OpType.ARITHMETIC, OpType.COMPARE, OpType.LOGIC, OpType.SHIFT, OpType.ADDRESS_GENERATION}
```

tilk (commented):
You also need to extend `frontend/decoder.py`, but otherwise, pretty nice change.

wojpok:
I've just fixed all the typos and extended frontend decoder and it seems like this draft is ready to be converted to pull request

tilk (approved):

coreblocks/params/isa.py
```diff
@@ -46,15 +46,17 @@ class Funct3(IntEnum):
...
     BGEU = AND = CSRRCI = REMU = REMUW = 0b111
 
 
 class Funct7(IntEnum):
     SL = SLT = ADD = XOR = OR = AND = 0b0000000
     SA = SUB = 0b0100000
-    MULDIV = 0b0000001
+    MULDIV = SH1ADD = 0b0000001
+    SH2ADD = 0b0000010
+    SH3ADD = 0b0000011
```
tilk, line 59: Change encoding of SHADD instructions according to the spec.

tilk (changes requested):

coreblocks/params/isa.py
```diff
@@ -43,18 +43,19 @@ class Opcode(IntEnum):
...
-    BLTU = OR = CSRRSI = REM = REMW = 0b110
+    BLTU = OR = CSRRSI = REM = REMW = SH3ADD = 0b110
     BGEU = AND = CSRRCI = REMU = REMUW = 0b111
 
 
 class Funct7(IntEnum):
     SL = SLT = ADD = XOR = OR = AND = 0b0000000
     SA = SUB = 0b0100000
     MULDIV = 0b0000001
+    SHADD = 0b0010000
```
tilk, line 58: Let's use the names from the spec: SH1ADD, SH2ADD and SH3ADD (their value is the same, but so what, same thing happens with AND, OR for example).

tilk (approved):

coreblocks/params/isa.py
wojpok: Resolved. In the future we may want to rename some other parameters as well, since there are some other operations in ALU that share parameter names

tilk:
Damn it, that's another detail missed! I sure hope it's the last one.

PR #248, opened 2023-03-15
Add a script exporting the core to Verilog

### LINKED ISSUE #204: Export core to Verilog
After #96 we will finally have full support for RV32I. In order to actually do something with the core (in particular, test it thoroughly), we need to export it to Verilog, exposing the Wishbone interface(s).

xThaid:
~There is at least one thing yet to improve. Right now, the signal names of the top module are not really descriptive: `wbMaster__dat_w, wbMaster__rst, wbMaster__ack, wbMaster__adr, wbMaster__cyc ...`~ Done

Fixes #204


scripts/gen_verilog.py
```diff
@@ -0,0 +1,86 @@
...
+        m.d.comb += wb_master_data.wbMaster.connect(self.wb_data)
+
+        m.submodules.wb_master_instr = wb_master_instr
+        m.submodules.wb_master_data = wb_master_data
+        m.submodules.c = self.core
+
+        return tm
+
+
+def record_to_signals(record: Record) -> list[Signal]:
```
tilk, line 50: Check if `flatten` from `test/gtkw_extension.py` works for this use case, and if so, move `flatten` to utils and give it a sensible type.

xThaid: Done

coreblocks/utils/_typing.py
```diff
@@ -10,6 +10,7 @@
 ShapeLike = Shape | int | range | Type[Enum]
 StatementLike: TypeAlias = Statement | Iterable["StatementLike"]
 LayoutLike = Layout | Sequence[Tuple[str, ShapeLike | "LayoutLike"] | Tuple[str, ShapeLike | "LayoutLike", Direction]]
+SignalLike = Signal | Record | Iterable["SignalLike"] | Mapping[str, "SignalLike"]
```
tilk, line 13: 1. This is a section containing the types needed to type Amaranth.
2. This is actually the same as DebugSignals, maybe just find a better name for it?

tilk: SignalBundle sounds OK I guess.

xThaid: Done

tilk (approved):

lekcyjna123 (approved):

PR #256, opened 2023-03-19
Abstract away implementation of Wishbone master

xThaid:
As discussed in our weekly meeting, it is more reasonable to expose Wishbone bus instead of Wishbone master in the `Core` module. In short, a back-end user of `Core` module shouldn't care about how we implemented the bus master.

I put `WishboneBus` in the constructor of `Core` module to emphasize the fact that these are the only wires that the user need to take care of.

Depends on #251, #204


lekcyjna123 (commented):
It looks good.

tilk (approved):
Seems okay.

lekcyjna123 (approved):

PR #259, opened 2023-03-21
Enums enums enums

Arusekk:
Maybe these two commits should be in separate PRs; the only thing in common is Enum, but they are both relatively small (ekhm), so let's keep it this way for now.  +1 for splitting from myself.

### DIFF
--- a/coreblocks/params/isa.py
+++ b/coreblocks/params/isa.py
@@ -35,6 +35,9 @@ class BitEnum(ValueCastableHack, Enum):
     """
     A helper class that defines Amaranth enums with a width
     """
+
+    __width: int
+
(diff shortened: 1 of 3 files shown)


Arusekk:
Actually, some progress is pending in Amaranth with the widthed enums. We should probably defer them until they reach some stability.

lekcyjna123 (approved):
For me it looks good. I think that we can introduce this enums till there will be official release of 0.4 (maybe with comment in code, that they should be removed on new amaranth version)

tilk:
Can your widthed enums be used as shapes (and therefore be used instead of bit widths in various layouts)?

tilk (approved):

PR #260, opened 2023-03-22
Multiport fifo

lekcyjna123:
To have proper implementation of interrupts we are going to need multiport FIFO for FreeRF, so here is a draft of such FIFO.

ToDo:
- [x] Implement MultiportFIFO
- [x] Write one port tests for MultiportFIFO
- [x] Write many port tests for MultiportFIFO
- [x] ~Bonus: create Multi selection module~

### DIFF
--- a/coreblocks/utils/fifo.py
+++ b/coreblocks/utils/fifo.py
@@ -1,10 +1,9 @@
 from amaranth import *
 from coreblocks.transactions import Method, def_method, Priority, TModule, Transaction
-from amaranth.utils import log2_int
 from coreblocks.transactions._utils import MethodLayout
 import coreblocks.transactions.lib as tlib
-from coreblocks.utils._typing import ValueLike, ModuleLike
-from coreblocks.utils.utils import popcount, count_leading_zeros
+from coreblocks.utils._typing import ValueLike
+from coreblocks.utils.utils import popcount
 
 
 class BasicFifo(Elaboratable):
@@ -181,7 +180,6 @@ class _InternalMultiportFifo(Elaboratable):
         m.d.top_comb += write_grants.eq(Cat(method.run for method in self.write_methods))
         m.d.top_comb += write_grants_count.eq(popcount(write_grants))
 
-
         with m.FSM():
             for i in range(self.fifo_count):
                 with m.State(f"current_write_{i}"):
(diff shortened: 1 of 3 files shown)


coreblocks/utils/fifo.py
```diff
@@ -108,3 +110,144 @@ def _() -> None:
             m.d.sync += self.write_idx.eq(0)
 
         return m
+
+
+def popcount(m: Module, x: Signal) -> Signal:
```
lekcyjna123, line 115: Move popcount to utils

tilk:
So it's _just an optimization_. ;)

lekcyjna123:
Yes, but I think that this optimization, can simplify implementation, because we will not have to take into account some weird corner cases e.g. RobAllocation want to write on not already dealocated old entry.

tilk:
> because we will not have to take into account some weird corner cases e.g. RobAllocation want to write on not already dealocated old entry.

Now you lost me again.

coreblocks/utils/fifo.py
```diff
@@ -108,3 +122,171 @@ def _() -> None:
...
+        self.depth = depth
+        self.port_count = port_count
+        self.fifo_count = fifo_count
+        self.init = init
+
+        self.connected_reads = 0
+        self.connected_writes = 0
+
+        self._read_methods = [Method(o=self.layout) for _ in range(self.port_count)]
+        self._write_methods = [Method(i=self.layout) for _ in range(self.port_count)]
```
Arusekk, line 187: The methods will be anonymous, maybe we could give them some (partially) useful names?

test/utils/test_fifo.py
```diff
@@ -59,14 +77,14 @@ def source():
...
 
-                v = random.randint(0, (2**fifoc.fifo.width) - 1)
-                yield from fifoc.fifo_write.call(data=v)
-                expq.appendleft(v)
-
                 if random.random() < 0.005:
                     yield from fifoc.fifo_clear.call()
                     expq.clear()
 
+                v = random.randint(0, (2**fifoc.fifo.width) - 1)
```
Arusekk, line 84: ;-)
```suggestion
                v = random.randrange(2**fifoc.fifo.width)
```

lekcyjna123:
I uploaded tests for multiporting. Additionally there is some weird bug, which cause that input of one of writing ports freeze. This is random error and hard to reproduce, so I will have to dig deeper. In most cases it was connected with sending `clear` signal, but now I have found an reproduction where no clearing was issued.

Arusekk (commented):
Better to ask for forgiveness than permission :)

lekcyjna123:
Fixed implementation of multiport fifo.

PR #263, opened 2023-03-24
New syntax for method mocks

Arusekk:
Inspired by #212.  BTW, def_method_mock can be unified with def_class_method_mock if we wish (using smart `__get__` field descriptors).


test/scheduler/test_rs_selection.py
```diff
@@ -102,7 +102,7 @@ def process():
         return process
 
     def create_rs_alloc_process(self, io: TestbenchIO, rs_id: int, rs_optypes: set[OpType], random_wait: int = 0):
-        def mock(_):
+        def mock(arg):
```
tilk, line 105: Remove `arg`.

coreblocks/transactions/core.py
```diff
@@ -758,6 +759,26 @@ def debug_signals(self) -> SignalBundle:
         return [self.ready, self.run, self.data_in, self.data_out]
 
 
+def call_func_smart(method, func: Callable[..., T], arg=None, /, **kwargs) -> T:
```
tilk, line 762: Look for a better name.

Arusekk: Maybe `auto_keyword_record_call` or `record_keyword_helper`?
Also, I think the function could be moved to `_utils`, as it is more of an implementation detail.

tilk: Or `method_call_helper`? I think that exposing the word `record` is not helpful, as the records are an implementation detail, and also we will be trying to go away from records here in favor of `amaranth.lib.data`.

As for moving to `_utils`, good idea.

lekcyjna123 (commented):
Looks good. After fixing comment from above we can merge it.

tilk (approved):

xThaid (approved):
Looks good

PR #265, opened 2023-03-25
Make 'enable' parameter in the method mock a lambda

xThaid:
This will allow mocks to be dynamically enabled or disabled.

I also removed `condition` parameter as I find it redundant when we have `enable`.

~I also believe that by making combinatorial tests synchronous we can get rid of this cryptic parameter `settle`. Let me know if I am wrong.~
edit: I convinced myself that we may need extra `yield Settle()` to impose correct function scheduling.


tilk:
> However, this causes a problem one of the existing tests.

That's exactly the test I had issues with when introducing mocks. This warrants a closer look.

xThaid:
I have found the problem, but haven't come up with a neat solution yet. The problem is in the order in which we run mock functions. Consider the following code:

```python
foo_q = deque()
bar_q = deque()

foo_q.append(11)
bar_q.append(11)

@def_method_mock(lambda: ...)
def method_foo(arg):
    return foo_q.popleft()

@def_method_mock(lambda: ...)
def method_bar(arg):
    self.assertEqual(arg, bar_q.popleft())
```

and in HDL we have:

```python
with Transaction().body(m):
    foo_res = self.foo(m)

    self.bar(m, foo_res)
```

The correctness now relies on the order in which the mock functions are executed. There are two scenarios to consider:
1.If `method_foo` runs first, then we remove an item from the queue and transmit signals. Thus, when `method_bar` is called, its argument is up-to-date, and the assertion will not fail.
2.If `method_bar` runs first, its argument will not have been updated yet, resulting in the assertion failing.

A simple solution would be to manually specify the order by adding one `yield Settle()`:
```python
...
@def_method_mock(lambda: ..., settle=0)
def method_foo(arg):
    return foo_q.popleft()

@def_method_mock(lambda: ..., settle=1)
def method_bar(arg):
    self.assertEqual(arg, bar_q.popleft())
```

Perhaps renaming the `settle` parameter to a more descriptive name, such as `priority`, would be beneficial. 

Regarding the failed test, the functions were originally scheduled in the correct order, but after making changes, they were scheduled differently, causing one of the assertions to fail for the reason previously mentioned.

tilk:
> Perhaps renaming the `settle` parameter to a more descriptive name, such as `priority`, would be beneficial.

I wonder. There might be uses of `Settle` outside the mocks, and they will interact with the mocks. Changing the name hides the fact that the parameter controls the use of `Settle`, which may lead to surprises later. Maybe `settle_count` or something like that?

xThaid:
Let's meet half way. I left `settle` name in `TestbenchIO`, but stayed with `sched_prio` in mocks, as I feel like this is the real concern that a test writer need to worry about. If mocks will be scheduled in the correct order, signals will be propagated correctly.

The parameter name in `method_handle` is `extra_settle_count` is because I want to emphasize the fact that there is already one `Settle`, which I believe is necessary. Without it, `enable` signal wouldn't be propagated and `done()` would be stale (as `done` combinatorially depends on `enable`).

tilk (approved):

lekcyjna123 (approved):

PR #268, opened 2023-03-30
Cocotb support

tilk:
This PR implements cocotb testing support, currently using Icarus Verilog as the simulator. Implementing correct Verilator support is left for future PR, as my experiments with cocotb + current Verilator failed.

- [x] Memory models in cocotb
- [x] Build some tests
- [x] Run some tests


tilk:
This `cocotbext_wishbone` `WishboneSlave` stuff looks like nonsense, WTF.

tilk:
pyelftools

tilk:
Good news everyone, the tests run now. Bad news everyone, only two pass ;) My intuition is that the core deadlocks after a certain amount of instructions executed, as most of the tests are timeouting, and only the two shortest are passing. Probably the retirement phase needs another look (?).

piotro888:
just reporting: tests unfortunatelly fail to run on my machine.
It fails with following error: 
```
coreblocks/test/cocotb/test.py", line 224, in test
data += bytes(memsz - len(data))
ValueError: negative count
```
I checked that segment on which it is failing is `.riscv.attributes` with `memsz = 0` and `len(data) = 23`

tilk:
> I checked that segment on which it is failing is `.riscv.attributes` with `memsz = 0` and `len(data) = 23`

Added a quick workaround, try now.

tilk:
Now only 10 tests pass. Few of them are obvious: `ma_data` (misaligned accesses), `fence_i` (self-modifying code), `s*` (stores, didn't implement them yet in the test script). Others are a mystery, e.g. `bgeu`.

Kristopher38:
Branches weren't properly tested so I don't see it as a surprise.

tilk:
Added store support and fixed a bug in the decoder, preventing some load instructions from working correctly. This didn't increase the number of tests passing, though.

.github/workflows/cocotb.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Run cocotb
+        run: |
+          cd test/cocotb && make
+
+      - name: Test Report
+        uses: EnricoMi/publish-unit-test-result-action@v2
+        with:
+          files: test/cocotb/results.xml
+          check_name: cocotb tests
```
piotro888, line 47: Could you rename this check name to something like `cocotb test results`, please?. If found it very confusing when searching for test logs. (Optionally `build` action could be also renamed to something like build and test/run)

tilk:
During the development, four different bugs were discovered and fixed! (#279, #282, #283, #284) These PRs should be merged before cleaning up and merging this one.

lekcyjna123 (approved):

piotro888 (approved):
Looks good, let's merge it! 

Optional change would be to raise error on `RWX` flags combinaton with current model, if we care about running other external ELFs.

PR #269, opened 2023-03-30
Auto debug signals from list

lekcyjna123:
As a part of debuging #260 I have found that `auto_debug_signals` have problems when methods are in list. Here is extension which add support for this case.

### DIFF
--- a/coreblocks/utils/debug_signals.py
+++ b/coreblocks/utils/debug_signals.py
@@ -14,6 +14,9 @@ def auto_debug_signals(thing) -> SignalBundle:
     slist: list[SignalBundle] = []
     smap: dict[str, SignalBundle] = {}
 
+    if "__dict__" not in dir(thing):
+        return []
+
     for v in vars(thing):
         a = getattr(thing, v)
         if isinstance(a, HasDebugSignals):
@@ -27,6 +30,18 @@ def auto_debug_signals(thing) -> SignalBundle:
             smap[v] = a
         elif isinstance(a, Signal) or isinstance(a, Record):
             slist.append(a)
+        elif isinstance(a, list):
+            submap={}
+            for i, e in enumerate(a):
+                if isinstance(e, HasDebugSignals):
+                    e.name = f"{v}[{i}]"
+                    submap[e.name] = e.debug_signals()
+                else:
+                    sublist = auto_debug_signals(e)
+                    if sublist:
+                        submap[f"{v}[{i}]"] = sublist
+            if submap:
+                smap[v]=submap
 
     slist.append(smap)
     return slist
--- a/coreblocks/utils/debug_signals.py
+++ b/coreblocks/utils/debug_signals.py
@@ -31,17 +31,16 @@ def auto_debug_signals(thing) -> SignalBundle:
         elif isinstance(a, Signal) or isinstance(a, Record):
             slist.append(a)
         elif isinstance(a, list):
-            submap={}
+            submap = {}
             for i, e in enumerate(a):
                 if isinstance(e, HasDebugSignals):
-                    e.name = f"{v}[{i}]"
-                    submap[e.name] = e.debug_signals()
+                    submap[f"{v}[{i}]"] = e.debug_signals()
                 else:
                     sublist = auto_debug_signals(e)
                     if sublist:
                         submap[f"{v}[{i}]"] = sublist
             if submap:
-                smap[v]=submap
+                smap[v] = submap
 
     slist.append(smap)
     return slist


coreblocks/utils/debug_signals.py
```diff
@@ -27,6 +30,17 @@ def auto_debug_signals(thing) -> SignalBundle:
             smap[v] = a
         elif isinstance(a, Signal) or isinstance(a, Record):
             slist.append(a)
+        elif isinstance(a, list):
```
tilk, line 33: Is generalizing this to `Collection` useful?

lekcyjna123: Yes. In case of list of methods which can have length depending on generation parameters. E.g. in multiport fifo number of ports depends on parameters given by user, so methods are stored in the list.

On the other hand we can decide to don't store methods in list and instead of that use `self[f"method_name_{i}"]` but I am not sure if this don't constrain us in future.

coreblocks/utils/debug_signals.py
```diff
@@ -27,6 +30,17 @@ def auto_debug_signals(thing) -> SignalBundle:
...
+            submap = {}
```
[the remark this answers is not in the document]
lekcyjna123: Because I want to have mapping name -> signals similar as in `smap`

coreblocks/utils/debug_signals.py
tilk: You didn't answer the question...

coreblocks/utils/debug_signals.py
```diff
@@ -27,6 +30,17 @@ def auto_debug_signals(thing) -> SignalBundle:
...
+            for i, e in enumerate(a):
+                if isinstance(e, HasDebugSignals):
+                    submap[f"{v}[{i}]"] = e.debug_signals()
+                else:
+                    sublist = auto_debug_signals(e)
+                    if sublist:
+                        submap[f"{v}[{i}]"] = sublist
```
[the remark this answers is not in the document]
lekcyjna123: Yes. It is some duplication of functionality, but without that generation from lists don't worked. If I call `auto_debug_signals` directly on `e`, then I will have infinity loop, because `e` which is potentially a `Method` will be treated as `Module` and there will be an iteration over all `for v in vars(thing)`

coreblocks/utils/debug_signals.py
lekcyjna123: Sorry. I wrongly understand your question. I will try to generalize.

lekcyjna123: Fixed.

coreblocks/utils/debug_signals.py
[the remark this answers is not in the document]
lekcyjna123: I probably found one reference cycle. In TransactionBase we store `self.owner : Elaboratable`, which is needed for graphs. So when we iterate over vars of method we see it and by recursion we call `auto_debug_signals` after few steps of processing owner we are back in this same method.

lekcyjna123:
I have rewritten this function so that calling `auto_debug_signals` on element with `debug_signal` will correctly call `debug_signal` method. This also break reference cycle, because we don't traverse `owner` in instances of `TransactionBase`

Arusekk (approved):
Noice.

lekcyjna123:
I have found another reference cycle. This time in Amaranth, which use MustUse class which point on it self to check if Elaboratables are used. So I introduced following changes:
- added support for Mapping
- added visited set to break dependecy cycles
- added ignoring of private fields to get pretty print (to don't print _MustUse_context)

I acknowledge that last point can be controversial, so I am ready to remove it

coreblocks/utils/debug_signals.py
```diff
@@ -1,8 +1,10 @@
 from amaranth import *
 from ._typing import SignalBundle, HasDebugSignals
+from collections.abc import Collection, Mapping
+import coreblocks.transactions.core
 
 
-def auto_debug_signals(thing) -> SignalBundle:
+def auto_debug_signals(thing, *, visited: set = set()) -> SignalBundle:
```
Arusekk, line 7: Note this is a potential footgun, because this function is not reentrant (the set is instantiated once on function definition).

coreblocks/utils/debug_signals.py
```diff
@@ -14,19 +16,51 @@ def auto_debug_signals(thing) -> SignalBundle:
...
-            smap[v] = a.debug_signals()
-        elif isinstance(a, Elaboratable):
-            smap[v] = auto_debug_signals(a)
+
+        # Check for reference cycles e.g. Amaranth's MustUse
+        if id(a) in visited:
+            continue
+        visited.add(id(a))
+
+        # ignore private fields (mostly to ignore _MustUse_context to get preaty print)
```
Arusekk, line 35:
```suggestion
        # ignore private fields (mostly to ignore _MustUse_context to get pretty print)
```

Arusekk (approved):

coreblocks/utils/debug_signals.py
lekcyjna123: Added docstring.

coreblocks/utils/debug_signals.py
```diff
@@ -1,32 +1,71 @@
...
             slist.append(a)
+        elif isinstance(a, Mapping):
+            submap = {}
+            for i, e in a.items():
+                sublist = auto_debug_signals(e, visited=visited)
+                if sublist:
+                    submap[f"{v}[{i}]"] = sublist
+            if submap:
+                smap[v] = submap
+        elif isinstance(a, Collection):
```
tilk, line 61: Maybe check if `a` is not a `str`? A silly special case, I know, but...

Arusekk (changes requested):

coreblocks/utils/debug_signals.py
```diff
@@ -11,22 +13,65 @@ def auto_debug_signals(thing) -> SignalBundle:
...
+
+            # Check for reference cycles e.g. Amaranth's MustUse
+            if id(a) in _visited:
+                continue
+            _visited.add(id(a))
+
+            # ignore private fields (mostly to ignore _MustUse_context to get pretty print)
+            if v[0] == "_":
+                continue
+
```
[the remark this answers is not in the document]
lekcyjna123: Done

coreblocks/utils/debug_signals.py
```diff
@@ -11,22 +13,65 @@ def auto_debug_signals(thing) -> SignalBundle:
...
+                    sublist = auto_debug_signals_internal(e, _visited=_visited)
+                    if sublist:
+                        submap[f"{v}[{i}]"] = sublist
+                if submap:
+                    smap[v] = submap
+            # handle strings separately so that we don't get into infinity recursion because of
+            # string being `Collection`s
+            elif isinstance(a, str):
+                continue
+            elif isinstance(a, Collection):
```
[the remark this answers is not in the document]
lekcyjna123: Function has been rewritten

Arusekk (approved):

tilk (approved):
Let's finally merge this, hopefully it will work well.

PR #271, opened 2023-03-30
Make the fetch unit use instruction cache

xThaid:
I refactored the fetch unit so it uses the instruction cache.

Depends on #258


coreblocks/frontend/fetch.py
```diff
@@ -11,58 +11,71 @@ class Fetch(Elaboratable):
...
+        self.cache_req = cache_req
+        self.cache_resp = cache_resp
         self.cont = cont
 
-        self.pc = Signal(gen_params.isa.xlen, reset=gen_params.start_pc)
-        self.halt_pc = Signal(bus.wb_params.addr_width, reset=2**bus.wb_params.addr_width - 1)
-
         self.verify_branch = Method(i=self.gp.get(FetchLayouts).branch_verify)
 
+        # Signals used for debugging purposes
```
xThaid, line 32: Do we actually want to keep them?

coreblocks/frontend/icache.py
```diff
@@ -96,7 +96,7 @@ def elaborate(self, platform) -> Module:
         flush_start = Signal()
         flush_finish = Signal()
 
-        with m.FSM() as fsm:
+        with m.FSM(reset="LOOKUP") as fsm:
```
tilk, line 99: Shouldn't this be in #258?

tilk, line 99: Shouldn't this be in #258?

tilk (approved):
Looks okay.

lekcyjna123:
You have completely rewritten Fetch unit. Instead of that I would rather have a new unit, so that we can choose on generation time if we want to use cache or connect core directly to memory.

tilk:
> You have completely rewritten Fetch unit. Instead of that I would rather have a new unit, so that we can choose on generation time if we want to use cache or connect core directly to memory.

Even though we would like to have at least a small instruction cache for most cases, I agree, it's more fitting for our philosophy to have alternative fetch units.

xThaid:
> You have completely rewritten Fetch unit. Instead of that I would rather have a new unit, so that we can choose on generation time if we want to use cache or connect core directly to memory.

> Even though we would like to have at least a small instruction cache for most cases, I agree, it's more fitting for our philosophy to have alternative fetch units.

If we want to switch between big modules (like the fetch unit), shouldn't we have more strict interfaces? Right now, the fetch unit (and other big modules) are loosely defined, i.e. attributes are accessed throughout the codebase and we don't really specify what should be the names. Maybe we can use Python Abstract Base classes.

lekcyjna123:
As for know we don't use ABC for interface in FUs and they all have the same interface, but I think that it is a good idea, to somehow ensure, that interface is correct on each HW block.

tilk:
> As for know we don't use ABC for interface in FUs

But we (currently) use protocols for type checking. I'm still torn between ABCs and protocols here: ABCs are more explicit (you need to declare something is an instance of a given ABC), while protocols are more implicit.

tilk:
Branch probably requires a `rebase --onto`.

tilk:
I think that the most probable reason for the tests failing is that the sizes of the memory regions need to be rounded up.

coreblocks/frontend/fetch.py
```diff
@@ -11,59 +11,77 @@ class Fetch(Elaboratable):
...
+            # but this could potentially change in the future since there's a reserved
+            # (currently unused) bit pattern in the spec, see table 19.1 in RISC-V spec v2.2
+            is_branch = res.instr[4:7] == 0b110
+            is_system = res.instr[2:7] == 0b11100
+
+            with m.If(discard_next):
+                m.d.sync += discard_next.eq(0)
+
+            # TODO: find a better way to fail when there's a fetch error.
+            with m.Elif(res.error == 0):
```
Kristopher38, line 73: Really nitpicking but could you remove that empty line there? You had me thinking that `Elif` was actually a separate `If` for a while

test/frontend/test_fetch.py
```diff
@@ -49,37 +43,49 @@ def elaborate(self, platform):
...
-                next_pc = rand.randint(0, (2**self.gp.isa.ilen - 1)) & ~0b11
+                # randomize being a branch instruction
+                if is_branch:
+                    data |= 0b1100000
+
+                output_q.append({"instr": data, "error": 0})
+
+                # Speculative fetch. Skip
+                if addr != next_pc:
+                    continue
```
Kristopher38, line 83: I don't fully understand this part yet. Can this condition happen in the current core implementation?

test/test_core.py
```diff
@@ -178,13 +182,9 @@ def test_simple(self):
 
 class TestCoreRandomized(TestCoreBase):
     def randomized_input(self):
-        halt_pc = len(self.instr_mem) * self.gp.isa.ilen_bytes
-
-        # set PC to halt at specific instruction (numbered from 0)
-        yield self.m.core.fetch.halt_pc.eq(halt_pc)
-
+        infloop_addr = (len(self.instr_mem) - 1) * 4
```
Kristopher38, line 185: Why `self.gp.isa.ilen_bytes` got changed to `4`?

test/test_core.py
```diff
@@ -226,7 +226,10 @@ def test_randomized(self):
         self.software_core.execute(init_instr_list)
         self.software_core.execute(instr_list)
 
-        self.instr_mem = list(map(lambda x: x.encode(), init_instr_list + instr_list))
+        # We add JAL instruction at the end to effictively create a infinite loop at the end of the program.
```
Kristopher38, line 229: typo - `effectively`

test/test_core.py
```diff
@@ -178,13 +182,9 @@ def test_simple(self):
...
     def randomized_input(self):
-        halt_pc = len(self.instr_mem) * self.gp.isa.ilen_bytes
-
-        # set PC to halt at specific instruction (numbered from 0)
-        yield self.m.core.fetch.halt_pc.eq(halt_pc)
-
+        infloop_addr = (len(self.instr_mem) - 1) * 4
         # wait for PC to go past all instruction
-        while (yield self.m.core.fetch.pc) < halt_pc:
+        while (yield self.m.core.fetch.pc) != infloop_addr:
```
Kristopher38, line 187: I see a bug here - later in the code you build an instruction sequence of *some* length, supply it to `TestElaboratable` and then align it to the length of the cache line. If the last JAL instruction doesn't end up at `infloop_addr` you'll loop forever (until simulation times out).

We had plans to refactor how these tests behave (`halt_pc` was a temporary solution) - the way we envisioned it was we would monitor some register dedicated to signaling test finish by writing nonzero value to it, or something along those lines. You can choose whether or not to implement it in this PR.

Kristopher38 (changes requested):

test/frontend/test_fetch.py
xThaid: Yes, the fetch unit speculatively fetches instructions, which may not end up executed.

test/test_core.py
xThaid: `self.gp.isa.ilen_bytes` defines the maximum instruction width, so using that variable here is definitely incorrect. And since in this particular test we use only instruction of size 4 bytes, I hardcoded 4 here.

test/test_core.py
xThaid: Alignment is done in `TestElaboratable`, so `infloop_addr` is calculated using the unaligned value. I do agree however that the current way of signaling the end of the test is not perfect. But I will leave it to another PR.

xThaid:
I have found the bug -- the cache wasn't cleaned on reset.

- Under Icarus Verilog, all tests (so in particular the first one as well) failed, so perhaps iverilog doesn't initialize memory and propagates uninitialized signals?
- Under Verilator, the first test passed, but the following ones failed, as the data in the cache was outdated.

xThaid:
Now the cache can be turned off using the configuration. Some of the changes are in the separate PR #307

coreblocks/frontend/fetch.py
```diff
@@ -11,59 +12,71 @@ class Fetch(Elaboratable):
...
+            res = self.icache.accept_res(m)
+
+            # bits 4:7 currently are enough to uniquely distinguish jumps and branches,
+            # but this could potentially change in the future since there's a reserved
+            # (currently unused) bit pattern in the spec, see table 19.1 in RISC-V spec v2.2
+            is_branch = res.instr[4:7] == 0b110
+            is_system = res.instr[2:7] == 0b11100
+
+            with m.If(discard_next):
+                m.d.sync += discard_next.eq(0)
```
Kristopher38, line 66: I looks like `discard_next` is unused (?)

Kristopher38 (approved):

Issue #262, opened 2023-03-23
Pass `self` variable to `condition` lambda in `def_class_method_mock`


xThaid:
`def_class_method_mock` decorator is used in test benches to mock a method. It takes an argument `condition`, a lambda returning a flag denoting whether the method is ready to run. The problem is that this lambda is expected not to take any arguments, but `def_class_method_mock` is meant to be used to wrap methods, so a variable `self` would be very useful (to access objects variables).

xThaid:
Fixed by #272

PR #272, opened 2023-03-31
Class method alternative approach

Arusekk:
BTW Python is object-oriented, so OOP >>>>> nested lambdas functional stuff :wink:

Depends on https://github.com/kuznia-rdzeni/coreblocks/pull/263
See https://github.com/kuznia-rdzeni/coreblocks/pull/265

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -293,7 +293,7 @@ class TestbenchIO(Elaboratable):
 
 
 def def_method_mock(
-    tb_getter: Callable[[], TestbenchIO], **kwargs
+    tb_getter: Callable[[], TestbenchIO] | Callable[[Any], TestbenchIO], **kwargs
 ) -> Callable[[Callable[..., Optional[RecordIntDict]]], Callable[[], TestGen[None]]]:
     """
     Decorator function to create method mock handlers. It should be applied on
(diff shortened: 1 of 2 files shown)


test/common.py
```diff
@@ -327,68 +328,41 @@ def process(arg):
...
+        self.kwargs = kwargs
+
+    def __call__(self, func: Callable[..., Optional[RecordIntDict]]) -> Callable[[], TestGen[None]]:
         @functools.wraps(func)
-        def mock(self) -> TestGen[None]:
-            tb = tb_getter(self)
+        def mock(func_self=None, /) -> TestGen[None]:
+            f = func
+            tb_getter = self.tb_getter
+            if func_self is not None:
```
xThaid, line 357: It'd be useful to wrap other lambdas as weel. For example `condition` lambda.

xThaid: That would effectively solve issue #262

tilk: I'd leave this to another PR, this one is pretty clear and uncontroversial (other than the use of `class` instead of `def` XD)

tilk:
If you're force-pushing anyway, why not do a proper rebase? ;)

Arusekk:
Done. 😄

xThaid:
I don't really know why it works, but I like the result of the change -- helps writing tests more efficiently.

xThaid (approved):

tilk (approved):

PR #273, opened 2023-03-31
DependencyManager: ListKey, key locking

piotro888:
Adds a few of useful features to `Dependency` system.
Partially exported from (and *dependency* for) #267.

* `ListKey` - New, very straight-forward key
* `dependency_exitsts` method - as the name says, needed to avoid `dependency not provided` runtime error
* key locking - New mechanism for locking dependency keys after `get_dependency`. Very useful to catch bugs when new dependencies are added to key, which was already read an processed because of some declaration order. There aren't now any use cases for adding after get, but there is a parameter `lock_on_get` in `DependencyKey` that can be overriden in specific key classes to disable this behavior if needed


tilk:
Is `dependency_exists` really needed? I checked both uses you have, and both can be easily eliminated:

* One use is subsumed by key locking.
* One use is eager checking for duplicates, which can be easily replaced by lazy checking (after `get_dependency`). Downsides of your current solution:
  - Iterates over the whole CSR list for every CSR added - quadratic-time behavior.
  - Incompatible with key locking as implemented here.

piotro888:
I guess the name was  misleading (named because of `self.dependencies`), changed to `key_exists`.

* If I understand your comment correctly, second use case is in the `CSRRegister`, but it will be eliminated by key locking and was not meant for duplicate checking. It was there to check if get_dependency()[-1] exists. This block of code will be entirely eliminated, but I haven't updated it yet.
* Use case in `CSRUnit` is still necessary. In case when no `CSRRegisters` are synthetised and no entries are added to `CSRListKey()`, calling `get_dependency` will throw `KeyError` that key was not provided as dependency. This function is used simply to skip calling `get_dependency`. This is a special case for CSR use case and I don't think that adding `Optional` to `get_dependency` or providing  default key values would be good for overall dependency system.

tilk:
Maybe it's time to change perspective? Instead of delegating checking to the user, handle it in the `DependencyManager`? For the `ListKey`, empty list return is completely fine. So maybe add an additional attribute to the keys, which decides if the "key is empty" situation is okay, and handle it in `get_dependency`?

lekcyjna123 (approved):

PR #274, opened 2023-04-01
Bump linter versions

Arusekk:
For some reason, my distro-provided versions of packages (black 23.1.0) result in

    scripts/lint.sh format

causing such changes.  They do not look so bad IMO, but it is quite unsettling to work with 9 not formatted files.

Also included is a fix for a typo causing my pyright 1.1.282 to throw errors.

I am only not sure about changes to test/transactions/test_assign.py (pep8-naming==0.13.3 and flake8==6.0.0 do not like variables named lowercase L),
because I realised that the file is straight up unreadable to me, the one-letter field/variable names look way more like obfuscated code than example names.  My suggestion would be field1, field2, field3 instead of a, b, c.  WDYT?

### DIFF
--- a/coreblocks/fu/mul_unit.py
+++ b/coreblocks/fu/mul_unit.py
@@ -180,7 +180,6 @@ class MulUnit(Elaboratable):
 
         @def_method(m, self.issue)
         def _(arg):
-
             m.d.comb += decoder.exec_fn.eq(arg.exec_fn)
             i1, i2 = get_input(arg)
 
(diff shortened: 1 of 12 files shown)


tilk:
The recommended development environment for Coreblocks is a venv, with precise versions from `requirements-dev.txt` installed. If you want to introduce changes required for new versions of pyright and linters, it might be a good idea to bump the versions in the requirements.

Also, I agree that readability of `test_assign` is low, but it's also very compact, testing a lot of variants using a small amount of code. Changing `l` to `lay`, or maybe even `layout`, seems sensible. Using `field1`, `field2`... instead of `a`, `b`... doesn't really help, the verbosity is increased without any comprehensibility gains (one could even argue that comprehensibility is lost, as, at least for me, letters are easier to distinguish than numbers.)

lekcyjna123 (approved):

Arusekk:
Notable changes include:
- PyRight since 1.1.296 (Feb 28) now supports *running offline* 😮 
- PyRight detects more interface mismatches
- Flake8 detects more misleading names
- Black detects more redundant parentheses

tilk (approved):

Issue #223, opened 2023-02-02
Connect Coreblocks to LiteX


tilk:
The task is to connect our core to the [LiteX framework](https://github.com/enjoy-digital/litex). This involves creating a `pythondata-cpu-coreblocks` module, which can then be used from inside LiteX. For example, see [pythondata-cpu-minerva](https://github.com/litex-hub/pythondata-cpu-minerva) and other existing modules.

piotro888:
https://github.com/kuznia-rdzeni/litex and https://github.com/kuznia-rdzeni/pythondata-cpu-coreblocks

PR #275, opened 2023-04-01
Core generation configurations

piotro888:
First step #223 LiteX support.

It would make sense to supply some set of fixed selectable configurations to work with LiteX.
Good LiteX flow (inspired by minerva) works by LiteX soc/cpu/core.py file calling generate verilog script from our repo with configurable CLI params, dependent on litex fixed core versions, and then verilog file added as source.
Other CPUs use only one fixed config or modifies each generation parameters via CLI options, but as we support much deeper customization with `GenParams`, it would make sense to provide some basic set of different GenParams options.

Currently PR adds set of fixed GenParams arguments and option to select them with gen_verilog script (they could be collected better with name in base class, but for now it is draft).
That's the most straightforward way to do that, but as number of parameters and complexity would increase, what do you think of making `GenParams` accepting simple configuration class (like this `CoreConfiguration`) instead of parameters? For most cases in tests default configuration like `BasicCoreConfig` would be used. This is analogical to passing our current `basic_configuration` for Functional Units, so maybe that idea should be applied to all parameters? That would also generalize core synthesis in a few of our scripts (and with externally selectable options).
In other case we would need to keep that separate class for fixed configs parameters in sync with GenParams changes and duplicate default values.
I think that this is good interface to export custom cores by adding new fixed config to configurations file instead of changing GenParams options is the middle of gen_verilog scirpt (and it would be nice if this interface was consistent and also used with our internal gen params options)


xThaid:
I agree that we should keep a few pre-defined core parameters. I am feeling like Builder pattern may be useful here. We would have a few default builders, which could be easily overridden if someone wants for example a bigger cache.
Code to show the idea:
```python
cb = CoreBuilder.tiny()

if something:
    cb.setICacheLineSize(32)

cb.gen()
```

and `cb.gen()` could either return `GenParams` or `Core` module.

coreblocks/params/genparams.py
```diff
@@ -4,7 +4,11 @@
 from typing import TypeVar, Type, Protocol, runtime_checkable
 
 from .isa import ISA
-from .fu_params import BlockComponentParams
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .configurations import CoreConfiguration
```
tilk, line 11: Is there a cyclic dependency? If so, can it be removed?

coreblocks/params/configurations.py
```diff
@@ -8,3 +9,64 @@
...
+    phys_regs_bits: int
+        Size of the Physical Register File is 2**phys_regs_bits.
+    rob_entries_bits: int
+        Size of the Reorder Buffer is 2**rob_entries_bits.
+    start_pc: int
+        Initial Program Counter value.
+    """
+
+    isa_str: str = "rv32i"
+    func_units_config: list[BlockComponentParams] = field(default_factory=lambda: basic_configuration)
```
tilk, line 37: Why the `default_factory`? Is this because of the risk of mutation? Then:
1. This doesn't fix all mutation problems. For example, the `RSBlockComponent` inside `basic_configuration` can be mutated.
2. You don't actually mutate `func_units_config` anywhere in your code.
3. I much prefer the immutable data structure way instead of fighting mutation. For example, you could change the `list` type to `Collection` to block mutation in typing, or even use tuples instead of lists to block mutation in runtime (but I believe this is overkill).

coreblocks/params/configurations.py
```diff
@@ -8,3 +9,64 @@
...
+class TinyCoreConfig(CoreConfiguration):
+    """Minmal core configuration."""
+
+    func_units_config: list[BlockComponentParams] = field(
+        default_factory=lambda: [
+            RSBlockComponent([ALUComponent(), JumpComponent()], rs_entries=2),
+            LSUBlockComponent(),
+        ]
+    )
+    rob_entries_bits: int = field(default=6)
```
tilk, line 61: Subclassing just to change defaults is super-weird to me. Why not have just:
```python
tiny_core_config = CoreConfiguration(rob_entries_bits=6, func_units_config=...)
```

test/scheduler/test_scheduler.py
```diff
@@ -109,7 +111,9 @@ class TestScheduler(TestCaseWithSimulator):
 
     def setUp(self):
         self.rs_count = len(self.optype_sets)
-        self.gen_params = self.gen_params = test_gen_params("rv32i", rs_block_number=self.rs_count)
+        self.gen_params = GenParams(
+            TestCoreConfig(func_units_config=[RSBlockComponent([], rs_entries=4) for _ in range(self.rs_count)])
```
tilk, line 115: Dataclasses support copying with changes. You could just use:
```python
self.gen_params = GenParams(
    dataclasses.replace(test_core_config,
        func_units_config=[RSBlockComponent([], rs_entries=4) for _ in range(self.rs_count)]))
```

tilk (commented):
I like the decision to use dataclasses, but I believe you are misusing them.

coreblocks/params/genparams.py
piotro888: Yes, there is a cyclic dependency and I don't think it can be avoided. `GenParams` needs to import `CoreConfiguration` for typing annotation, but `CoreConfigurations` provides some FU layouts, from which all elements import `GenParams`

coreblocks/params/configurations.py
piotro888: I don't worry about mutation, but that was the only way that `@dataclass` would compile with field with type of `list` :p.
Thanks, I didn't know about different behavior of `Collection`, I will check it and it would be nice if it would work directly

coreblocks/params/configurations.py
```diff
@@ -8,3 +9,64 @@
...
+    rob_entries_bits: int
+        Size of the Reorder Buffer is 2**rob_entries_bits.
+    start_pc: int
+        Initial Program Counter value.
+    """
+
+    isa_str: str = "rv32i"
+    func_units_config: list[BlockComponentParams] = field(default_factory=lambda: basic_configuration)
+
+    phys_regs_bits: int = field(default=6)
```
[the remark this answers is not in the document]
piotro888: Motivation behind that change is in the commit comment. It can be removed if we would not use subclassing.
I added it to raise error when some one would override field in child class but forget to add type annotation. Dataclass would then silently ignore that field and use value from parent class instead. I think that this is a possible weird bug-source. By using `field()` it would raise error in that case. But if we would get rid of subclassing that's no longer an issue, as ignoring fields would always create error of non existing field in `GenParams`.

coreblocks/params/configurations.py
piotro888: I decided to use subclassing, because it allowed to use that nice kwargs constructor override for all sub-configurations (and I thought that it could be implemented more directly without that workarounds) 
But after writing that I see that's a misuse of dataclasses and can be confusing in definition, so maybe it is better to avoid that?
Downside of it would be less nice field overriding in tests, but that would  at lest use standard and expected way of overriding dataclases (and I didn't know about `dataclasses.replace` before).

test/scheduler/test_scheduler.py
piotro888: In comment for review above

coreblocks/params/configurations.py
tilk: So `dataclass` is the entity avoiding mutation problems here ;) The static type is not the problem here, the dynamic one is. Use a tuple.

test/scheduler/test_scheduler.py
tilk: If you want nicer overriding, you could add a method for that, for example:

```python
def replace(self, **kwargs):
    return dataclasses.replace(self, **kwargs)
```

You could even override `__call__`, although this would probably be less readable.

piotro888: Nice idea, thanks! 

BTW, unfortunately both new ways are too confusing for language serves to propose valid dataclass parameters/fields (but that's only optional convince)

coreblocks/params/configurations.py
piotro888: done

test/scheduler/test_wakeup_select.py
```diff
@@ -42,7 +45,10 @@ def elaborate(self, platform):
 
 class TestWakeupSelect(TestCaseWithSimulator):
     def setUp(self):
-        self.gen = test_gen_params("rv32i", rs_entries=16)
+        dataclasses.replace(test_core_config, rob_entries_bits=3)
+        self.gen = GenParams(
+            test_core_config.replace(func_units_config=tuple(RSBlockComponent([], rs_entries=4) for _ in range(16)))
```
lekcyjna123, line 50: You have added here more RS than it was in original version. On the other hand, your RS have only 4 entries instead of 16.

piotro888: Thanks for catching that! Line above it was also accidentally commited leftover from my testing, sorry.

coreblocks/params/configurations.py
```diff
@@ -1,10 +1,63 @@
...
+    phys_regs_bits: int
+        Size of the Physical Register File is 2**phys_regs_bits.
+    rob_entries_bits: int
+        Size of the Reorder Buffer is 2**rob_entries_bits.
+    start_pc: int
+        Initial Program Counter value.
+    """
+
+    isa_str: str = "rv32i"
+    func_units_config: tuple[BlockComponentParams, ...] = basic_configuration
```
tilk, line 36: Could you use `Collection` or `Sequence` as the field type instead of tuples?

coreblocks/params/configurations.py
```diff
@@ -1,10 +1,63 @@
...
+@dataclass(kw_only=True)
+class CoreConfiguration:
+    """
+    Core configuration parameters.
+
+    Parameters
+    ----------
+    isa_str: str
+        RISCV ISA string. Examples: "rv32i", "rv32izicsr". Used for instruction decoding.
+    func_units_config: list[BlockComponentParams]
```
tilk, line 24: Also update the type info in the docstring.

tilk (approved):

PR #276, opened 2023-04-04
Nestable transactions

tilk:
This PR introduces the ability to nest transaction/method bodies inside other transaction/method bodies. Nested transactions/methods are available only if the parent transaction/method is active (run/called). Nesting introduces a scheduling order dependency.

Please note that if the nested transaction fails to run doesn't prevent the parent transaction from running. Dependencies like this might be useful (for example in order to have a one-of behavior, or to have a conditional call which doesn't block the target method), but I believe they require functionality like in #186.

To do:

- [x] Write tests.
- [x] Document.

### DIFF
--- a/coreblocks/transactions/core.py
+++ b/coreblocks/transactions/core.py
@@ -366,7 +366,7 @@ class _TransactionBaseStatements:
 
 
 class TransactionBase(Owned):
-    stack: ClassVar[list["TransactionBase"]] = []
+    stack: ClassVar[list[Union["Transaction", "Method"]]] = []
     comb: ClassVar[_TransactionBaseStatements] = _TransactionBaseStatements()
 
     def __init__(self):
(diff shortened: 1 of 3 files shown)


lekcyjna123 (commented):
If I understand correctly, nested transaction is under `with m.If(self.grant)` of parent transaction, so as long as parent is not granted, `request` signal of the child have default value `0`. If yes, then as for now I don't see any problems.

lekcyjna123 (approved):

xThaid (changes requested):
This extension allows to define a whole range of incorrect transactions, for example:
```python
  with Transaction().body(m):
      method(m)
      with Transaction().body(m):
          method(m)
```
This one is obvious, but I can imagine that in the future in multi-nested transactions and methods, the error can be more subtle. It would be very useful to detect such erroneous transactions (right now, the previous code just deadlocks).

tilk:
Yes, this is true, but notice that only thing this PR does is introduce syntactic sugar for things you can already express. In particular, your example is:
```python
with (t1 := Transaction()).body(m):
    method(m)
with (t2 := Transaction()).body(m, request=t1.grant):
    method(m)
t1.schedule_before(t2)
```
And by the way, this code shouldn't deadlock *completely*, the outer transaction `t2` can run fine - just the inner one can't, as it always conflicts with the outer one.

Of course, one can argue that by making things like this easier to express, you encourage writing them, and therefore the need to check for issues like this is greater.

xThaid:
The problem is that it does deadlock. Consider the following test case:
```python
class DeadlockTestCircuit(Elaboratable):
        m = Module()
        tm = TransactionModule(m)

        method = Method(i=[("n", 1)])

        self.arg = Signal()

        @def_method(m, method)
        def _(n):
            m.d.comb += self.arg.eq(n)

        with tm.transaction_context():
            with Transaction().body(m):
                method(m, n=C(0))

                with Transaction().body(m):
                    method(m, n=C(1))

        # so that Amaranth allows us to use add_clock
        dummy = Signal()
        m.d.sync += dummy.eq(1)

        return tm


class TestDeadlock(TestCaseWithSimulator):
    def setUp(self):
        random.seed(42)

    def test_scheduling(self):
        m = DeadlockTestCircuit()

        def process():
            for _ in range(10):
                yield
                print((yield m.arg))

        with self.run_simulation(m) as sim:
            sim.add_sync_process(process)
```
The simulation doesn't proceed even to the first cycle.

tilk:
Thanks for the test case! It looks like the issue needs to be investigated before moving foward. The simulation deadlock is a bad sign.

tilk:
@xThaid: it turns out this was an earlier bug (#324).

xThaid (approved):
With the bug fixed, we are good to go.

PR #277, opened 2023-04-04
Refactor dependency management and fu config

tilk:
This PR contains three changes.

1. Separation of the `DependencyManager` stuff to a separate file.
2. Use of frozen dataclasses for `*ComponentParams`. This can play nice with #275. (I didn't make a dataclass out of attribute-less `*Params` classes. Should I?)
3. Removal of `from __future__ import annotations`. Its future (ahem) is uncertain, see #172.

### DIFF
--- a/coreblocks/fu/mul_unit.py
+++ b/coreblocks/fu/mul_unit.py
@@ -1,6 +1,6 @@
 from enum import IntFlag, unique, IntEnum
 from typing import Tuple
-from dataclasses import dataclass
+from dataclasses import KW_ONLY, dataclass
 
 from amaranth import *
 
@@ -248,6 +248,7 @@ class MulUnit(Elaboratable):
 @dataclass(frozen=True)
 class MulComponent(FunctionalComponentParams):
     mul_unit_type: MulType
+    _: KW_ONLY
     dsp_width: int = 32
 
     def get_module(self, gen_params: GenParams) -> FuncUnit:
--- a/coreblocks/fu/mul_unit.py
+++ b/coreblocks/fu/mul_unit.py
@@ -1,5 +1,6 @@
 from enum import IntFlag, unique, IntEnum
 from typing import Tuple
+from dataclasses import dataclass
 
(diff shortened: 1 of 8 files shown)


coreblocks/fu/mul_unit.py
```diff
@@ -244,10 +245,10 @@ def _(arg):
         return m
 
 
+@dataclass(frozen=True)
 class MulComponent(FunctionalComponentParams):
-    def __init__(self, mul_unit_type: MulType, *, dsp_width: int = 32):
-        self.mul_unit_type = mul_unit_type
-        self.dsp_width = dsp_width
+    mul_unit_type: MulType
+    dsp_width: int = 32
```
piotro888, line 251: What would you think about changing `dataclass` fields to 
```python
mul_unit_type: MulType
_: KW_ONLY
dsp_width: int = 32
```
?
That would preserve the original constructor with `(mul_type, *, dsp_width=32)` and could be helpful if there would be more fields/paramters with default values, and we would like to override one of them.

lekcyjna123 (commented):
For me it looks good. I think that suggestion form piotro888 is good, so i will wait with approve until implemented.

coreblocks/fu/mul_unit.py
tilk: > and could be helpful if there would be more fields/paramters with default values, and we would like to override one of them.

There seems to be much confusion about argument types in Python. Nothing stops you from supplying a value for an argument by keyword if it's not keyword-only. Keyword-only means "positional use is disallowed", not "keyword use is enabled".

Anyway, positional use of `dsp_width` is unreadable, so not a bad suggestion.

piotro888: Good to know; but I agree that use of `dsp` parameter shouldn't be positional

piotro888 (approved):
Nice changes! Good to see that extended use of dataclasses for parameters

lekcyjna123 (approved):

PR #278, opened 2023-04-04
Add a function aligning numbers and a function for yielding for many cycles

xThaid:
Nothing controversial (I hope). I use these function in PR #258 and #271 , but as they aren't really related to the instruction cache, here they are.


coreblocks/utils/comb.py
```diff
@@ -0,0 +1,11 @@
+from amaranth import *
+
+
+def rotate_left(sig: Value, shift: int = 1) -> Value:
```
lekcyjna123, line 4: These two functions are already by default in amaranth (as a method in Value class)

test/utils/test_comb.py
```diff
@@ -0,0 +1,56 @@
+from amaranth import *
+from amaranth.sim import *
+
+from coreblocks.utils import rotate_left, rotate_right
+
+from test.common import TestCaseWithSimulator
+
+
+class CircularShiftCircuit(Elaboratable):
```
lekcyjna123, line 9: I am not sure if this shouldn't be moved to coreblocks/utils.

tilk: Isn't this basically a test circuit? I don't see a practical need for generating all possible rotations at once.

xThaid: It was just a test circuit

tilk (approved):

lekcyjna123 (approved):

PR #279, opened 2023-04-07
Fix FreeRF FIFO initialization

piotro888:
This is a fix for issues reported in #268. Let's hope all tests will pass now!
The issue was that `init` parameter from Amaranth's `Memory` initializes memory to `init` only on full power cycle instead of reset (unlike everything else).  From docs: `Initial values. At power on, each storage element in this memory is initialized to the corresponding element of init`. Cocotb tests were the first test to actually use the reset signal. (debugging log in teams channel)

This PR initializes FreeRF with simple transaction that runs on reset and removes misleading init parameter from `BasicFIFO`.
Possible improvements (let me know what you think):
* `Retirement` Transcation could be `scheduled_before` this init transaction. This way it would not delay core by blocking retirement, and supply new regs when they are needed (and we can see that few cycles of delay in fibonacci mem test). That would involve passing retirement transaction from `elaborate` stage some dependecy key (it cannot be declared in constructor), so I have mixed feelings about that for now
* Make a variant of BasicFIFO that uses `Array` of Signals instead of `Memory`, that can be `init`ed like we want

### DIFF
--- a/test/utils/test_fifo.py
+++ b/test/utils/test_fifo.py
@@ -11,15 +11,14 @@ import random
 
 
 class BasicFifoTestCircuit(Elaboratable):
-    def __init__(self, depth, init):
+    def __init__(self, depth):
         self.depth = depth
-        self.init = init
 
     def elaborate(self, platform):
         m = Module()
         tm = TransactionModule(m)
 
-        m.submodules.fifo = self.fifo = BasicFifo(layout=data_layout(8), depth=self.depth, init=self.init)
+        m.submodules.fifo = self.fifo = BasicFifo(layout=data_layout(8), depth=self.depth)
 
         m.submodules.fifo_read = self.fifo_read = TestbenchIO(AdapterTrans(self.fifo.read))
         m.submodules.fifo_write = self.fifo_write = TestbenchIO(AdapterTrans(self.fifo.write))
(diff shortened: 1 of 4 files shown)


test/utils/test_fifo.py
```diff
@@ -29,25 +28,18 @@ def elaborate(self, platform):
...
     depth: int
-    init_len: int
 
     def test_randomized(self):
-        init_values = [x for x in range(self.init_len)]
-
-        fifoc = BasicFifoTestCircuit(depth=self.depth, init=init_values)
-        expq = deque(reversed(init_values))  # first expected element is at the start of init_list
+        fifoc = BasicFifoTestCircuit(depth=self.depth)
+        expq = deque()  # first expected element is at the start of init_list
```
tilk, line 42: The comment is outdated.

test/utils/test_fifo.py
```diff
@@ -29,25 +28,18 @@ def elaborate(self, platform):
...
+    ("name", "depth"),
     [
-        ("notpower_notfull", 5, 3),
-        ("notpower_full", 5, 5),
-        ("notpower_empty", 5, 0),
-        ("power_notfull", 4, 3),
-        ("power_full", 4, 4),
-        ("power_empty", 4, 0),
+        ("notpower", 5, 0),
+        ("power", 4, 0),
```
tilk, line 34: The third argument is not used anymore.

tilk (changes requested):

tilk (approved):

lekcyjna123 (approved):

PR #281, opened 2023-04-07
Add a sanity check for ordered method definitions

Arusekk:
We talked about this when discussing #187, but I thought it would be harder.

### DIFF
--- a/coreblocks/transactions/core.py
+++ b/coreblocks/transactions/core.py
@@ -8,6 +8,7 @@ from typing_extensions import Self
 from amaranth import *
 from amaranth import tracer
 from amaranth.hdl.ast import Statement
+from itertools import count
 
 from coreblocks.utils import AssignType, assign
 from ._utils import *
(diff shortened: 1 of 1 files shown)


coreblocks/transactions/core.py
```diff
@@ -366,6 +366,7 @@ def clear(self):
 class TransactionBase(Owned):
     current: ClassVar[Optional["TransactionBase"]] = None
     comb: ClassVar[_TransactionBaseStatements] = _TransactionBaseStatements()
+    defined: bool = False
```
lekcyjna123, line 369: After introducing of that field, is the initialisation of `defined` field in `Method` class still needed?

tilk (commented):
I have mixed feelings about this change. On the one hand, this is simple and short. On the other:

1. This relies on `schedule_before` being called before the method definition. Currently this is not required.
2. This relies on an implementation detail -- the fact that if `conflict` is `False`, then `priority` is `Priority.RIGHT`.
3. The check is done only for methods, not for transactions.

lekcyjna123 (approved):

tilk (approved):

PR #282, opened 2023-04-08
Decoder: set invalid registers to 0 and fix CSR decoding

piotro888:
@tilk another fix for #268 .

Decoder produced `rl_s1_v, rl_s2_v and rl_rd_v` fields but they were not used anywehere.
It caused allocation of random registers (only a waste of resources, not breaking things), but Jump/Branch unit was the first FU that sometimes should produc results (AUIPC) and sometimes not (but according to FU convention it always pushed this result). In this case computed branch result was commited to random register (even if it was not valid like in BGEU).

This PR muxes invalid registers to 0 in decoder, so they have no effect on core (writes ignored, allocation ignored, reads ignored).

I also removed the ` rl_**_v` field from layout, which was pushed through whole core, because I don't feel like it will be useful anywhere. Functional Units only use source fields to computation based on FUs internal decoders. With this PR, RSes would not wait for 0 register (and I see no point in pushing random registers values through core), and in case of J/B unit produced result would be ignored if rd is not valid. 

Edit: Also fixes CSR decoding issue found by tilk. I decided to split `CSR_REG` and `CSR_IMM` `OpTypes`, as only `CSR_IMM` needs special correction to standard instruction encoding (part of it was done in both cases, but invalidating rs1 was not). It is also useful in `CSRUnit`

### DIFF
--- a/test/frontend/test_decoder.py
+++ b/test/frontend/test_decoder.py
@@ -163,16 +163,16 @@ class TestDecoder(TestCaseWithSimulator):
             self.assertEqual((yield self.decoder.funct12_v), test.funct12 is not None)
 
             if test.rd is not None:
-                self.assertEqual((yield self.decoder.rd_v), 1)
                 self.assertEqual((yield self.decoder.rd), test.rd)
+            self.assertEqual((yield self.decoder.rd_v), test.rd is not None)
 
             if test.rs1 is not None:
-                self.assertEqual((yield self.decoder.rs1_v), 1)
                 self.assertEqual((yield self.decoder.rs1), test.rs1)
+            self.assertEqual((yield self.decoder.rs1_v), test.rs1 is not None)
 
             if test.rs2 is not None:
-                self.assertEqual((yield self.decoder.rs2_v), 1)
                 self.assertEqual((yield self.decoder.rs2), test.rs2)
+            self.assertEqual((yield self.decoder.rs2_v), test.rs2 is not None)
 
             if test.imm is not None:
                 self.assertEqual((yield self.decoder.imm), test.imm)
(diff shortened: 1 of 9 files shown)


tilk:
I also discovered this problem, but went deeper into fixing it, and discovered a problem with decoding CSR registers.

https://github.com/kuznia-rdzeni/coreblocks/pull/268/commits/21ecc666bca2df32fc7b53c738813d545886b9f9

tilk:
I think your solution looks nicer, but you could also change ‘test_decoder’ to check that the valid signal is 1 if and only if the register is expected (currently only the implication expected -> valid is checked).

tilk (approved):

test/frontend/test_decoder.py
```diff
@@ -163,16 +163,16 @@ def process():
             self.assertEqual((yield self.decoder.funct12_v), test.funct12 is not None)
 
             if test.rd is not None:
-                self.assertEqual((yield self.decoder.rd_v), 1)
                 self.assertEqual((yield self.decoder.rd), test.rd)
+            self.assertEqual((yield self.decoder.rd_v), test.rd is not None)
 
             if test.rs1 is not None:
                 self.assertEqual((yield self.decoder.rs1_v), 1)
-                self.assertEqual((yield self.decoder.rs1), test.rs1)
```
lekcyjna123, line 171: Typo. You have removed the test for correctness of value inside of `decoder.rs{1,2}`. Instead of that you check two times the valid bit.

lekcyjna123 (changes requested):

xThaid (approved):

test/frontend/test_decoder.py
```diff
@@ -163,16 +163,16 @@ def process():
...
+            self.assertEqual((yield self.decoder.rs1_v), test.rs1 is not None)
 
             if test.rs2 is not None:
                 self.assertEqual((yield self.decoder.rs2_v), 1)
-                self.assertEqual((yield self.decoder.rs2), test.rs2)
```
tilk, line 175: Same problem here.

tilk (changes requested):

test/frontend/test_decoder.py
piotro888: ahh, such a stupid typo. fixed

lekcyjna123 (approved):

tilk (approved):

PR #283, opened 2023-04-09
LSU addressing fix

tilk:
Found while doing #268. Wishbone is word addressed, yet the LSU still sent byte addresses (with lower bits being 0). We missed that during CR.

### DIFF
--- a/test/lsu/test_dummylsu.py
+++ b/test/lsu/test_dummylsu.py
@@ -118,10 +118,9 @@ class TestDummyLSULoads(TestCaseWithSimulator):
 
             self.announce_queue.append(ann_data)
             exec_fn = {"op_type": op[0], "funct3": op[1], "funct7": 0}
-            mask = shift_mask_based_on_addr(mask, addr)
 
-            # calculate aligned address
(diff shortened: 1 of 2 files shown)


test/lsu/test_dummylsu.py
```diff
@@ -289,11 +292,10 @@ def generate_instr(self, max_reg_val, max_imm_val):
...
 
             exec_fn = {"op_type": op[0], "funct3": op[1], "funct7": 0}
-            mask = shift_mask_based_on_addr(mask, addr)
 
-            # calculate aligned address
-            rest = addr % 4
-            addr = addr - rest
+            # calculate word address and mask
+            addr = addr >> 2
+            mask = shift_mask_based_on_addr(mask, addr)
```
xThaid, line 298: Shouldn't the mask be calculated using non-truncated address?

tilk: Yes.

lekcyjna123 (approved):

piotro888 (approved):

PR #291, opened 2023-04-17
Add multiplication tests (cocotb)

tilk:
This PR adds multiplication tests from the `riscv-tests` set.


piotro888 (approved):
LGTM, Interesting makefile feature with `define`

PR #293, opened 2023-04-18
Vector register

lekcyjna123:
After preparing high level schema of vector core I start the implementation with the implementation of very simple (and very not optimal) vector registers. Next step is to create a more "intelligent" implementation.
- [ ] Continuous vector register with tests
- [ ] Interlacing vector register with tests

### DIFF
--- a/coreblocks/fu/vector_unit/v_layouts.py
+++ b/coreblocks/fu/vector_unit/v_layouts.py
@@ -1,0 +1,7 @@
+from coreblocks.params.vector_params import VectorParameters
+
+class RegisterLayouts:
+    def __init__(self, gen_params: GenParams, v_params : VectorParameters):
+        self.read_resp = [
+            ("data", v_params.elen)
+        ]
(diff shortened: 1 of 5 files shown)


lekcyjna123:
I tried to implement inside VRRFragment a logic to make forwarding between reading methods e.g:
- Let assume that we have two ports (`p1` and `p2`)
- We get request to read from `v1` element on position `x` on both ports
- In simple implementation this will take two cycles (first cycle to read by `p1` and second to read by `p2`)
- I tried to implement forwarding, so in case when the same element is read `p1` read data from register and `p2` copy read data from `p1`

Problem which I hit is that register is synchronous memory, so I need to have request - response pattern and I have to synchronize responses to have correct data forwarding from `p1` to `p2` (both `p1` and `p2` responses have to be called in the same cycle). I was able to write implication that `p2` response is ready when `p1` is run, but I don't see how to do that in the other way (I thought about some helper structures e.g buffers/fifos).

As for now I plan to drop this forwarding functionality and it will have to be handled either by unit reading register or there will be an additional cycle of delay. The other possibility was to leave only this implication, but this would be error prone due to timings dependency (both methods run in the same cycle).

lekcyjna123:
Tomorrow I plan to add docstrings and merge it to development branch.

PR #294, opened 2023-04-21
Zbc functional unit

pa000:
Functional unit for Zbc extension[1][2]. Inspired by the multiplication FU. Currently uses a basic iterative algorithm.

[1]: https://en.wikipedia.org/wiki/Carry-less_product
[2]: https://github.com/riscv/riscv-bitmanip/releases/download/1.0.0/bitmanip-1.0.0-38-g865e7a7.pdf


coreblocks/fu/zbc.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def elaborate(self, platform):
+        m = Module()
+        output = Signal(unsigned(self.gen_params.isa.xlen * 2))
+
+        i1 = Signal(unsigned(self.gen_params.isa.xlen * 2))
+        i2 = Signal(unsigned(self.gen_params.isa.xlen))
+        busy = Signal()
+
+        @def_method(m, self.issue, ready=~busy)
```
Kristopher38, line 54: Both Zbc and ZbcUnit having a method interface is a precedent (usually only the core-facing part has it), but I guess this also looks good. I'm curious to see what others think.

test/fu/test_zbc.py
```diff
@@ -0,0 +1,64 @@
+from typing import Dict
+
+from coreblocks.fu.zbc import ZbcFn, ZbcComponent
+from coreblocks.params import *
+from coreblocks.params.configurations import test_core_config
+
+from test.fu.functional_common import GenericFunctionalTestUnit
+
+
+def clmul(i1: int, i2: int, xlen: int) -> int:
```
Kristopher38, line 10: perhaps it would be a good idea to include a link to the spec where the procedures below are outlined https://five-embeddev.com/riscv-bitmanip/draft/bitmanip.html#insns-clmul

Kristopher38 (commented):
Looks pretty solid to me

tilk:
I believe you didn't update the frontend decoder with the opcode->optype translation for `CLMUL`.

coreblocks/fu/zbc.py
pa000: I based it on how it's done in the multiplication unit but now I've removed the methods because zbc is not as complicated

coreblocks/fu/zbc.py
```diff
@@ -0,0 +1,170 @@
...
+            m.d.sync += i1.eq(arg.i1)
+            m.d.sync += i2.eq(arg.i2)
+            m.d.sync += busy.eq(1)
+
+        @def_method(m, self.accept, ready=~i2.bool() & busy)
+        def _():
+            m.d.sync += busy.eq(0)
+            return {"o": output}
+
+        with m.If(busy):
```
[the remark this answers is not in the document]
pa000: fixed, thanks

pa000:
I updated the frontend decoder and added support for parallel computation so now it's ready for review

lekcyjna123 (approved):

coreblocks/fu/zbc.py
```diff
@@ -0,0 +1,259 @@
...
+            ],
+            2,
+        )
+        m.submodules.decoder = decoder = ZbcFn.get_decoder(self.gen_params)
+        m.submodules.zbc = clmul = ClMultiplier(self.gen_params.isa.xlen, self.recursion_depth)
+
+        m.d.sync += clmul.reset.eq(0)
+        in_progress = Signal()
+
+        @def_method(m, self.accept, ready=~clmul.busy & ~clmul.reset & in_progress)
```
xThaid, line 194: The `ready` signal here (and for `issue` method) is not completely obvious for me. It seems that there is some redundancy here. 

If I understand correctly, in the current setting, there is no pipelining at all. That is, one cycle after the result is accepted, the next request can be issued (this is imposed by the `in_progress` flag). Therefore, the multiplier is idle for two cycles between requests. It is not necessarily a problem, but in this we can simplify it a bit:
1. Reduce the size of params FIFO to 1 (size 2 is used to do pipelining)
2. In `issue` method we don't need to specify `ready` signal. Readiness of `params_fifo.write` will be controlling if we can issue a request.
3. In `accept` method the ready signal can be as simple as `~clmul.busy`
4. We can remove `in_progress` flag completely.

coreblocks/fu/zbc.py
```diff
@@ -0,0 +1,259 @@
...
+    """
+
+    def __init__(self, bit_width: int, recursion_depth: int):
+        """
+        Parameters
+        ----------
+        bit_width: int
+            Bit width of inputs
+        recursion_depth: int
+            Depth of recursive submodules for parallel computation (assumes bit_width to be a power of 2)
```
tilk, line 62: Assumptions like this should be checked.

coreblocks/fu/zbc.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        if bit_width == 1 and recursion_depth != 0:
+            raise ValueError("Too large recursion depth")
```
tilk, line 65: This check should be done for other bit widths, too.

coreblocks/fu/zbc.py
```diff
@@ -0,0 +1,259 @@
...
+        self.i2 = Signal(unsigned(bit_width))
+        self.result = Signal(unsigned(bit_width * 2))
+        self.reset = Signal()
+        self.busy = Signal()
+
+    def elaborate(self, platform):
+        if self.recursion_depth == 0:
+            return self.iterative_module()
+        else:
+            return self.recursive_module()
```
tilk, line 80: If the recursion depth is such that it goes down to single bits, this module could be combinatorial, with 0 cycles of delay, which would allow the CPU to run one such multiplication per cycle. Currently this can't be realized, as the `iterative_module` needs to be always invoked, and it always introduces a single cycle of delay. I wonder if that's useful; but given that the `recursive_module` is pretty simple, it might be possible to do single-cycle multiplications like this without loss of Fmax.

But given that this is _an optimization_, this could be done later.

coreblocks/fu/zbc.py
```diff
@@ -0,0 +1,259 @@
...
+
+            result = Mux(params.high_res, output[xlen:], output[:xlen])
+            reversed_result = Mux(params.rev_res, result[::-1], result)
+
+            m.d.sync += in_progress.eq(0)
+
+            return {"rob_id": params.rob_id, "rp_dst": params.rp_dst, "result": reversed_result}
+
+        @def_method(m, self.issue, ready=~clmul.busy & ~in_progress)
+        def _(arg):
```
tilk, line 209: This uses an old convention, we currently prefer to name arguments explicitly (`imm`, `exec_fn`...).

coreblocks/fu/zbc.py
```diff
@@ -0,0 +1,259 @@
...
+                    m.d.comb += high_res.eq(0)
+                    m.d.comb += rev_res.eq(1)
+                    m.d.comb += value1.eq(i1[::-1])
+                    m.d.comb += value2.eq(i2[::-1])
+
+            params_fifo.write(m, rob_id=arg.rob_id, rp_dst=arg.rp_dst, high_res=high_res, rev_res=rev_res)
+
+            m.d.sync += clmul.i1.eq(value1)
+            m.d.sync += clmul.i2.eq(value2)
+            m.d.sync += clmul.reset.eq(1)
```
tilk, line 246: Can these be combinatorial? We would then be one cycle faster.

tilk:
Is there a reason for the lack of progress under this PR? Do we merge it as-is?

pa000:
I will finish it this weekend

coreblocks/fu/zbc.py
pa000: ok

coreblocks/fu/zbc.py
pa000: It was checked implicitly by creating the recursive submodules. Doing it explicitly is more readable though so I changed it.

coreblocks/fu/zbc.py
pa000: I tried doing that and I think it would require different handling in the ZbcUnit: the result would have to be read in the `issue` method. For now I would leave it like that.

coreblocks/fu/zbc.py
pa000: thanks

coreblocks/fu/zbc.py
pa000: they can, thanks

coreblocks/fu/zbc.py
pa000: I also didn't like how it looked but couldn't get it to work otherwise. Thanks a lot!

pa000:
It took way longer than needed but I believe it's ready now

tilk (approved):
Let's merge it finally.

PR #298, opened 2023-04-21
More readable exception on mismatched method return layout

Kristopher38:
Turns out I was wrong, you can deduce where the error originated (somewhat) from the stack trace, see first line:
```
  ...
  File "/home/kris/Dropbox/amaranth-tests/coreblocks/coreblocks/structs_common/csr.py", line 324, in elaborate
    def _():
  File "/home/kris/Dropbox/amaranth-tests/coreblocks/coreblocks/transactions/core.py", line 823, in decorator
    m.d.comb += assign(out, ret_out, fields=AssignType.ALL)
  File "/home/kris/.local/lib/python3.10/site-packages/amaranth/hdl/dsl.py", line 38, in __iadd__
    self._builder._add_statement(assigns, domain=self._domain, depth=self._depth)
  File "/home/kris/.local/lib/python3.10/site-packages/amaranth/hdl/dsl.py", line 484, in _add_statement
    for stmt in Statement.cast(assigns):
  File "/home/kris/.local/lib/python3.10/site-packages/amaranth/hdl/ast.py", line 1376, in cast
    return _StatementList(list(chain.from_iterable(map(Statement.cast, obj))))
  File "/home/kris/Dropbox/amaranth-tests/coreblocks/coreblocks/utils/utils.py", line 231, in assign
    raise KeyError("Field {} not present in rhs".format(name))
KeyError: 'Field old_pc not present in rhs'
```

But I made this PR anyway. New exception now looks like this (some stuff omitted for clarity):

```
    ...
    raise MismatchedLayoutError("Field {} not present in rhs".format(name), lhs_fields, rhs_fields)
coreblocks.utils.utils.MismatchedLayoutError: 'Field old_pc not present in rhs'

The above exception was the direct cause of the following exception:

  ...
  File "/home/kris/Dropbox/amaranth-tests/coreblocks/coreblocks/structs_common/csr.py", line 324, in elaborate
    def _():
  File "/home/kris/Dropbox/amaranth-tests/coreblocks/coreblocks/transactions/core.py", line 830, in decorator
    raise MismatchedLayoutError(
coreblocks.utils.utils.MismatchedLayoutError: "Method fetch_continue (defined in CSRUnit in module coreblocks.structs_common.csr) has output layout defined as {'old_pc'} but returned layout is {'next_pc'}"
```

Looking for feedback, in particular:
- is custom exception an overkill?
- what to do with other exception types raised in `assign`? Should we perhaps just catch them all instead and add method name and its location to the message (and get rid of the custom exception, but have a less nice message instead)?

Also this fixes a bug when no error was generated if method returned `None` but had a nonempty output layout.

There's also another assign in the transaction core for the input arguments which I'll tackle but I want some early feedback.


lekcyjna123:
For me custom exception is fine. The only issue which I see is in the name: `coreblocks.utils.utils.MismatchedLayoutError` We have two times `utils`. How about change it to `coreblocks.utils.exceptions.MismatchedLayoutError`?

tilk:
As you say, the way this solution works makes some of the possible problems detected in `assign` reported in a clearer way, and some are left as is. I would be in favor of a more clean solution. Maybe our chief Python master @Arusekk has some ideas?

coreblocks/utils/utils.py
```diff
@@ -115,6 +116,13 @@ class AssignType(Enum):
 AssignArg: TypeAlias = ValueLike | Mapping[str, "AssignArg"]
 
 
+class MismatchedLayoutError(KeyError):
```
Arusekk, line 119: I like the base class stays `KeyError`.

coreblocks/transactions/core.py
```diff
@@ -819,7 +819,18 @@ def decorator(func: Callable[..., Optional[RecordDict]]):
         with method.body(m, ready=ready, out=out) as arg:
             ret_out = method_def_helper(method, func, arg, **arg.fields)
 
-        if ret_out is not None:
+        if ret_out is None:
+            ret_out = Record([])
```
Arusekk, line 823:
```suggestion
            return
```
am I reading correctly?

coreblocks/transactions/core.py
```diff
@@ -819,7 +819,18 @@ def decorator(func: Callable[..., Optional[RecordDict]]):
...
+        try:
             m.d.comb += assign(out, ret_out, fields=AssignType.ALL)
+        except MismatchedLayoutError as e:
+            owner = type(method.owner)
+            lhs = e.lhs_fields if e.lhs_fields else "empty"
+            rhs = e.rhs_fields if e.rhs_fields else "empty"
+            msg = "Method {} (defined in {} in module {}) has output layout defined as {} but returned layout is {}"
+            raise MismatchedLayoutError(
+                msg.format(method.name, owner.__name__, owner.__module__, lhs, rhs), e.lhs_fields, e.rhs_fields
+            ) from e
```
Arusekk, line 834: I think just re-raising the exception (`raise e`, `raise` would not influence the traceback) optionally silencing the internals of layout assignment (`raise e from None`, but it could be kind-of useful to see the trace in some cases; or could it?) is enough here if the exception had a nice [`__str__`](https://docs.python.org/3/reference/datamodel.html#object.__str__).

Method name is very useful though.

[Docs about exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining) for those interested.

Arusekk (changes requested):

coreblocks/transactions/core.py
Kristopher38: Nope. If the method definition didn't return anything, we still want to run `assign(out, ret_out)` so that it can throw an exception if `out` isn't an empty layout (the method was supposed to return something but it didn't), but `ret_out` can't be `None` so I use type-correct equivalent - empty layout.

coreblocks/transactions/core.py
```diff
@@ -819,7 +819,18 @@ def decorator(func: Callable[..., Optional[RecordDict]]):
...
             ret_out = method_def_helper(method, func, arg, **arg.fields)
 
-        if ret_out is not None:
+        if ret_out is None:
+            ret_out = Record([])
+
+        try:
             m.d.comb += assign(out, ret_out, fields=AssignType.ALL)
+        except MismatchedLayoutError as e:
+            owner = type(method.owner)
```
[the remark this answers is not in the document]
Kristopher38: I very much don't care about memory usage. We're doing metaprogramming (not *real* programming) and here I'm gathering every bit of useful info I can get my hands on ;)

coreblocks/transactions/core.py
Kristopher38: Just reraising an exception would leave the user wondering why the mismatch occurs - with a message like `layout mismatch lhs: {...} rhs: {...}` I don't know what's lhs or rhs in the context of a method expected return layout and returning some record from it. I want to be helpful by explicitly saying which is which, so that I have to do less work - I immediately see *why* (and where) the mismatch occured.

Kristopher38:
I've tried something else, but I'm not sure how I feel about it (is it atrocious? or is it a viable compromise?). I've also looked at method tests and this doesn't cover all cases like I thought it would (assign can also throw `TypeError` when doing method call input argument assignment). I'm considering just adding a note to the original exception via `.add_note(str)` :unamused:

lekcyjna123:
I think that using `add_note` will be cleaner solution :)

Kristopher38:
I've settled for using `add_note` except it's a python3.11 feature... I've added a crude compatibility to not force us to use 3.11 yet.

coreblocks/transactions/_utils.py
```diff
@@ -150,3 +151,9 @@ def get_caller_class_name(default: Optional[str] = None) -> tuple[Optional[Elabo
         return None, default
     else:
         raise RuntimeError("Not called from a method")
+
+
+def assign_exception_helper(e, method):
+    new_e = AssignError(e.__str__())  # this can be removed in python 3.11
+    new_e.add_note("Caused by: {}".format(method))
+    return new_e
```
tilk, line 159: Is this the only place where `AssignError` is constructed? If so, why isn't this behavior in the constructor?

Kristopher38: >Maybe just monkey-patch the exception object with an add_note method, which extends the message, and be done with it?

I've looked into it and concluded that it would be even uglier. See https://stackoverflow.com/a/6062677
I can do it this way if you really want. Or as I said I could replace every exception in `assign` with `AssignError` and remove the `assign_exception_helper` weirdness.

tilk: Thanks for the link. I think what you're trying to make the code easy to transition to Python 3.11. If so, maybe just separate the magic (making the special exception) from the non-magic (using `add_note`). Or whatever, this discussion is basically bikeshedding.

Kristopher38: I feel like this whole PR is bikeshedding ;)

Kristopher38: There, should be good to go now.

lekcyjna123 (approved):
Looks good. Lets merge it finally :)

PR #299, opened 2023-04-22
Zba enable flag

wojpok:
This is a controversial change. Basically it is not as simple as it seemed.
There was no convenient way to remove ZBA extension from ALU decoder, so even if ALU was not using ZBA, ZBA was still present in the op_types. The workaround for this issue was to add a function that generates an Alu Decoder Manager with respect to ZBA enable flag. It works, but I would consider this an ugly solution


tilk:
Indeed this is controversial: for marginal gains (reducing the size of a single multiplexer before addition), ugliness is introduced.

I think that part of the problem is that `DecoderManager` is a completely static being: every method there is a `@classmethod`. This means that to parametrize a `DecoderManager`, you have to do this strange class-in-a-function contraption.

wojpok:
Agreed. I have an idea for yet another approach but I need some time to prepare proof of concept

coreblocks/fu/alu.py
```diff
@@ -9,108 +9,69 @@
...
+                with OneHotCase(self.enum["SRA"]):
+                    m.d.comb += self.out.eq(Cat(self.in1, Repl(self.in1[xlen - 1], xlen)) >> self.in2[0:xlen_log])
+            if hasattr(self.enum, "SLT"):
+                with OneHotCase(self.enum["SLT"]):
+                    m.d.comb += self.out.eq(self.in1.as_signed() < self.in2.as_signed())
+            if hasattr(self.enum, "SLTU"):
+                with OneHotCase(self.enum["SLTU"]):
+                    m.d.comb += self.out.eq(self.in1 < self.in2)
+            if hasattr(self.enum, "SH1ADD"):
+                with OneHotCase(self.enum["SH1ADD"]):
```
tilk: There is no real use case for disabling single opcodes in the ALU. There is no such thing as a RISC-V compliant processor without SLT, for example. And using this to make specialized mini-ALUs is overkill, as:

* The addition itself is the heavy part of the ALU. Separating addition-related opcodes (ADD, SUB, SLT, SLTU, SH*ADD) is pretty much pointless.
* Bitwise operations come at basically no cost, there is no reason not to use them as a package.

Basically, I believe that the change introduces complexity for no reason at all.

coreblocks/fu/alu_builder.py
```diff
@@ -0,0 +1,96 @@
+from enum import IntFlag, auto
+from typing import Sequence, Type
+from coreblocks.fu.fu_decoder import DecoderManager
+from coreblocks.fu.alu import Alu, ALUComponent
+from coreblocks.params import OpType, Funct3, Funct7
+
+
+class InstExt(IntFlag):
+    DEFI = auto()
```
Arusekk, line 9: defi sounds a lot like those crypto scams 😉

test/fu/test_alu.py
```diff
@@ -69,34 +72,34 @@ def process():
             sim.add_process(process)
 
     def test_add(self):
-        self.check_fn(AluFn.Fn.ADD, operator.add)
+        self.check_fn(self.enum["ADD"], operator.add)
```
Arusekk, line 75: This could be resolved with some typing I guess, so that we can still `self.enum.ADD` etc.

coreblocks/params/configurations.py
```diff
@@ -74,7 +74,15 @@ def replace(self, **kwargs):
 full_core_config = CoreConfiguration(
     isa_str="rv32imzicsr",
     func_units_config=(
-        RSBlockComponent([ALUComponent(), JumpComponent()], rs_entries=4),
+        RSBlockComponent(
+            [
+                # Separate ALU for shifting
+                AluBuilder(tag_set=InstTag.ANY ^ InstTag.SHIFT).build_component(),
```
Arusekk, line 80: idk if readable enough
```suggestion
                AluBuilder(tag_set=InstTag.ANY & ~InstTag.SHIFT).build_component(),
```

Arusekk (commented):
This whole PR sounds a bit complicated to me, but I guess it must be.

coreblocks/fu/alu.py
tilk: Yes, you can do it, but you can also do a lot of things you will absolutely never want to do. This PR, in its current shape, takes something that should be dead simple and makes a weird contraption out of it.

Issue #300, opened 2023-04-23
Add timeout for github verification


lekcyjna123:
We should add timeout in github verification to prevent of burning computation minutes in case of loops in simulations. As for now in case of errors jobs can run more than 5h which is way too long:
https://github.com/kuznia-rdzeni/coreblocks/actions/runs/4777452613

tilk:
This is probably the right way to to do it:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

PR #302, opened 2023-04-23
Added timeouts for workflows.

lekcyjna123:
This PR should be first fix for issue, that in case of problems with tests our workflows can stuck and run way too long (till github default timeout which is 6 hours).

Fix problem reported in #300

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


tilk:
Aren't these a little *too* restrictive? I would multiply them by 2, at least, to give some tolerance for variability and some space to grow.

lekcyjna123:
They are all based on current data. They are circa twice as long as needed, but I can make them longer.

tilk (approved):

piotro888 (approved):

PR #303, opened 2023-04-23
RS debug signals

tilk:
This allows to see RS and FU signals in GTKWave files, which should aid debugging. This PR is a temporary solution - #269, if done right, would subsume this.


lekcyjna123 (approved):

Kristopher38 (approved):
lgtm, also realized that i've been using `zip(range(len(iterable)), iterable)` instead of `enumerate(iterable)` my entire life

xThaid (approved):

PR #304, opened 2023-04-23
MemoryBank and Serializer

lekcyjna123:
Here is set of changes in common code which was a part of #293. Changes:
- name anonymous submodules to have better printing in gtkwave (in transactron and in `SimpleTestCircuit`)
- Add `MemoryBank`
- Add `Serializer`
- Extend interface of `SimpleTestCircuit`
- improve stubs
- changes in `BasicFifo` interface

ToDo:
- [x] Add doc strings

If needed I can split this PR into smaller once.


tilk:
Docstrings are needed, it's not clear on the first glance what the new modules do.

test/common.py
```diff
@@ -128,7 +129,17 @@ def elaborate(self, platform):
         for name, attr in [(name, getattr(self._dut, name)) for name in dir(self._dut)]:
             if isinstance(attr, Method):
                 self._io[name] = TestbenchIO(AdapterTrans(attr))
-                m.submodules += self._io[name]
+                m.submodules[name] = self._io[name]
+            if isinstance(attr, list):
+                for i, elem in enumerate(attr):
+                    if isinstance(elem, Method):
+                        self._io[name + str(i)] = TestbenchIO(AdapterTrans(elem))
+                        m.submodules[name + str(i)] = self._io[name + str(i)]
```
tilk, line 137: No, not these `list` shenanigans again. I'm starting to think that traversing some random lists is something possibly painful. More points against this:

* Possibly not generic enough. Someday, you might want lists of lists. Or dicts. Or something else.
* The string manipulation business for keys leads to ugly code and possibly errors.
* The change, together with the other one, makes the `SimpleTestCircuit` not that simple anymore.

test/common.py
```diff
@@ -109,9 +109,10 @@ def signed_to_int(x: int, xlen: int) -> int:
 
 
 class SimpleTestCircuit(Elaboratable, Generic[_T_HasElaborate]):
-    def __init__(self, dut: _T_HasElaborate):
+    def __init__(self, dut: _T_HasElaborate, *, external_submodules: Optional[list[_T_HasElaborate]] = None):
         self._dut = dut
         self._io = dict[str, TestbenchIO]()
+        self.external_submodules = external_submodules
```
tilk, line 115: I'm not certain I like this. This feature is used to pass `TestbenchIOs` for methods to be mocked. I believe there is a better abstraction to be found.

lekcyjna123:
I have added doc-strings so this pull request is ready for review in my opinion. I will be thinking how to provide better abstraction for operating on collections of methods and how to prepare TestbenchIO for method stubs.

lekcyjna123:
> No, not these list shenanigans again. I'm starting to think that traversing some random lists is something possibly painful. More points against this:
>
 >  - Possibly not generic enough. Someday, you might want lists of lists. Or dicts. Or something else.
 >  - The string manipulation business for keys leads to ugly code and possibly errors.

I agree with that points. But we have to find something out, because there will be a lot of such code in vector core and in future in superscalar parts. So to start the discussion here is my first alternative proposition:

### MethodContainer protocol

In this solution we can define our own protocol which will describe a class which can store collection of methods. In basic implementation it will be a wrapper on list, but with added some methods to abstract away some problems.

Functionalities which we need:
- get debug signals from all methods
- get all methods with names to put methods as a submodules
- add new elements to the container

tilk:
> MethodContainer protocol

This looks like a great idea. Probably a generic wrapper could be created, which would work for various data structures. Or alternatively, maybe, a combinator approach would also work? Regardless, I think that "adding" shouldn't be part of the interface, as this depends on the underlying data structure.

lekcyjna123:
Ok. So I will prepare some proof of concept.

coreblocks/transactions/core.py
```diff
@@ -122,7 +122,7 @@ def trivial_roundrobin_cc_scheduler(
         Linear ordering of transactions which is consistent with priority constraints.
     """
     sched = Scheduler(len(cc))
-    m.submodules += sched
+    add_next_submodule(m, sched, "_transactron_sheduler")
```
tilk, line 125:
```suggestion
    add_next_submodule(m, sched, "_transactron_scheduler")
```

coreblocks/transactions/lib.py
```diff
@@ -83,6 +87,93 @@ def _():
...
+                ("addr", self.addr_width),
+                ("data", self.data_layout),
+                ("mask", self.width // self.granularity),
+            ]
+
+        self.read_req = Method(i=self.read_req_layout)
+        self.read_resp = Method(o=self.data_layout)
+        self.write = Method(i=self.write_layout)
+
+        self.read_resp.schedule_before(self.read_req)
```
tilk, line 141: I know we don't have this standardized, but I think I prefer having this in `elaborate`, near the method definitions. Why? Because then it's close to the things whose correct behavior depends on the scheduling order.

coreblocks/transactions/lib.py
```diff
@@ -709,3 +810,79 @@ def elaborate(self, platform):
...
+
+            @def_method(m, self.serialize_in[i])
+            def _(arg):
+                pending_requests.write(m, {"id": i})
+                self.serialized_req_method(m, arg)
+
+            @def_method(m, self.serialize_out[i], ready=(pending_requests.head.id == i))
+            def _():
+                pending_requests.read(m)
+                return self.serialized_resp_method(m)
```
tilk, line 884: This has one problem: if one of the response received is stubbornly stopped for some reason, the others can't run, even though their responses are already in the queue. Why not multiple queues? This would still preserve relative order for each user., without introducing unnecessary blocking.

tilk:
By the way, this is one hell of a PR, mixing multiple unrelated changes.

coreblocks/utils/utils.py
```diff
@@ -316,3 +317,29 @@ def align_to_power_of_two(num: int, power: int) -> int:
...
+
+    """
+    i = 0
+    while True:
+        try:
+            getattr(m.submodules, name + str(i))
+            i += 1
+        except AttributeError:
+            break
+    m.submodules[name + str(i)] = sub
```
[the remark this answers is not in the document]
Arusekk: I think if we are doing this, then we should not be doing this.

Why not use a separate module, and then put everything like this underneath without any names?  Does this approach have any obvious benefit?

Arusekk (changes requested):
This is a massive PR meaning-wise.  Such changes should be IMO split into several functional parts.
There are 8 distinct parts to it:
* A new MemoryBank library module.
* A new Serializer library module.
* Changes in Forwarder library module.
* Stub updates.
* Changes from `submodules += mod` to `submodules.mod = mod`.
* The weird `"_name" + str(i) for i in range()` design pattern helping the above change maybe?  I left a comment for that specifically.
* A functional change for accepting some `external_submodules`.
* A nice debug message change (I mean `f'Trying to use method {} which is not defined yet'`).

I'm sure 2 distinct parts *can* be sometimes groupped together, if they are helpful or depend on each other, but maybe they should be split into 2 commits, lkml-style.
But 8... they should be just split into at least 2 PRs as far as I'm concerned.
Especially if they are not some huge changes, where this would be particularly cumbersome to do.

We demand easier reviews! 😄

coreblocks/transactions/lib.py
```diff
@@ -83,6 +87,93 @@ def _():
...
+        prev_read_addr = Signal(self.addr_width)
+        m.d.comb += read_port.addr.eq(prev_read_addr)
+
+        # read_resp has to be defined before read_req, to handle read_output_valid signal correctly
+        @def_method(m, self.read_resp, ready=read_output_valid)
+        def _():
+            m.d.sync += read_output_valid.eq(0)
+            return read_port.data
+
+        @def_method(m, self.read_req, ready=~read_output_valid | self.read_resp.run)
```
xThaid, line 159: `self.read_resp.run` creates a combinatorial path between the request and response methods. Not only it may extend the critical path, but also the user of this module may inadvertently create a loop. I think this path can be cut by using Forwarder.

lekcyjna123:
As requested changes from this review have been split into set of smaller reviews #314 #317 #319. I change status of this PR to draft till these 3 PRs will be merged.

coreblocks/transactions/lib.py
tilk: Actually, this can probably be handled by external queues.

coreblocks/transactions/lib.py
```diff
@@ -83,6 +87,93 @@ def _():
...
+        self.write = Method(i=self.write_layout)
+
+        self.read_resp.schedule_before(self.read_req)
+
+    def elaborate(self, platform) -> Module:
+        m = Module()
+
+        mem = Memory(width=self.width, depth=self.elem_count)
+        m.submodules.read_port = read_port = mem.read_port()
+        m.submodules.write_port = write_port = mem.write_port()
```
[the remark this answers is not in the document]
lekcyjna123: Yes.
https://github.com/amaranth-lang/amaranth/blob/main/examples/basic/mem.py

coreblocks/transactions/lib.py
lekcyjna123: Forwarder uses exactly the same trick, so it doesn't help for critical path. To don't create loop there is `shedule_before`.

coreblocks/transactions/lib.py
```diff
@@ -115,13 +206,19 @@ def __init__(self, layout: MethodLayout):
...
+        self.clear.add_conflict(self.read, Priority.LEFT)
+        self.clear.add_conflict(self.write, Priority.LEFT)
 
     def elaborate(self, platform):
         m = Module()
 
         reg = Record.like(self.read.data_out)
         reg_valid = Signal()
         read_value = Record.like(self.read.data_out)
+        m.d.comb += self.head.eq(read_value)
```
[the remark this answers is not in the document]
lekcyjna123: To have an interface which is in depended from inner representation.

test/common.py
lekcyjna123: Fixed with the use of ModuleConnector and recursion.

coreblocks/utils/utils.py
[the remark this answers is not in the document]
lekcyjna123: Removed

coreblocks/transactions/core.py
lekcyjna123: Changes done in different PR.

coreblocks/transactions/lib.py
lekcyjna123: Fixed

lekcyjna123:
Status of this PR is set back to ready for review, because all needed PRs have been already merged.

test/transactions/test_transaction_lib.py
```diff
@@ -466,3 +520,75 @@ def method_process():
...
+
+        random.seed(14)
+
+        serialized_data = deque()
+        port_data = [deque() for _ in range(port_count)]
+
+        got_request = False
+
+        def random_wait(rand: int):
+            yield from self.tick(random.randrange(rand) + 1)
```
Arusekk, line 556: inconsistent (one, 1/2)

Maybe it would be useful if it could be extracted to some mix-in class?

test/transactions/test_transaction_lib.py
```diff
@@ -108,6 +110,58 @@ def process():
...
+        data_width = 6
+        m = SimpleTestCircuit(MemoryBank(data_layout=[("data", data_width)], elem_count=max_addr))
+
+        data_dict: dict[int, int] = dict((i, 0) for i in range(max_addr))
+        read_req_queue = deque()
+
+        random.seed(seed)
+
+        def random_wait(rand: int):
+            yield from self.tick(random.randrange(1, rand + 1))
```
Arusekk, line 127: inconsistent (one, 2/2)

test/transactions/test_transaction_lib.py
```diff
@@ -466,3 +520,75 @@ def method_process():
...
+            def f():
+                for _ in range(test_count):
+                    d = random.randrange(2**data_width)
+                    yield from m.test_circuit.serialize_in[i].call(field=d)
+                    port_data[i].append(d)
+                    yield from random_wait(requestor_rand)
+
+            return f
+
+        def responser(i: int):
```
Arusekk, line 580: maybe responder?

test/common.py
```diff
@@ -111,11 +112,35 @@ def signed_to_int(x: int, xlen: int) -> int:
...
+        self._io = dict[str, TestbenchIO | ModuleConnector]()
 
-    def __getattr__(self, name: str):
+    def __getattr__(self, name: str) -> Any:
         return self._io[name]
 
+    class NotMethodException(TypeError):
+        pass
+
+    @staticmethod
```
Arusekk, line 123: effectively a classmethod (actual cls instead of SimpleTestCircuit in the body maybe)

tilk: Or a private function in `elaborate`, no need to pollute the attribute namespace. (I know Pythonists don't care, buuut...)

lekcyjna123: > Also, isn't this doing the dreaded recursion over structures in modules? How do we know it doesn't run into the same issues auto_debug_signals had?

No. Here we have recursion over `list[Method] | dict[str, Method] | Method`, on the other hand in `auto_debug_signals` we do recursion over fields in classes.

test/transactions/test_transaction_lib.py
lekcyjna123: Inconsistent fixed.

> Maybe it would be useful if it could be extracted to some mix-in class?

This is good idea, but to implement under different PR, because impact of such changes will be big.

test/common.py
lekcyjna123: > effectively a classmethod (actual cls instead of SimpleTestCircuit in the body maybe)

Moved as a private function of `elaboratable`.

test/common.py
```diff
@@ -111,12 +112,29 @@ def signed_to_int(x: int, xlen: int) -> int:
...
-    def __getattr__(self, name: str):
+    def __getattr__(self, name: str) -> Any:
         return self._io[name]
 
+    class NotMethodException(TypeError):
+        pass
+
     def elaborate(self, platform):
+        def transform_methods_to_testbenchios(
+            container: _T_nested_collection[Method],
```
tilk, line 125: This function isn't actually called with this type; later in the body, you call it on something of type `Method | list | dict`. And you use it, too: if `container` was really a `_T_nested_collection[Method]`, the `else` would be unreachable, but it isn't and performs a vital function.

lekcyjna123: Yes. It is because Python don't support deep runtime type checking out of the box. It has to by implemented manually and I have done it in this function. But maybe it is wrong idea, and I should create first TypeGuard to do deep check in runtime and first after that call `transform_methods_to_testbenchios`. Such solution will be cleaner.

tilk: Or just write the actual type the function works with in the first place and write one or two lines of comments.

tilk: This is now cleaner. The `nested_collection` stuff looks potentially useful, maybe it should be moved? Also, the name `_T_nested_collection` should probably be changed, the `_T` prefix should probably be left to TypeVars.

coreblocks/utils/utils.py
```diff
@@ -338,6 +338,17 @@ def __init__(self, *args: HasElaborate, **kwargs: HasElaborate):
         self.args = args
         self.kwargs = kwargs
 
+    def __getattr__(self, name: str) -> Any:
```
tilk, line 341: Why `Any`?

test/common.py
lekcyjna123: I thought about that, but as for now we don't have any direct usage for that in other places, so I don't know if this should be moved to coreblock/utils or if this should be a part of utils in tests. So I decided to not move at all and delay this decision till first example of usage in other place will be available, so that no guessing will be needed.

coreblocks/utils/utils.py
lekcyjna123: Because `HasElaborate` is too weak. It hasn't any information about methods available in elaboratables, so pyright reported errors, even for correct usage. So because of lack of any other ideas how to type that I decided to disable checking of returned types.

coreblocks/transactions/lib.py
tilk: Actually, I don't understand what is happening here, as `read_resp` doesn't set anything combinationally. Where's the dependency?

lekcyjna123: Dependency is in `self.read_resp.run` in `ready` signal. It is used to as combinational signal to reduce pipeline latency. If there is no valid request currently processed, there is no dependency. Only `read_req` is ready. But when we currently process a request, then we can receive new request only if someone will read old results, so in `read_req` we need to already know if `read_resp` was called or not.

coreblocks/utils/utils.py
tilk: Couldn't you just retain the original, well-typed, references to the modules? Can you point me to the usage site?

If this is not practical, and you add homogenously-typed modules, using generics is one possible solution.

lekcyjna123: One example is in `TestSerializer`. We create `ModuleConnector`, which has `Serializer` instance and two instances of `TestbenchIO`, next we use instance of this module connector to get access to methods which are part of `Serializer.

```py
self.req_method = TestbenchIO(Adapter(i=layout))
self.resp_method = TestbenchIO(Adapter(o=layout))

test_circuit = SimpleTestCircuit(
    Serializer(
        port_count=port_count,
        serialized_req_method=self.req_method.adapter.iface,
        serialized_resp_method=self.resp_method.adapter.iface,
    )
)
m = ModuleConnector(test_circuit=test_circuit, req_method=self.req_method, resp_method=self.resp_method)
```

```py
def requestor(i: int):
    def f():
        for _ in range(test_count):
            d = random.randrange(2**data_width)
            yield from m.test_circuit.serialize_in[i].call(field=d)
            port_data[i].append(d)
            yield from random_wait(requestor_rand)
```

Without `Any`, there is error from pyright that `m.test_circuit` don't have `serialize_in` field.

tilk: Why won't you just use `test_circuit` instead of `m.test_circuit` there?

tilk: Another take: the `ModuleConnector` was supposed to be a module *combinator*, something that packs a number of modules into one module, to be added into `submodules`. It wasn't supposed to be a module *container*. Note that `submodules` don't work well with this use case, too - it's used to put modules there and that's all, it forgets the types of the modules inside, and this is fine.

If, for some reason, a module *container* will be needed, I'd look into dataclass mixins for that, as this would allow to keep track of types *and* keep the nice constructor syntax of `ModuleConnector`. For now, I recommend removing these two methods.

tilk (approved):

tilk:
This PR desperately needs a review!

coreblocks/transactions/lib.py
xThaid: I wasn't clear enough. I still believe that it is possible. The instruction cache works that way -- fully pipelined, 1 cycle latency between request and response and no combinatorial path between the request and response methods.

So basically it works by employing one FIFO of depth 2 and one forwarder. After the request method is called, we put the request to the FIFO and at the same time we query the memory. In the next cycle, we read the response from the memory and put it in the forwarder. In the response method we read both Forwarder and FIFO. The FIFO basically works as a rate-limiter (because we don't care about it contents, but only about the number of elements), so a simplified version can be used (something like a semahore)

test/transactions/test_transaction_lib.py
```diff
@@ -109,6 +111,58 @@ def process():
...
+        def random_wait(rand: int):
+            yield from self.tick(random.randrange(rand) + 1)
+
+        def writer():
+            for i in range(test_count):
+                d = random.randrange(2**data_width)
+                a = random.randrange(max_addr)
+                yield from m.write.call(data=d, addr=a)
+                yield Settle()
+                # print("w", a, d)
```
xThaid, line 136: The comments here and below can be probably removed.

test/transactions/test_transaction_lib.py
```diff
@@ -109,6 +111,58 @@ def process():
             sim.add_sync_process(process)
 
 
+class TestMemoryBank(TestCaseWithSimulator):
```
xThaid, line 114: The test doesn't test if the memory bank is pipelined.

coreblocks/transactions/lib.py
tilk: This worries me a little, as (if I understand you correctly) your current implementation reorders reads with respect to writes. This can possibly reduce the usefulness of this module. Did you really intend this to happen?

tilk: I've read your code, and I'm now certain that your code doesn't reorder reads and writes - *your test does*. Don't mess with the internals (you really like that, don't you?) and fix the test.

tilk (changes requested):
The test for `MemoryBank` improperly models reads and writes, which somehow lead you to hack up the test to make it pass. Fix the test, please.

coreblocks/transactions/lib.py
```diff
@@ -811,6 +723,151 @@ def elaborate(self, platform):
...
+        prev_read_addr = Signal(self.addr_width)
+        m.d.comb += read_port.addr.eq(prev_read_addr)
+
+        self._internal_read_resp = Method(o=self.data_layout)
+        # internal_read_resp has to be defined before read_req, to handle read_output_valid signal correctly
+        @def_method(m, self._internal_read_resp, ready = read_output_valid)
+        def _():
+            m.d.sync += read_output_valid.eq(0)
+            return read_port.data
+
```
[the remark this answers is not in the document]
tilk: Oh, but this code *doesn't* run always! This runs only when the forwarder is ready, which can make you miss the cycle when the results are available.

test/common.py
```diff
@@ -105,18 +106,52 @@ def signed_to_int(x: int, xlen: int) -> int:
     return x | -(x & (2 ** (xlen - 1)))
 
 
+def guard_nested_collection(cont: Any, t: Type[T]) -> TypeGuard[_T_nested_collection[T]]:
+    if isinstance(cont, list):
+        return all([guard_nested_collection(elem, t) for elem in cont])
+    elif isinstance(cont, dict):
+        return all([guard_nested_collection(elem, t) for elem in cont.values()])
```
Arusekk, line 113: Not sure if we want Yet Another Dict Tree Traversal... I kind of think this can also be done in a tiny bit more generic way, like:
```suggestion
    if isinstance(cont, (list, dict)):
        if isinstance(cont, dict):
            cont = cont.values()
        return all([guard_nested_collection(elem, t) for elem in cont])
```

we should just do one generic version of it, and pass a callback/aggregator function along.

test/common.py
```diff
@@ -105,18 +106,52 @@ def signed_to_int(x: int, xlen: int) -> int:
...
         self._dut = dut
-        self._io = dict[str, TestbenchIO]()
+        self._io: dict[str, _T_nested_collection[TestbenchIO]] = {}
 
-    def __getattr__(self, name: str):
+    def __getattr__(self, name: str) -> Any:
         return self._io[name]
 
     def elaborate(self, platform):
+        def transform_methods_to_testbenchios(
```
Arusekk, line 132: Yet Another Dict Tree Traversal.  Do we really need it?

coreblocks/transactions/lib.py
```diff
@@ -709,3 +723,224 @@ def elaborate(self, platform):
...
+        self.addr_width = bits_for(self.elem_count - 1)
+
+        self.read_req_layout = [("addr", self.addr_width)]
+        if self.granularity is None:
+            self.write_layout = [("addr", self.addr_width), ("data", self.data_layout)]
+        else:
+            self.write_layout = [
+                ("addr", self.addr_width),
+                ("data", self.data_layout),
+                ("mask", self.width // self.granularity),
```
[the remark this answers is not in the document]
tilk: I think I also prefer the `append` approach; it makes it *more* explicit that the rest of the fields are the same in both cases.

coreblocks/transactions/lib.py
xThaid: Instead of decreasing the size of the semaphore, we could increase the size of the forwarder (I know that our forwarder right now has a constant size, but right now I can't find a reason why Forwarder wouldn't have arbitrary size).

Another solution (the one I used in the cache module) is that we can just buffer (let's call it **B**) the latest request and if the method `request` is called, we read the memory using the argument of the method, if it is not, we use **B** to read the memory. That way, if we have two consecutive reads, the first read's response will be buffered in the forwarder and the second read address will be sitting in the buffer.

test/common.py
lekcyjna123: > we should just do one generic version of it, and pass a callback/aggregator function along.

As for now we don't have dict/list tree traversal in any other place, so i decided to don't make it separate function as for now. In future if this will be needed, it can be done separate.

test/common.py
lekcyjna123: If you have an idea for to realise this transformation in other way I am happy to hear you.

coreblocks/transactions/lib.py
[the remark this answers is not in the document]
lekcyjna123: As explained on meeting this will work correctly.

tilk:
Did you stop using a LLM for grammar fixes? Would approve, but the docstrings are atrocious.

coreblocks/transactions/lib.py
```diff
@@ -709,3 +723,280 @@ def elaborate(self, platform):
...
+
+        @def_method(m, self.read_req, ~write_pending)
+        def _(addr):
+            m.d.sync += read_output_valid.eq(1)
+            m.d.comb += read_port.addr.eq(addr)
+            m.d.sync += prev_read_addr.eq(addr)
+            zipper.write_args(m, valid=1)
+
+        if self.safe_writes:
+            self.read_req.schedule_before(self.write)
```
tilk, line 912: Is this needed?

tilk: Or let's state it another way. There is some combinational dependence between method bodies. But does it influence scheduling?

Arusekk (approved):

coreblocks/transactions/lib.py
lekcyjna123: It turns out that, this dependence is not needed. It have been removed.

PR #305, opened 2023-04-23
Fix Verilator on cocotb

tilk:
So, there is some silliness in `cocotb_bus` which prevented Verilator from working right. (https://github.com/cocotb/cocotb/issues/3259)

Also updated `cocotb` to the development version to benefit from updates to the Verilator support.

### DIFF
--- a/.github/workflows/cocotb.yml
+++ b/.github/workflows/cocotb.yml
@@ -26,7 +26,7 @@ jobs:
         run: |
           python3 -m pip install --upgrade pip
           pip3 install -r requirements-dev.txt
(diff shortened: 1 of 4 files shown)


piotro888:
Can you add `TRACE ?= 0` variable that adds `--trace-fst --trace-structs` arguments, please?

piotro888 (approved):

tilk:
> Can you add `TRACE ?= 0` variable that adds `--trace-fst --trace-structs` arguments, please?

Don't have to, `VERILATOR_TRACE=1` does that out of the box.

xThaid (approved):

Issue #289, opened 2023-04-15
Generate compiler options from `isa_str`


piotro888:
We use some of the extensions (like Zmmul) that for some reason are not supported by compliers. We should generate superset of our internal isa_str that is accepted by gcc or binutills as `isa` valid paramter. This is needed in core tests generated from assembly.

Example `rv32i_zmmul` -> would be converted to `rv32im` for gcc and binutils (possibly with extra parameter `-no-div` for gcc).
Disscussion for Zmmul in gcc: https://groups.google.com/a/groups.riscv.org/g/isa-dev/c/M-37xaIllsw, change was not accepted in 2021 because this extension was not yet frozen, but now it is ratified since 2022 (https://github.com/riscv/riscv-isa-manual/blob/master/src/m.tex, maybe we should file an issue to gcc?). Anyway, problem would probably exist for some other extensions too.

piotro888:
For Zmmul, it turns out that support for it was commited 7/8 months ago to gcc and binutils; and it should be available in the latest release of binutils 2.40, but not yet in gcc (we have to wait for next release).

tilk:
I've backported binutils 2.40 for Debian stable, and I can confirm that it understands Zmmul.

PR #306, opened 2023-04-23
Generate isa_str from func units config

### LINKED ISSUE #295: Add ISA string generator
Actually we have data duplication in core configuration. We have isa string and separately we provide structure with functional units which define supported extension. Instead of having ISA string generate it from used functional units.

piotro888:
Easier said than done; it was is tricky to implement with extension implications. I needed to modify ISA and optype decoding to make it easier to implement. I also removed Privileged spec instructions from I extension, they don't belong to any extension.

Computes *partialy supported extensions* isa string and *fully suported extensions* isa string. We need to use *partial* isa string for decoder and compilers (if we want to temporarily support some OpTypes from extensions that are not yet completed (and maybe in some experimental configs)). Note that we still don't support `I` extension, `FENCE` is missing. 
Full ISA string would be used as extra information to user (printed later in gen verilog script; also with prvileged information, as it is not included in isa string) and with special flag in configuration, error could be raised if configuration where  `partial != full` would be used, as it may cause compiler to generate unsported instructions (and I think it should be enabled by default in export configurations in future).

TODO: Tests, add to GP, some extra conversions, compiler isa string (temporary for zmmul #289). 
Fixes #295

### DIFF
--- a/coreblocks/params/fu_params.py
+++ b/coreblocks/params/fu_params.py
@@ -2,13 +2,15 @@ from abc import abstractmethod, ABC
 from typing import Iterable
 
 from coreblocks.utils.protocols import FuncBlock, FuncUnit
+from coreblocks.params.isa import Extension, extension_implications
+from coreblocks.params.optypes import optypes_required_by_extensions
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Collection
+
 
 if TYPE_CHECKING:
     from coreblocks.params.genparams import GenParams
     from coreblocks.params.optypes import OpType
-
 
 __all__ = [
     "BlockComponentParams",
@@ -39,3 +41,50 @@ class FunctionalComponentParams(ABC):
 
 def optypes_supported(block_components: Iterable[BlockComponentParams]) -> set["OpType"]:
     return {optype for block in block_components for optype in block.get_optypes()}
+
+
+def _remove_implications(extensions: set[Extension]):
+    implied_extensions = set()
+    for ext in extensions:
+        if ext in extension_implications:
+            implied_extensions |= extension_implications[ext]
+    return extensions - implied_extensions
+
+
+## Move that to genparams?
+def extensions_supported(fu_config: Collection[BlockComponentParams]) -> tuple[set[Extension], set[Extension]]:
+    optypes = optypes_supported(fu_config)
+
+    # Fully and partially supported extensions
+    extensions_parital: set[Extension] = set()
+    # Fully supported extensions
+    extensions_full: set[Extension] = set()
+
+    # OK: Add global switch if we want to use partial extensions with warning of unsupported ops (and default for now). If not selected error if partial != full
+
+    for ext in Extension:
+        ext_added_optypes = optypes_required_by_extensions({ext}, resolve_implications=False, ignore_unsupported=True)
+        if ext_added_optypes & optypes:
+            extensions_parital.add(ext)
+
+        ext_all_optypes = optypes_required_by_extensions({ext}, resolve_implications=True, ignore_unsupported=True)
+        if optypes and ext_all_optypes and optypes.issuperset(ext_all_optypes):
+            extensions_full.add(ext)
+
+    # needed for extensions that just imply others without adding new optypes (like B).
+    # add them to partial extensions if they are fully supported for implied removal
+    extensions_parital |= extensions_full
+
+    # remove implied extensions
+    extensions_parital = _remove_implications(extensions_parital)
+    extensions_full = _remove_implications(extensions_full)
+
+    # special modifications
+
+    # TODO: convert to G; E->I (if gp); ISA STR ORDER MATTERS
+
+    #: generate isa strings in another function (isa_str, compiler_isa_str (bypass for zmull), extra info (with partial/full info and machine/sup mode)
+
+    print(extensions_parital, extensions_full)
+
+    return (extensions_parital, extensions_full)
--- a/coreblocks/params/optypes.py
+++ b/coreblocks/params/optypes.py
@@ -52,6 +52,8 @@ optypes_by_extensions = {
         OpType.LOAD,
         OpType.STORE,
         OpType.FENCE,
+    ],
+    Extension.MACHINE_MODE: [
         OpType.ECALL,
         OpType.EBREAK,
         OpType.MRET,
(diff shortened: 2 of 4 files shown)


tilk:
Maybe the decoder shouldn't depend on the extensions, but on optypes? It uses the extensions just to iterate over optypes anyway.

coreblocks/params/optypes.py
```diff
@@ -80,11 +82,22 @@ class OpType(IntEnum):
...
-def optypes_required_by_extensions(extensions: Iterable[Extension]) -> set[OpType]:
+def optypes_required_by_extensions(
+    extensions: set[Extension], resolve_implications=True, ignore_unsupported=False
+) -> set[OpType]:
     optypes = set()
+
+    if resolve_implications:
+        implied_extensions = set()
+        for ext in extensions:
+            if ext in extension_implications and ext in optypes_by_extensions:
```
Arusekk, line 93: Pretty sure this would work even with `|` instead of sets.

coreblocks/params/optypes.py
```diff
@@ -80,11 +82,22 @@ class OpType(IntEnum):
...
+                implied_extensions |= extension_implications[ext]
+        extensions |= implied_extensions
+
     for ext in extensions:
         if ext in optypes_by_extensions:
             optypes = optypes.union(optypes_by_extensions[ext])
-        else:
-            raise Exception(f"Core do not support {ext} extension")
+        elif not ignore_unsupported:
+            raise Exception(f"Core do not support {ext!r} extension")
```
Arusekk, line 101: does?

coreblocks/stages/rs_func_block.py
```diff
@@ -1,6 +1,7 @@
 from collections.abc import Collection
 from amaranth import *
 from dataclasses import dataclass
+from typing import Iterable
```
Arusekk, line 4: is it needed?

coreblocks/params/fu_params.py
```diff
@@ -39,3 +41,50 @@ def get_optypes(self) -> set["OpType"]:
...
 def optypes_supported(block_components: Iterable[BlockComponentParams]) -> set["OpType"]:
     return {optype for block in block_components for optype in block.get_optypes()}
+
+
+def _remove_implications(extensions: set[Extension]):
+    implied_extensions = set()
+    for ext in extensions:
+        if ext in extension_implications:
+            implied_extensions |= extension_implications[ext]
+    return extensions - implied_extensions
```
Arusekk, line 51: this would just work with flags instead of sets (xor would be needed maybe instead of minus)

coreblocks/params/fu_params.py
```diff
@@ -2,14 +2,16 @@
...
+
+from typing import TYPE_CHECKING, Collection
 
-from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from coreblocks.params.genparams import GenParams
     from coreblocks.params.optypes import OpType
 
-
```
Arusekk, line 12: Linter angry.  Linter no like.
(90% sure it requires double newlines on toplevel declarations)

Arusekk (commented):
Nice!

coreblocks/stages/rs_func_block.py
tilk: If needed, import from `collections.abc`.

coreblocks/params/fu_params.py
piotro888: changed to flags everywhere

coreblocks/stages/rs_func_block.py
piotro888: it is needed, it was accidentally imported with some other module before

coreblocks/params/genparams.py
```diff
@@ -38,9 +39,18 @@ class GenParams(DependentCache):
     def __init__(self, cfg: CoreConfiguration):
         super().__init__()
 
-        self.isa = ISA(cfg.isa_str)
         self.func_units_config = cfg.func_units_config
 
+        ext_parital, ext_full = extensions_supported(self.func_units_config, cfg.embedded, cfg.compressed)
+        extensions = ext_parital if cfg.allow_partial_extensions else ext_full
+        if not cfg.allow_partial_extensions and ext_parital != ext_full:
+            raise RuntimeError(f"Extenstions {ext_parital&~ext_full!r} are only partialy supported")
```
tilk, line 47:
```suggestion
        ext_partial, ext_full = extensions_supported(self.func_units_config, cfg.embedded, cfg.compressed)
        extensions = ext_partial if cfg.allow_partial_extensions else ext_full
        if not cfg.allow_partial_extensions and ext_partial != ext_full:
            raise RuntimeError(f"Extensions {ext_partial&~ext_full!r} are only partially supported")
```

piotro888:
> Maybe the decoder shouldn't depend on the extensions, but on optypes? It uses the extensions just to iterate over optypes anyway.

Maybe that's a good idea, but decoding  would still need to be dynamically disabled by whole extensions with `misa` CSR writes, so I'm not sure if that would be very helpful. Anyway I think that's for another PR.

coreblocks/params/genparams.py
piotro888: thanks, fixed

piotro888: the same typo is in other places, I will fix it later

lekcyjna123 (approved):

tilk:
My main issue with this change is that `E` and `C` extensions are silently accepted, while they in reality are not supported. Otherwise, looks great.

piotro888:
I think that E extension is supported, because is only reduces number of logical registers and it is already handled in https://github.com/kuznia-rdzeni/coreblocks/blob/18adb721b2e3652e1b341b79605add203077879e/coreblocks/params/isa.py#L301, so it is done from RISCV point of view (we don't support exceptions yet)

For C extension I can add temporary error message. It can be removed from decoder test, because it doesn't contain any compressed instructions for now. If needed in future in tests without full support, it can be bypassed via _extension_implications (and thats the intended use case for tests).

tilk:
> so it is done from RISCV point of view (we don't support exceptions yet)

The decoder has an illegal instruction signal, which is currently unused, but is there nonetheless. Is this signal correctly indicating illegal instructions for registers which don't exist in RV32E?

tilk (approved):

PR #307, opened 2023-04-25
Add instruction cache bypass

xThaid:
This will allow to generate core without instruction cache.

### DIFF
--- a/coreblocks/frontend/icache.py
+++ b/coreblocks/frontend/icache.py
@@ -1,6 +1,6 @@
 from functools import reduce
 import operator
-from dataclasses import dataclass
+from typing import Protocol
 
 from amaranth import *
 from amaranth.utils import log2_int
@@ -9,6 +9,7 @@ from coreblocks.transactions.core import def_method, Priority
 from coreblocks.transactions import Method, Transaction
 from coreblocks.params import ICacheLayouts, ICacheParameters
 from coreblocks.utils import assign, OneHotSwitchDynamic
+from coreblocks.utils._typing import HasElaborate
 from coreblocks.transactions.lib import *
 from coreblocks.peripherals.wishbone import WishboneMaster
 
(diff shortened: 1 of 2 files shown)


coreblocks/frontend/icache.py
```diff
@@ -12,10 +13,99 @@
...
+        m.d.comb += instr_out.eq(word)
+    else:
+        with m.If(addr[2] == 0):
+            m.d.comb += instr_out.eq(word[:32])  # Take lower 4 bytes
+        with m.Else():
+            m.d.comb += instr_out.eq(word[32:])  # Take upper 4 bytes
+    return instr_out
+
+
+@dataclass
```
tilk, line 31: This class is used as a base class. Dataclass is probably not the thing we want, as dataclasses are data carriers; they automatically generate `__init__`, `__repr__`, possibly comparison operators, `__hash__`, etc. This should probably be a protocol (see https://peps.python.org/pep-0544/). We already use protocols for this use case (see https://github.com/kuznia-rdzeni/coreblocks/blob/master/coreblocks/utils/protocols.py). Protocols behave as ABCs at runtime (https://peps.python.org/pep-0544/#non-goals), subclassing them as you do here is valid.

coreblocks/frontend/icache.py
```diff
@@ -8,14 +9,102 @@
...
-__all__ = ["ICache", "SimpleWBCacheRefiller"]
+__all__ = ["ICache", "ICacheBypass", "ICacheInterface", "SimpleWBCacheRefiller"]
 
 
-class ICache(Elaboratable):
+def extract_instr_from_word(m: Module, params: ICacheParameters, word: Signal, addr: Value):
+    instr_out = Signal(params.instr_width)
+    if len(word) == 32:
+        m.d.comb += instr_out.eq(word)
+    else:
```
lekcyjna123, line 24: I think, that we can add here a check and raise an error when `len(word)` is different than 32 or 64. This will be good, for future possible extensions to 128 bits.

lekcyjna123 (approved):

tilk (approved):

coreblocks/frontend/icache.py
tilk: This is not critical, but I think that this is a good change, as it makes the assumption explicit.

PR #308, opened 2023-04-26
Transaction core refactor

tilk:
This PR refactors the transaction core in preparation for future changes. The refactor should also increase the readability of the code by factoring out common functionality. Summary of the changes:

* `MethodMap` class is introduced, which creates and keeps the transaction->method and method->transaction maps.
* The `_call_graph` function is removed. Maps are created now in `MethodMap`, and the method uses dict with the relation list are created using `MethodMap`.
* Dynamically created fields in the `TransactionManager` are removed. Instead, data dependencies are now explicit.
* Types such as `Mapping` are used to protect from accidental mutation.
* `TransactionOrMethod` alias is introduced.
* Some imports from `typing` are now imported from `collections.abc` instead, as imports from `typing` are deprecated since Python 3.9 (see e.g. https://docs.python.org/3/library/typing.html#typing.Mapping).

### DIFF
--- a/coreblocks/transactions/core.py
+++ b/coreblocks/transactions/core.py
@@ -1,14 +1,14 @@
 from collections import defaultdict
+from collections.abc import Iterable, Sequence, Callable, Mapping, Iterator
 from contextlib import contextmanager
 from enum import Enum, auto
-from typing import Callable, ClassVar, Mapping, TypeAlias, TypedDict, Union, Optional, Tuple, Iterator
-from types import MethodType
+from typing import ClassVar, TypeAlias, TypedDict, Union, Optional, Tuple
 from graphlib import TopologicalSorter
 from typing_extensions import Self
 from amaranth import *
 from amaranth import tracer
 from amaranth.hdl.ast import Statement
-from itertools import count
+from itertools import count, chain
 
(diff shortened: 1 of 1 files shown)


coreblocks/transactions/core.py
```diff
@@ -48,17 +49,55 @@ class Priority(Enum):
...
+                if not method.defined:
+                    raise RuntimeError("Trying to use method which is not defined yet")
+                if method in self.methods_by_transaction[transaction]:
+                    raise RuntimeError("Method can't be called twice from the same transaction")
+                self.methods_by_transaction[transaction].append(method)
+                self.transactions_by_method[method].append(transaction)
+                rec(transaction, method)
+
+        for transaction in transactions:
+            self.methods_by_transaction[transaction] = []
```
lekcyjna123, line 77: this is probably not needed due to `defaultdict`

lekcyjna123 (approved):

coreblocks/transactions/core.py
tilk: Not true! Without this line, there are no entries for transactions which don't call methods (a rarity, but permitted), which leads to problems later.

xThaid (approved):
Looks better now.

PR #309, opened 2023-04-26
[RFC] Add submodules to any Elaboratable

lekcyjna123:
In #304 there was proposition to prepare abstract way to add submodules to Elaboratable in tests. Here is first proposition, where I add submodules using function. I don't like this implementation, because of checking if we got Module, TransactionModule or something else is tricky and not future-proof. So I plan to prepare now second solution using class and next we can compere them.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -107,10 +107,9 @@ def signed_to_int(x: int, xlen: int) -> int:
 
 
 class SimpleTestCircuit(Elaboratable, Generic[_T_HasElaborate]):
-    def __init__(self, dut: _T_HasElaborate, extra: Optional[list[_T_HasElaborate]] = None):
+    def __init__(self, dut: _T_HasElaborate):
         self._dut = dut
         self._io = dict[str, TestbenchIO]()
-        self.extra = extra
 
     def __getattr__(self, name: str):
         return self._io[name]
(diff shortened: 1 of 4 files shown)


test/common.py
```diff
@@ -105,13 +106,11 @@ def signed_to_int(x: int, xlen: int) -> int:
     return x | -(x & (2 ** (xlen - 1)))
 
 
-_T_HasElaborate = TypeVar("_T_HasElaborate", bound=HasElaborate)
-
-
 class SimpleTestCircuit(Elaboratable, Generic[_T_HasElaborate]):
-    def __init__(self, dut: _T_HasElaborate):
+    def __init__(self, dut: _T_HasElaborate, extra: Optional[list[_T_HasElaborate]] = None):
```
lekcyjna123, line 110: Changes in SimpleTestCircuit should be removed. They were only for test purposes.

lekcyjna123:
I have uploaded second proposition to add submodules. This time using Elaboratable wrapper. Sadly, this solution is also ugly, because we have to first add submodules containing method mocks and first after that we can add TransactionModule, which impose order of adding submodules in AddSubmodulesWrapper.

Because both my solutions are ugly I would like to ask if someone have suggestions how to improve that?

lekcyjna123:
I am closing this pull request as changes proposed here are implemented in:
#317 #318 #319 #321

PR #310, opened 2023-04-26
Method container

lekcyjna123:
As stated in #304 I have prepared some proof of concept for method containers. This will solve following problems:
- adding generic container of methods as a submodule
- generating debug signals from methods

Problems which are open as for now:
- support for other data types than Methods (see #303)

### DIFF
--- a/coreblocks/utils/method_collections.py
+++ b/coreblocks/utils/method_collections.py
@@ -1,0 +1,61 @@
+from amaranth import *
+
+from coreblocks.transactions import Method
+from typing import Optional, TypeVar, Callable, Union, Generic
+from ._typing import HasElaborate, SignalBundle
+from .protocols import MethodCollection, MethodCollectionElement, MethodCollectionOutput
+from abc import abstractmethod, ABC
+
+
+__all__ = ["ListMethodCollection", "DictMethodCollection"]
+
+T = TypeVar('T')
+_T_nested_dict = dict[str, Union[T, '_T_nested_dict']]
+
+class MethodCollectionBase(ABC, MethodCollection):
+    @abstractmethod
+    def get_all_methods(self) -> MethodCollectionOutput:
+        raise NotImplementedError()
+
+    def transform_methods(self, f : Callable[[Method], T]) -> _T_nested_dict[T]:
+        output = {}
+
+        methods = self.get_all_methods()
+        for k, elem in methods.items():
+            if isinstance(elem, Method):
+                output[k] = f(elem)
+            else:
+                output[k] = self.transform_methods(f)
+        return output
+
+    def debug_signals(self) -> SignalBundle:
+        return self.transform_methods(lambda m: m.debug_signals())
+
+class ListMethodCollection(MethodCollectionBase):
+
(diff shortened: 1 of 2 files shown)


coreblocks/utils/method_collections.py
```diff
@@ -0,0 +1,61 @@
...
+    @abstractmethod
+    def get_all_methods(self) -> MethodCollectionOutput:
+        raise NotImplementedError()
+
+    def transform_methods(self, f : Callable[[Method], T]) -> _T_nested_dict[T]:
+        output = {}
+
+        methods = self.get_all_methods()
+        for k, elem in methods.items():
+            if isinstance(elem, Method):
```
tilk, line 25: I would do it the other way around: test if it's a `MethodCollectionBase` first. This would ease generalization.

coreblocks/utils/method_collections.py
```diff
@@ -0,0 +1,61 @@
...
+from .protocols import MethodCollection, MethodCollectionElement, MethodCollectionOutput
+from abc import abstractmethod, ABC
+
+
+__all__ = ["ListMethodCollection", "DictMethodCollection"]
+
+T = TypeVar('T')
+_T_nested_dict = dict[str, Union[T, '_T_nested_dict']]
+
+class MethodCollectionBase(ABC, MethodCollection):
```
tilk, line 15: Maybe generalization is in order, with type of the internal structure as the type argument?

coreblocks/utils/method_collections.py
```diff
@@ -0,0 +1,61 @@
...
+            if isinstance(elem, Method):
+                output[k] = f(elem)
+            else:
+                output[k] = self.transform_methods(f)
+        return output
+
+    def debug_signals(self) -> SignalBundle:
+        return self.transform_methods(lambda m: m.debug_signals())
+
+class ListMethodCollection(MethodCollectionBase):
```
tilk, line 34: Maybe generalization is in order, with type of the stored values as the type argument?

lekcyjna123:
Yes, these amorphous is by design. I wanted to have flexibility to pass any MethodCollection as an element to other MethodCollection, so that this can be easy nested and to have abstraction in parent container.

I agree that current structure is an sandwich and it is not prepared to be edited. Main concept was, that first the whole data structure is created and first after that it is wrapped by MethodCollection. After that point we can access each element and mutate it, but we can not add any new element.

tilk:
My question is, do you actually need it that way, or is it for some future, unspecified and unknown, use cases? (If so, I still think it would be best if this was optional, and not mandatory.)

lekcyjna123:
No, I don't need such flexibility. As for now even simple list is enough, but I try to be future-proof as you have suggested:
> No, not these list shenanigans again. I'm starting to think that traversing some random lists is something possibly painful. More points against this:
>
> -  Possibly not generic enough. Someday, you might want lists of lists. Or dicts. Or something else.

So I already think about lists of lists or lists of dicts.

tilk:
Lists of lists and lists of dicts are perfectly normal, and easy to express with generics. Lists of maybe a list, maybe a dict, and maybe a Method seem less useful.

PR #314, opened 2023-04-28
Remove anonymous submodules

lekcyjna123:
- Refactor of Schedulers to remove Module from theirs arguments. 
- Changed anonymous submodules to named

Depends on #317

### DIFF
--- a/coreblocks/transactions/core.py
+++ b/coreblocks/transactions/core.py
@@ -10,7 +10,7 @@ from amaranth import tracer
 from amaranth.hdl.ast import Statement
 from itertools import count, chain
 
-from coreblocks.utils import AssignType, assign
+from coreblocks.utils import AssignType, assign, ModuleConnector
 from ._utils import *
 from ..utils._typing import StatementLike, ValueLike, SignalBundle
 from .graph import Owned, OwnershipGraph, Direction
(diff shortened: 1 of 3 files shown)


coreblocks/transactions/core.py
```diff
@@ -292,7 +294,8 @@ def elaborate(self, platform):
         m = Module()
 
         for cc in _graph_ccs(rgr):
-            self.cc_scheduler(method_map, m, cgr, cc, porder)
+            mm = self.cc_scheduler(method_map, cgr, cc, porder)
+            m.submodules += mm
```
tilk, line 298: Why the temporary variable?

lekcyjna123: removed

coreblocks/transactions/core.py
```diff
@@ -292,7 +294,7 @@ def elaborate(self, platform):
         m = Module()
 
         for cc in _graph_ccs(rgr):
-            self.cc_scheduler(method_map, m, cgr, cc, porder)
+            m.submodules += self.cc_scheduler(method_map, cgr, cc, porder)
```
xThaid, line 297: nit: maybe we can name these submodules like `scheduler0` etc.

xThaid (approved):

coreblocks/transactions/core.py
tilk: Maybe let's do like @Arusekk proposed during the meeting: put them all inside a single Module, and add that module with a name.

lekcyjna123: Fixed

tilk (approved):

PR #316, opened 2023-04-29
Added new test to workflow to check if traces works.

lekcyjna123:
Based on our last experience it looks like auto_debug_signals is very fragile. So here I added an additional test step to workflow, to check if traces works on small subset of tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,6 +34,9 @@ jobs:
       - name: Run tests
         run: ./scripts/run_tests.py --verbose
 
+      - name: Check traces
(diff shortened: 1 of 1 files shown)


tilk:
Well, this doesn't check a lot. Doing this on some simple full-core test would probably catch more problems.

lekcyjna123:
Changed

Kristopher38 (changes requested):
As suggested, maybe add a `-c|--count` argument to `run_tests` to run only a specific amount matching the regex (not sure where else this would be useful but this is as good reason as any to add this)

tilk (approved):

PR #317, opened 2023-04-29
Module connector

lekcyjna123:
Module which can add other elements as submodules.

### DIFF
--- a/coreblocks/utils/utils.py
+++ b/coreblocks/utils/utils.py
@@ -317,16 +317,18 @@ def bits_from_int(num: int, lower: int, length: int):
     """Returns [`lower`:`lower`+`length`) bits from integer `num`."""
     return (num >> lower) & (1 << (length) - 1)
 
+
 class ModuleConnector(Elaboratable):
     """
     An Elaboratable to create a new module, which will have all arguments
     added as its submodules.
     """
+
     def __init__(self, *args, **kwargs):
         """
         Parameters
         ----------
-        *args 
+        *args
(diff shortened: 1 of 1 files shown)


coreblocks/utils/utils.py
```diff
@@ -316,3 +316,33 @@ def align_to_power_of_two(num: int, power: int) -> int:
...
+    An Elaboratable to create a new module, which will have all arguments
+    added as its submodules.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
+        Parameters
+        ----------
+        *args
+            Modules which should be named as anonymous submodules.
```
tilk, line 332: ?
```suggestion
            Modules which should be added as anonymous submodules.
```

coreblocks/utils/utils.py
```diff
@@ -316,3 +316,33 @@ def align_to_power_of_two(num: int, power: int) -> int:
...
     return (num >> lower) & (1 << (length) - 1)
+
+
+class ModuleConnector(Elaboratable):
+    """
+    An Elaboratable to create a new module, which will have all arguments
+    added as its submodules.
+    """
+
+    def __init__(self, *args, **kwargs):
```
tilk, line 327: Types missing. There was some `HasElaborate` thing or something like this which would fit perfectly.

lekcyjna123: Fixed

tilk (approved):

xThaid (approved):

PR #318, opened 2023-04-29
Refactor of ROB test

lekcyjna123:
Here is small refactor of ROB test to use SimpleTestCircuit.


tilk (approved):
Why not.

xThaid (approved):
`+15 −40` I like it

PR #319, opened 2023-04-29
Refactor tests with TransactionModule

lekcyjna123:
As agreed I have prepared refactor of our unit tests and moved TransactionModule initialisation to TestCaseWithSimulator. 

Additionally I have added a `silence_must_use_warnings` which I need to silent warnings from tests which throw exceptions (TestDefMethod.invalid....). Before my changes modules were created only in one files so there was suppressed using amaranth tricks. Now I use there modules created in `test/common.py` so I needed to suppress more selectively. Of course live can not be easy and UnusedMustUse warning are raised in `__del__` function, which is invoked by garbage collector. So I have to invoke garbage collection by hand to be sure, that warning will be handled using proper context.

Depends on #317

### DIFF
--- a/coreblocks/utils/_typing.py
+++ b/coreblocks/utils/_typing.py
@@ -1,4 +1,4 @@
-from typing import Protocol, Sequence, Tuple, Type, TypeAlias, Iterable, Mapping, runtime_checkable
+from typing import Protocol, Sequence, Tuple, Type, TypeAlias, Iterable, Mapping, runtime_checkable, Union
 from enum import Enum
 from amaranth import *
 from amaranth.hdl.ast import ShapeCastable, Statement, ValueCastable
@@ -17,7 +17,7 @@ LayoutList = list[Tuple[str, ShapeLike | "LayoutList"]]
 
 
 class HasElaborate(Protocol):
-    def elaborate(self, platform) -> "HasElaborate":
+    def elaborate(self, platform) -> Union["HasElaborate", Fragment]:
         ...
 
 
--- a/coreblocks/utils/_typing.py
+++ b/coreblocks/utils/_typing.py
@@ -17,7 +17,7 @@ LayoutList = list[Tuple[str, ShapeLike | "LayoutList"]]
 
 
 class HasElaborate(Protocol):
-    def elaborate(self, platform) -> Union["HasElaborate", Fragment]:
+    def elaborate(self, platform) -> "HasElaborate":
         ...
 
 
--- a/coreblocks/utils/_typing.py
+++ b/coreblocks/utils/_typing.py
@@ -1,4 +1,4 @@
-from typing import Protocol, Sequence, Tuple, Type, TypeAlias, Iterable, Mapping, runtime_checkable, Union
+from typing import Protocol, Sequence, Tuple, Type, TypeAlias, Iterable, Mapping, runtime_checkable
 from enum import Enum
 from amaranth import *
 from amaranth.hdl.ast import ShapeCastable, Statement, ValueCastable
--- a/coreblocks/utils/utils.py
+++ b/coreblocks/utils/utils.py
@@ -16,6 +16,7 @@ __all__ = [
     "align_to_power_of_two",
     "bits_from_int",
     "ModuleConnector",
+    "silence_mustuse",
 ]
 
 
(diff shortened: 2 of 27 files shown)


coreblocks/utils/utils.py
```diff
@@ -345,4 +350,6 @@ def elaborate(self, platform):
         for name, elem in self.kwargs.items():
             m.submodules[name] = elem
 
+        if self.wrap_with_tm:
+            return coreblocks.transactions.core.TransactionModule(m)
```
tilk, line 354: No, absolutely not. That's not the job of this module.

test/common.py
```diff
@@ -382,3 +388,13 @@ def mock(func_self=None, /) -> TestGen[None]:
...
     return decorator
+
+
+def silence_must_use_warnings(f: Callable[[], T]) -> T:
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", UnusedMustUse)
+        ret = f()
+        # Unused warnings are raised in __del__ methods, so to be sure, that they will
+        # be reported, before we go out from catch_warnings contex we have to call garbage collection manualy
+        gc.collect()
```
tilk, line 399: GC hackery stinks. Isn't there a better way? The question to answer is - WHY is the unused warning happening? It looks like you added quite a lot of this silencing all over the place, and I don't understand why, and don't like it.

lekcyjna123: The unused warnings happen because there are tests which raise exceptions is test bodies ([tets/transactions/test_methods.py](https://github.com/kuznia-rdzeni/coreblocks/blob/98fd23ad5ba70013180be1228a3d61848fcafd19/test/transactions/test_methods.py#L86)) up to now these warnings was suppressed in such way:
https://github.com/kuznia-rdzeni/coreblocks/blob/98fd23ad5ba70013180be1228a3d61848fcafd19/test/transactions/test_methods.py#L1
But now I use modules created in different files so it is not enough.

tilk: Is this because you create the `TransactionManager` in a different place? You do it in the wrong one, anyway. If you construct `TransactionManager` in the test code, you could still use the same suppressing method and not introduce changes all over the code.

tilk: I think that the definition of Transaction module does not help. If it was changed to take a HasElaboratable and create a new module instead of hacking an existing one, it would be easier to use it differently.

lekcyjna123: > Is this because you create the TransactionManager in a different place? You do it in the wrong one, anyway. If you construct TransactionManager in the test code, you could still use the same suppressing method and not introduce changes all over the code.

It is because I instantiate `Module()` in TestCaseWithGenerator.

> I think that the definition of Transaction module does not help. If it was changed to take a HasElaboratable and create a new module instead of hacking an existing one, it would be easier to use it differently.

I have tried yesterday to change this interface and putting to it an HasElaborate type. Next I used this instance of TransactionModule in simulation, but this didn't work, because of `auto_debug_signals`. In this branch this function has still its old form, which go into loop on Modules. I will try to rebase this commit onto #269 and use TransactionModule one more time.

lekcyjna123: I have rewritten `TransactionModule`. I think that now it is cleaner and thanks for that, there is no much difference in comparison for our other `Elaboratables`. There is also no problems with `auto_debug_signals`.

coreblocks/utils/_typing.py
```diff
@@ -17,7 +17,7 @@
 
 
 class HasElaborate(Protocol):
-    def elaborate(self, platform) -> "HasElaborate":
+    def elaborate(self, platform) -> Union["HasElaborate", Fragment]:
```
tilk, line 20: Why? `Fragment` has Elaborate.

lekcyjna123: Removed

tilk (approved):

scripts/core_graph.py
```diff
@@ -17,10 +18,13 @@
...
 from coreblocks.params.configurations import basic_core_config  # noqa: E402
+from coreblocks.transactions.core import TransactionModule  # noqa: E402
 
 gp = GenParams(basic_core_config)
-elaboratable = TestElaboratable(gp)
-fragment = TracingFragment.get(elaboratable, platform=None).prepare()
+m = Module()
+tm = TransactionModule(m)
+m.submodules.elaboratable = TestElaboratable(gp)
+fragment = TracingFragment.get(tm, platform=None).prepare()
```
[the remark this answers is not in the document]
tilk: But actually, I don't understand why `TestElaboratable` from `test.test_core` is used, and the decision is not documented here. @Arusekk?

xThaid (approved):

PR #320, opened 2023-04-29
DummyLSU clear

lekcyjna123:
@Kristopher38  here is a clear for DummyLSU about which you have requested. I understand know your problems with testing of interrupts. There is a lot of corner cases and I suppose that my test can still have bugs and report errors even when LSU works correctly.

Big problem for my test case code (not for LSU) were sequences of requests where there are the same addresses and masks e.g.:
- (1, 0x123, 0xF), (2, 0x123, 0xF) - (clear on 1)
- (1, 0x123, 0xF), (2, 0x456, 0x1), (3, 0x123, 0xF) - (clear on 1 and 2)
Where we have: (request id, address, mask)

In such cases it was hard to distinguish in mock of wishbone slave if given request was issued before clear 9so we have to check for its correctness) or if there was no request send for given request id to wishbone slave mock, because clear was issued early. Sadly, such distinguish is needed to know next, what data are to be expected in consumer. So I have introduced the check to not issue clear if we have the same accesses nearby, but this is some kind of workaround and in future it would be good to find something better.


coreblocks/lsu/dummyLsu.py
```diff
@@ -264,6 +295,7 @@ def _(tag: Value, value: Value):
                 m.d.sync += current_instr.s2_val.eq(value)
                 m.d.sync += current_instr.rp_s2.eq(0)
 
+        # TODO fix, nie powinno się uruchamiać
```
Kristopher38, line 298: ???

Kristopher38 (commented):
Overall seems good, I couldn't spot any immediate bugs. I'll wait before reviewing the testing code until you implement timestamp tracking we discussed.
About the state machine: I'm starting to dislike complex boolean expressions inside the states themselves (e.g. `self.get_result_ack | (self.clear & ~op_initiated` and such). This is not necessarily an issue with the PR itself, but rather a systemic one - you try to perform transaction as soon as possible on some trigger condition as soon as it happens, and when that fails you do it unconditionally in the next state (and thus you need to keep track of `op_initiated`). This is a common pattern and I wonder if there's a better way to handle it and still have the code be efficient.

coreblocks/lsu/dummyLsu.py
lekcyjna123: I forgot to remove that.

lekcyjna123:
> Also: please change the base branch (what you're merging into) to feature/interrupts.

My changes don't cause any backward in-compatibility, so I am plan to merge it to master, to have that code in common branch.

Kristopher38:
The point of having a separate branch for interrupts was not only to have a place to introduce half-working changes but also so that we won't have to worry about synchronizing with master until we want to merge with it. In any case, it's not that important ;)

lekcyjna123:
This PR is probably no longer valid after last changes in handling stores. Merge conflicts should be resolved and clear should be tested one more time.

I change status of this PR to draft, but I don't plan to fix it in near future.

Kristopher38 (commented):
Turns out I had some review comments just sitting there that I forgot to submit... I couldn't figure out how the tests work - but this doesn't matter much since we'll be throwing these away anyway

tilk:
LSU is no longer in this form, and soon it will be very different. We also probably won't need to deal with clear methods everywhere.

PR #322, opened 2023-04-30
MustUse cleanups for TransactionModule refactor

tilk:
This PR addresses the reasons for the MustUse warnings, instead of painting over them.

### DIFF
--- a/test/transactions/test_methods.py
+++ b/test/transactions/test_methods.py
@@ -1,8 +1,7 @@
-# amaranth: UnusedElaboratable=no
 from amaranth import *
 from amaranth.sim import *
 
-from ..common import TestCaseWithSimulator, TestbenchIO, data_layout, silence_must_use_warnings
+from ..common import TestCaseWithSimulator, TestbenchIO, data_layout
 
 from coreblocks.transactions import *
 from coreblocks.transactions.lib import *
@@ -24,6 +23,7 @@ class TestDefMethod(TestCaseWithSimulator):
 
         def elaborate(self, platform):
             m = Module()
+            m._MustUse__silence = True  # type: ignore
 
             def_method(m, self.method)(self.method_definition)
 
@@ -35,13 +35,6 @@ class TestDefMethod(TestCaseWithSimulator):
     def do_test_definition(self, definer):
         with self.run_simulation(TestDefMethod.TestModule(definer)):
             pass
-
-    def assert_with_silence(self, definer, error_type):
-        def f():
-            with self.assertRaises(error_type):
-                self.do_test_definition(definer)
-
-        silence_must_use_warnings(f)
 
     def test_fields_valid1(self):
         def definition(arg):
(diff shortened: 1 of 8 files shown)


test/transactions/test_methods.py
```diff
@@ -153,6 +151,7 @@ def __init__(self):
 
             def elaborate(self, platform):
                 m = Module()
+                m._MustUse__silence = True  # type: ignore
```
lekcyjna123, line 154: You edit here class attribute. Where you restore default value after the test? Without setting this back to False, you silent warning for other test cases too. Or Unittest library restore default class attributes between test runs automatically?

lekcyjna123: It looks like I don't know python. This will work, because we don't modify here class variable. Instead of that we create an instance variable with the same name, which hide class variable.

tilk: Indeed, that's what happens.

lekcyjna123:
This don't work. You have disabled too much warnings. Here is counterexample: 

```git
diff --git a/test/utils/test_fifo.py b/test/utils/test_fifo.py
index 064d257..56aa8ec 100644
--- a/test/utils/test_fifo.py
+++ b/test/utils/test_fifo.py
@@ -2,6 +2,7 @@ from amaranth import *
 
 from coreblocks.utils.fifo import BasicFifo
 from coreblocks.transactions.lib import AdapterTrans
+from coreblocks.transactions.core import TransactionManager
 
 from test.common import TestCaseWithSimulator, TestbenchIO, data_layout
 from collections import deque
@@ -22,6 +23,8 @@ class BasicFifoTestCircuit(Elaboratable):
         m.submodules.fifo_write = self.fifo_write = TestbenchIO(AdapterTrans(self.fifo.write))
         m.submodules.fifo_clear = self.fifo_clear = TestbenchIO(AdapterTrans(self.fifo.clear))
 
+        tm = TransactionManager()
+
         return m
 ```
 

In this patch I edited a test in such way, that `TransactionManager` is not used. There is no expectations for exceptions so a warning should be issued. On master indeed there is warning, but in your patch set not. This is because you enable unused warning in `elaborate`, but when the `elaborate` have been called then you already know, that this Module is used...

lekcyjna123:
We would like something like:
"If an Elaboratable is being deleted, because of stack unwinding by exception, then don't raise warning."

As for now I see two ways to implement this requirement:
- wrap Elaboratable in our own class which will call original `elaborate` inside a `try ... except`. On exception use `inspect` module and iterate over local variables. If there is a Module set `_MustUse__silence = True`
- extend Elaboratable with setter to `_MustUse__silence` is this setter is called, then recursivly it should call setters for all its childrens. Code which handle exception will be responsible for setting `_MustUse__silence = True` on parent module

Sadly, both of these ideas are dirty and probably there are some problematic corner cases.

tilk:
You're right, that still has holes. I think we shouldn't spend too much time on this, as TransactionManager is never instantiated manually in practice. Still, if you have some idea how to do it better without reducing readability and invoking magic, go ahead.

lekcyjna123 (approved):
It looks good now. I am not sure if such solution will be scalable, but as for now it will be enough.

tilk:
@lekcyjna123: could you merge it into your PR, as this one retracts some of the changes you did there?

Arusekk:
I don't like this, but I guess I am late to the party.

tilk:
> I don't like this, but I guess I am late to the party.

If you have some better idea, go ahead.

PR #324, opened 2023-05-03
Fix `schedule_before`

tilk:
Bug discovered by @xThaid.

### DIFF
--- a/test/transactions/test_transactions.py
+++ b/test/transactions/test_transactions.py
@@ -331,3 +331,42 @@ class TestTransactionPriorities(TestCaseWithSimulator):
         with cm:
             with self.run_simulation(m):
                 pass
+
+
+class ScheduleBeforeTestCircuit(Elaboratable):
+    def elaborate(self, platform):
+        m = Module()
+        tm = TransactionModule(m)
+
+        method = Method()
+
+        @def_method(m, method)
+        def _():
+            pass
+
+        with tm.transaction_context():
+            with (t1 := Transaction()).body(m):
+                method(m)
+
+            with (t2 := Transaction()).body(m, request=t1.grant):
+                method(m)
+
+            t1.schedule_before(t2)
+
+        # so that Amaranth allows us to use add_clock
+        dummy = Signal()
+        m.d.sync += dummy.eq(1)
+
+        return tm
+
+
+class TestScheduleBefore(TestCaseWithSimulator):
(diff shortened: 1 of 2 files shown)


lekcyjna123:
Why this don't break `Forwarder` tests? Or why previously `Forwarder` tests were passing? As I understand reversing priority, should influence all current usages of `schedule_before`, so if somewhere priority was good, now should be wrong.

test/transactions/test_transactions.py
```diff
@@ -331,3 +331,59 @@ def test_unsatisfiable(self, priority: Priority):
...
+
+        def process():
+            to_do = 5 * [(0, 1), (1, 0), (1, 1)]
+            random.shuffle(to_do)
+            for r1, r2 in to_do:
+                yield m.r1.eq(r1)
+                yield m.r2.eq(r2)
+                yield
+                self.assertEqual((yield m.t1), r1)
+                self.assertFalse((yield m.t2))
```
lekcyjna123, line 386: Why `assertFalse` shoudn't be equal to `r1 &r2`?

tilk:
> Why this don't break `Forwarder` tests? Or why previously `Forwarder` tests were passing? As I understand reversing priority, should influence all current usages of `schedule_before`, so if somewhere priority was good, now should be wrong.

The dragon eggs start to hatch.

I think I know why: the scheduling order is taken into account only if there is a conflict between the ordered transactions. If there is no conflict, combinatorial dependencies are enough. (The bad thing is, when there is a conflict, this is basically an error. It looks like that handling situations like in `Forwarder` requires rethinking.)

By the way, this doesn't mean that `schedule_before` in `Forwarder` is completely useless, as it allows to detect combinational loops early.

test/transactions/test_transactions.py
tilk: I don't understand your comment.

tilk: In the test module, both transactions call the same method. Also, the second transaction can only run when the first transaction runs. But they can't run both, because then they would call the same method twice. So, only the first transaction can run, and the second never runs.

xThaid (approved):

lekcyjna123 (approved):

test/transactions/test_transactions.py
lekcyjna123: I understand now. Thanks!

PR #325, opened 2023-05-03
[RFC] Extended simulator

lekcyjna123:
We have nice framework to write normal code, but we still use low level tools (eg. Settle) in our tests. They have a lot of corner cases and it is hard to write tests which looks good and are easy to read. On the other hand our production code is going to be more complicated, so we need a better way to write tests.

I thought about writing unit tests in cocotb, but solution has some drawbacks:
- no direct support from amaranth
- need to go through verilog
- if  we use verilog then we loose our method abstraction
- non interface signals in verilog have strange names
- one unit tests would compose from two files (makefile and python code)

So I have decided to make exploration by coding and check how hard it will be to add more abstract commands to amaranth simulator. It looks like it is doable, here I present changes which add:
- SimulatorWrapper - in our simulation each cycle have 3 phases: Normal (standard simulation as known from amaranth), ReadOnly (only reading is allowed), NextCycle (all process ended this cycle and we are going to the next)
- ReadOnly - works simmilar to Settle() but wait till all process want to go to the next cycle, or want to be in ReadOnly phase
- WaitFor / Emit - conditional variables to synchronise different test process

Current status:
- Extended simulator is backward compatible with all synchronous tests
- First tests with ReadOnly works
- No tests to check if WaitFor/Emit works
- naming of classes and variables have to be improved

I would be happy to hear what you think about such changes.
- Do you think, that there will be useful?
- Have you some ideas for improvements?

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -3,7 +3,6 @@ import os
 import functools
 from contextlib import contextmanager, nullcontext
 from typing import Callable, Generic, Mapping, Union, Generator, TypeVar, Optional, Any, cast
-from enum import Enum, auto
 
 from amaranth import *
 from amaranth.hdl.ast import Statement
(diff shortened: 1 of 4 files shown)


tilk:
That's quite a lot of code! I sure hope this complexity is justified. For example, I would very like to see if this can be used to simplify tests we already have.

PR #331, opened 2023-05-07
Message test framework

lekcyjna123:
Hi,

I started to prepare message test framework and here is a draft version. As for now the main functionality of Messageframework is implemented and is already more or less typed (thanks @forell !). I started also to rewrite LSU test (function `test_body`), but due to time which was consumed by typing it is not ready yet.

ToDo:
- [ ] add support for Arusekk syntax in functions (probably as separate PR)
- [x] add function to combine messages in MessageQueueCombiner
- [ ] shorter names
- [x] add random wait
- [x] add build-in empty generator to start test flow
- [x] rewrite LSU test to new framework
- [ ] reduce delay (instead of `yield` use `Settle()` as condition var)
- [ ] debug problems with TypeVar appears only one (probably pyright bug simmilar to https://github.com/google/pytype/issues/379
- [ ] is it possible for type checker to treat `add_data_flow` as a relation which is transitive in terms of types?


lekcyjna123:
I have uploaded new version of message framework. Changes:
- Internal structure of MessageFrameworkProcess is now much simpler (but this make data forwarding between processing steps harder)
- MessageFrameworkProcess is now implemented as base class (this caused that test is longer)
- calls are now done using standard python syntax (but there is some magic under the hood)
- tests for clearing works correctly (more than 50 000 test iterations without any errors)

ToDo/ToDiscussion:
- sort MessageFrameworkProcess in topology order to insert correct amount of settles on the beginning
- add out-of-the box support for multi input process
- improve out-of-the box support for mocks

PR #333, opened 2023-05-10
ZBA enable flag

wojpok:
Yet another ZBA enable concept

What changed:

- All Decoder Managers are now instances not static classes
- ALU op codes are defined statically because there are no collisions and this approach is much simpler
- ALU components now accepts optional flag in constructors that disables/enables ZBA extension
  By default its enabled in ALU component and disabled in other classes

### DIFF
--- a/coreblocks/fu/alu.py
+++ b/coreblocks/fu/alu.py
@@ -1,5 +1,7 @@
 from typing import Sequence
+from typing_extensions import override
 from amaranth import *
+from amaranth import Value
 
 from coreblocks.transactions import *
 from coreblocks.transactions.core import def_method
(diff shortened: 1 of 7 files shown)


coreblocks/fu/alu.py
```diff
@@ -16,47 +16,72 @@
...
+            (Fn.SUB, OpType.ARITHMETIC, Funct3.ADD, Funct7.SUB),
+            (Fn.SLT, OpType.COMPARE, Funct3.SLT),
+            (Fn.SLTU, OpType.COMPARE, Funct3.SLTU),
+            (Fn.XOR, OpType.LOGIC, Funct3.XOR),
+            (Fn.OR, OpType.LOGIC, Funct3.OR),
+            (Fn.AND, OpType.LOGIC, Funct3.AND),
+            (Fn.SLL, OpType.SHIFT, Funct3.SLL),
+            (Fn.SRL, OpType.SHIFT, Funct3.SR, Funct7.SL),
+            (Fn.SRA, OpType.SHIFT, Funct3.SR, Funct7.SA),
+            (Fn.SH1ADD, OpType.ADDRESS_GENERATION, Funct3.SH1ADD, Funct7.SH1ADD),
```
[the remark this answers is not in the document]
tilk: These even should be added conditionally, as otherwise the decoder will think that the `SH*ADD` instructions are supported, even if they are not.

coreblocks/fu/alu.py
```diff
@@ -104,10 +131,11 @@ def elaborate(self, platform):
 
 
 class AluFuncUnit(Elaboratable):
-    optypes = AluFn.get_op_types()
+    optypes = AluFn().get_op_types()
```
tilk, line 134: See #334.

coreblocks/fu/alu.py
```diff
@@ -117,9 +145,9 @@ def __init__(self, gen: GenParams):
     def elaborate(self, platform):
         m = Module()
 
-        m.submodules.alu = alu = Alu(self.gen)
+        m.submodules.alu = alu = Alu(self.gen, zba_enable=self.zba_enable)
         m.submodules.fifo = fifo = FIFO(self.gen.get(FuncUnitLayouts).accept, 2)
-        m.submodules.decoder = decoder = AluFn.get_decoder(self.gen)
+        m.submodules.decoder = decoder = AluFn(zba_enable=self.zba_enable).get_decoder(self.gen)
```
tilk, line 150: And now you create the `AluFn` multiple times (another places are `Alu` and `ALUComponent`). As it's not static anymore, maybe you should construct it only once, especially since it's parametrized.

coreblocks/fu/jumpbranch.py
```diff
@@ -111,7 +109,7 @@ def elaborate(self, platform):
 
 
 class JumpBranchFuncUnit(Elaboratable):
-    optypes = JumpBranchFn.get_op_types()
+    optypes = JumpBranchFn().get_op_types()
```
tilk, line 112: See #334.

test/fu/test_fu_decoder.py
```diff
@@ -40,9 +40,9 @@ def handle_signals(self, decoder: Decoder, exec_fn: dict[str, int]) -> Generator
         return (yield decoder.decode_fn)
 
     def run_test_case(self, decoder_manager: Type[DecoderManager], test_inputs: Sequence[tuple]) -> None:
-        instructions = decoder_manager.get_instructions()
-        decoder = decoder_manager.get_decoder(self.gen_params)
-        op_type_dependent = len(decoder_manager.get_op_types()) != 1
+        instructions = decoder_manager().get_instructions()
+        decoder = decoder_manager().get_decoder(self.gen_params)
+        op_type_dependent = len(decoder_manager().get_op_types()) != 1
```
tilk, line 45: And here you also create the `DecoderManager` multiple times. Instead, you should change the type of `decoder_manager` from `Type[DecoderManager]` to `DecoderManager`. As the `DecoderManager` now can be parametrized, passing just the type is against some use cases.

coreblocks/fu/alu.py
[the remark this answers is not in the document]
wojpok: My bad, I refactored some code before pushing it and I have accidentally deleted conditional generation.

Ill fix it later

coreblocks/fu/alu.py
```diff
@@ -28,35 +33,47 @@ class Fn(IntFlag):
...
+                (self.Fn.SH1ADD, OpType.ADDRESS_GENERATION, Funct3.SH1ADD, Funct7.SH1ADD),
+                (self.Fn.SH2ADD, OpType.ADDRESS_GENERATION, Funct3.SH2ADD, Funct7.SH2ADD),
+                (self.Fn.SH3ADD, OpType.ADDRESS_GENERATION, Funct3.SH3ADD, Funct7.SH3ADD),
+            ]
+            if self.zba_enable
+            else []
+        )
+
+    @override
+    def get_function(self) -> Value:
```
xThaid, line 66: It won't work if somebody adds a next function after `SH3ADD`.  That function will get (from `auto()`) a value that cannot be represented using `len(self.get_instructions())` bits.

wojpok: Yes, but its not an issue since there won't be any other instructions:
ZBS and ZBC are in separate module and ZBB will be in separate module with shift-instructions

tilk: We don't know if some instruction outside of the B extension, or maybe in a new, future extension, will be useful to be implemented in the ALU. Also, we still haven't decided how to implement the RV64I *W instructions. Your assumption is unfounded.

coreblocks/fu/alu.py
```diff
@@ -28,35 +31,38 @@ class Fn(IntFlag):
...
+            (self.Fn.SH3ADD, OpType.ADDRESS_GENERATION, Funct3.SH3ADD, Funct7.SH3ADD),
+        ] * self.zba_enable
 
 
 class Alu(Elaboratable):
-    def __init__(self, gen: GenParams):
+    def __init__(self, gen: GenParams, alu_fn=AluFn()):
         self.gen = gen
+        self.zba_enable = alu_fn.zba_enable
+        self.alu_fn = alu_fn
```
tilk, line 63: Is this used anywhere?

wojpok: No, It served no purpose; I've just deleted it

tilk (approved):

lekcyjna123 (approved):

PR #335, opened 2023-05-10
Refactor cocotb tests

xThaid:
I did it for a couple of reasons:
1. Official RISC-V tests can now be run either using cocotb or using Amaranth simulator. Now we can easier debug these test -- we have all debug signals in the waveform.
2. Make the code reusable for running benchmarks using cocotb.

I also modified the script `run_tests.py`, so it also runs regression tests.

BTW, shall we already remove our own assembly tests?


test/utils/test_onehotswitch.py
```diff
@@ -3,7 +3,7 @@
```
xThaid, line 3: A completely unrelated change. I just moved it to a proper subdirectory.

tilk:
> BTW, shall we already remove our own assembly tests?

Good question. They were needed before we were able to properly run the entire core, I agree they now feel unnecessary.

Also, the failed cocotb run is probably because of the RISC-V toolchain versions in the Docker image(?).

lekcyjna123 (approved):

test/regression/pysim.py
```diff
@@ -0,0 +1,114 @@
...
+    ):
+        def f():
+            yield Passive()
+
+            while True:
+                yield from wb_ctrl.slave_wait()
+
+                word_width_bytes = self.gp.isa.xlen // 8
+
+                # Wishbone is addressing words, so we need to shit it a bit to get the real address.
```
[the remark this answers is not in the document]
tilk: That's a pretty 💩 typo ;)

scripts/run_tests.py
```diff
@@ -30,45 +35,107 @@ def flatten(suite):
...
+
+    exclude = {"rv32ui-ma_data", "rv32ui-fence_i", "rv32um-div", "rv32um-divu", "rv32um-rem", "rv32um-remu"}
+
+    return [f"regression.{test}" for test in all_tests - exclude]
+
+
+def run_regression_tests(tests: list[str], backend: Literal["pysim", "cocotb"], traces: bool, verbose: bool) -> bool:
+    short_test_names = [name[11:] for name in tests]
+
+    if backend == "cocotb":
```
tilk, line 56: Feels like the backends should be abstracted (at least separate functions?).

scripts/run_tests.py
```diff
@@ -30,45 +35,107 @@ def flatten(suite):
...
+        print("Did you forget to build them? If so, go to directory 'test/external/riscv-tests' and run 'make'.")
+        sys.exit(1)
+
+    exclude = {"rv32ui-ma_data", "rv32ui-fence_i", "rv32um-div", "rv32um-divu", "rv32um-rem", "rv32um-remu"}
+
+    return [f"regression.{test}" for test in all_tests - exclude]
+
+
+def run_regression_tests(tests: list[str], backend: Literal["pysim", "cocotb"], traces: bool, verbose: bool) -> bool:
+    short_test_names = [name[11:] for name in tests]
```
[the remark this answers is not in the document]
tilk: No! This number is there for a reason. Is it the length of some part of a path or something? This is likely to break on refactor. Avoid needless string manipulation, add some structure instead.

test/regression/pysim.py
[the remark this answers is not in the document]
xThaid: Whoops 🙃

scripts/run_tests.py
```diff
@@ -30,45 +37,113 @@ def flatten(suite):
...
+    all_tests = test.regression.test.get_all_test_names()
+    if len(all_tests) == 0:
+        res = subprocess.run(["make", "-C", "test/external/riscv-tests"])
+        if res.returncode != 0:
+            print("Couldn't build regression tests")
+            sys.exit(1)
+
+    exclude = {"rv32ui-ma_data", "rv32ui-fence_i", "rv32um-div", "rv32um-divu", "rv32um-rem", "rv32um-remu"}
+
+    return [REGRESSION_TESTS_PREFIX + test for test in all_tests - exclude]
```
tilk, line 53: My objection still stands: you add a prefix here, so that you strip it (assuming that it exists) six lines later; but the calls are related through code which are many lines later. Can this string manipulation be avoided here? I understand the prefix is just for searching (right?), so maybe only use it in the searching code?

tilk:
So that it is clearly stated for everyone: this can't be merged yet because of the RISC-V toolchain issues in the CI.

scripts/run_tests.py
xThaid: It should look better now.

.github/workflows/main.yml
```diff
@@ -9,9 +9,64 @@ on:
...
+          submodules: recursive
+
+      - name: Build riscv-tests
+        run: cd test/external/riscv-tests && make
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: "riscv-tests"
+          path: |
+            test/external/riscv-tests/test-*
```
tilk, line 32: Can the [cache action](https://github.com/actions/cache) be used instead, so that rebuilding riscv-tests can be avoided across workflow runs? (Maybe this should become a new issue?)

tilk (approved):

PR #337, opened 2023-05-12
[RFC] Multi-modules

tilk:
This PR eliminates unwanted combinational paths from the transaction manager created by that pesky implicit `If` inside the bodies of methods and transactions. The way this is achieved is by creating a `Module` look-alike, which actually contains two modules - one which includes that pesky `If`, and one which doesn't. When adding a combinational statement to the new module-like-thing, one can choose if the statement should be affected by the `If` or not.

This now seems to work (FSMs were tricky!), typing is yet to be fixed. The downside of this change is that the new module-like-thing must be used almost everywhere where `Module` was previously used.

Feedback is requested.

To do:
* [ ] fix typing
* [ ] fix documentation
* [ ] better naming scheme


coreblocks/transactions/core.py
```diff
@@ -410,6 +418,115 @@ def clear(self):
...
+            yield
+
+    @contextmanager
+    def If(self, cond: ValueLike):  # noqa: N802
+        with self.m1.If(cond):
+            with self.m2.If(cond):
+                yield
+
+    @contextmanager
+    def Elif(self, cond):  # noqa: N802
```
lekcyjna123, line 464: How about `ElifX` and `ElseX`? Aren't they needed?

lekcyjna123 (commented):
The only problem which I have with this proposition is that it isn't obvious for me, when I should use `If` or `IfX` and "comb" or "combx". In `ModuleX` names `m1`, `m2` don't suggest which module is for which use. Additionaly it is confusing for me why `IfX` operates on `m1` when `combx` is in `m2`. I would expect that `...x` will be in the same module.

Maybe my problems with understanding are because I don't remember now exact example, which was used to demonstrate problem.

tilk:
> The only problem which I have with this proposition is that it isn't obvious for me, when I should use `If` or `IfX` and "comb" or "combx". In `ModuleX` names `m1`, `m2` don't suggest which module is for which use. Additionaly it is confusing for me why `IfX` operates on `m1` when `combx` is in `m2`. I would expect that `...x` will be in the same module.
> 
> Maybe my problems with understanding are because I don't remember now exact example, which was used to demonstrate problem.

The example is right in the PR! See `Transaction.body`, `Method.body` and `Method.__call__` for example. The idea comes from that sometimes we want some statements to *avoid* some conditions. For example, call `enable` avoids the `grant`/`run` condition.

I know that the names are nonsense, I'd really like some better propositions. Some more info:

* `m1` is supposed to contain *every* condition, so that it's always safe to use for synchronous logic. All synchronous statements get added to `m1`. In particular, that's why FSMs live in `m1`: they're synchronous logic.
* `m2` is supposed to not contain some conditions (in particular, `grant` and `run`). Only combinatorial statements added to `combx` (another nonsense name) get added to `m2`.

lekcyjna123:
Thanks!  Now I understand and it looks good to me. According to naming I have following propositions:

Names for `combx`:
- `limited` - because there is limited number of conditions
- `weak` - because we have weaker assumptions (less conditions)

`m1` can be named `all_cond_module` and `m2` accordingly `limited_cond_module` or `weak_cond_module`.

I think that it would be great to have also different class name, because `ModuleX` could suggest, that we have undefined signal (e.g. in Clash functions operating on possibly undefined signals have X on the end of name). I would suggest to use name connected with our project e.g. `TransactronModule`, but this is long name and it will be often use, so maybe `TModule` or `ModuleT`?

Instead of `IfX` I would propose `PartialIf`. It is a little bit oxymoron, because condition is fulfilled or not. But in this context it can be interpreted that we apply this if, only to part of our HW description.

coreblocks/transactions/core.py
tilk: For this use case, they are not, and their existence could encourage mixing `If`/`Elif`/`Else` with `IfX`/`ElifX`/`ElseX`, which could lead to unexpected results.

This thing is a hack, and I'm not super-happy about it. But it solves the issue I have, and I don't see any better alternative.

lekcyjna123 (approved):

coreblocks/transactions/core.py
```diff
@@ -392,27 +395,141 @@ def elaborate(self, platform):
...
+    _m: "TModule"
+
+    def __init__(self, m: "TModule"):
+        object.__setattr__(self, "_m", m)
+
+    def __getattr__(self, name: str) -> _ModuleBuilderDomain:
+        if name == "av_comb":
+            return self._m.avoiding_module.d.__getattr__("comb")
+        elif name == "top_comb":
+            return self._m.top_module.d.__getattr__("comb")
```
tilk, line 408:
```suggestion
            return self._m.top_module.d.comb
```
?

lekcyjna123:
@Kristopher38 @Arusekk  @pa000  @wojpok can someone do a second review of these changes?

tilk:
For reviewers: the only changes that actually matter here are in `transactions/core.py` and `utils/_typing.py`. The line count seems large, but most of it is just search-and-replace.

Arusekk (approved):

PR #339, opened 2023-05-14
Fix debug_signals method bug introduced in #334

lekcyjna123:
I have found by coincidence, that #334 introduced a bug to `debug_signals` method and because of that core tests with traces failed. Here is a fix for that issue.

This case also show, that #316 is needed, so that problems with traces can be found before merge.


tilk:
And the type-checker didn't catch that because `auto_debug_signals` is, by necessity, weakly typed.

tilk (approved):

piotro888 (approved):

PR #345, opened 2023-05-18
One-hot implementation change

tilk:
In the current version of Amaranth we are using, the current implementation of one-hot multiplexers does not seem to generate them correctly on synthesis. However, somehow, plain switch does.

For now, I think that keeping our one-hot primitives is a good thing, if just for stating intention and error checking.


Kristopher38 (approved):
:)

xThaid (approved):
Looks good

lekcyjna123 (approved):

PR #346, opened 2023-05-18
Method argument passing optimization

tilk:
With methods calling methods, the multiplexers for method arguments can become unnecessarily large. This change reduces their size to the minimal needed. In particular, if some method is called only in one place, no multiplexer is generated, which eliminates combinational dependence of the argument on the calling transaction's `grant`.

Currently depends on #345, which I used to check for possible performance regressions.

### DIFF
--- a/coreblocks/transactions/core.py
+++ b/coreblocks/transactions/core.py
@@ -275,11 +275,17 @@ class TransactionManager(Elaboratable):
     @staticmethod
     def _method_uses(method_map: MethodMap) -> Mapping["Transaction", Mapping["Method", Tuple[ValueLike, ValueLike]]]:
         method_uses = defaultdict[Transaction, dict[Method, Tuple[ValueLike, ValueLike]]](dict)
+        enables: list[ValueLike] = []
 
-        for source in method_map.methods_and_transactions:
-            for transaction in method_map.transactions_for(source):
-                for method, use_data in source.method_uses.items():
-                    method_uses[transaction][method] = use_data
+        def rec(transaction: Transaction, source: TransactionOrMethod):
+            for method, (arg, enable) in source.method_uses.items():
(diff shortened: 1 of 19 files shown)


coreblocks/transactions/core.py
```diff
@@ -273,21 +274,41 @@ def add_edge(begin: Transaction, end: Transaction, priority: Priority, conflict:
...
+        m: Module, method_map: MethodMap
+    ) -> tuple[Mapping["Method", Sequence[ValueLike]], Mapping["Method", Sequence[ValueLike]]]:
+        args = defaultdict[Method, list[ValueLike]](list)
+        runs = defaultdict[Method, list[ValueLike]](list)
+
+        for source in method_map.methods_and_transactions:
+            if isinstance(source, Method):
+                run_val = Cat(transaction.grant for transaction in method_map.transactions_by_method[source]).any()
+                run = Signal()
+                m.d.comb += run.eq(run_val)
```
xThaid, line 304: Actually why is it needed? Can't we just make a normal python assignment `run = run_val`?

lekcyjna123 (approved):

coreblocks/transactions/core.py
tilk: It's needed to prevent duplication. Amaranth `Value`s in general correspond to Verilog *expressions*, not signals. Using the same expression in different places, as is done here, duplicates the expression. For example, take the following:

```python
v = self.i + self.i
m.d.comb += self.o.eq(v * v)
```

This is what Amaranth outputs:

image

xThaid: OK, thank you for the explanation. I wasn't paying too much attention that, but now I see that I should have.

xThaid (approved):

PR #347, opened 2023-05-18
Simultaneous transactions

tilk:
This is a proposal to extend the transaction system with simultaneity relations, which enforce that selected transactions/methods always run together in the same clock cycle. This is realized by "gluing" the simultaneous transactions into larger ones.

Two features are added to the library, hopefully encapsulating some typical use cases:

* `Connect`, which is kind of like `ConnectTrans`, but gender-reversed. `ConnectTrans` takes two methods and executes them together, with bidirectional data exchange. `Connect`, on the other hand, presents two methods, which always execute together and exchange data bidirectionally. It's kinda like `Forwarder`, but without the buffer and with a reverse channel.
* `condition`, which allows to build `If`-like statements inside transaction or method bodies, which, in contrast to typical `If`, can have different sets of methods called, and the methods not called are not blocked. The usefulness of this construction requires exploration.

This is a prototype, hopefully useful for solving some design challenges in our core. Comments are appreciated.

Depends on #337, #345 and #346.


coreblocks/transactions/lib.py
```diff
@@ -142,6 +144,58 @@ def _():
...
+        m = TModule()
+
+        read_value = Record.like(self.read.data_out)
+        rev_read_value = Record.like(self.write.data_out)
+
+        self.write.simultaneous(self.read)
+
+        @def_method(m, self.write)
+        def _(arg):
+            m.d.av_comb += read_value.eq(arg)
```
lekcyjna123, line 188: Why `av_comb` and not an `top_comb`?

coreblocks/transactions/lib.py
```diff
@@ -709,3 +763,87 @@ def elaborate(self, platform):
...
+            if not priority:
+                for transaction0 in transactions:
+                    transaction0.schedule_before(transaction)
+        transactions.append(transaction)
+
+    yield branch
+
+    if nonblocking:
+        with branch():
+            pass
```
lekcyjna123, line 844: What if user will add default branch in his code and mark `nonblocking=True`? If I correctly understand code, this will be an error.

coreblocks/transactions/lib.py
tilk: In this situation they are equivalent (as there are no further conditions here). Do you think that we should have some rules for using the additional domains? Like e.g. prefer `top_comb`, if you need conditions use `av_comb`, if you use signals across methods/transactions use `comb`.

coreblocks/transactions/lib.py
tilk: See line 833. If the user adds a default branch, the implicit default branch will not be added. (Yeah, mutating function arguments is confusing, maybe a separate variable should be added for that?)

test/transactions/test_simultaneous.py
```diff
@@ -0,0 +1,244 @@
...
+                    self.assertIsNotNone(call_result)
+                    possibles = reqv1 * [data ^ 1] + reqv2 * [data ^ 2]
+                    self.assertIn(result, possibles)
+
+        with self.run_simulation(m) as sim:
+            sim.add_sync_process(target_process)
+            sim.add_sync_process(process)
+
+
+class ConditionTestCircuit(Elaboratable):
```
lekcyjna123, line 177: This should be moved in my opinion to test_transaction_lib.py, because `condition` is defined in `transactions/lib.py`

lekcyjna123 (changes requested):
Overall looks good, but I have found one corner case which should be fixed before merge.

coreblocks/transactions/core.py
```diff
@@ -598,6 +695,52 @@ def use_method(self, method: "Method", arg: ValueLike, enable: ValueLike):
...
+        will be separately considered for execution.
+
+        Parameters
+        ----------
+        *others: Transaction or Method
+            The `Transaction`\\s or `Method`\\s to be executed simultaneously,
+            but mutually exclusive, with this `Transaction` or `Method`.
+        """
+        self.simultaneous(*others)
+        others[0].independent(*others[1:])
```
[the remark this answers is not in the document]
tilk: Independence is not transitive, but the relation you mention is implicit. Each call of `independent` defines a set for which each pair is considered independent. For example, `a.independent(b,c)` means that each of `(a, b)`, `(a, c)`, `(b, c)` are independent.

coreblocks/transactions/core.py
```diff
@@ -307,10 +307,104 @@ def _method_calls(
...
+
+        # step 2: transitivity computation
+        tr_simultaneous = set[frozenset[Transaction]]()
+
+        def conflicting(group: frozenset[Transaction]):
+            return any(tr1 != tr2 and tr1 in independents[tr2] for tr1 in group for tr2 in group)
+
+        q = deque[frozenset[Transaction]](simultaneous)
+
+        while q:
```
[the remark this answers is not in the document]
tilk: No. If this is unsatisfiable, `condition` should be unsatisfiable, too. Could you describe how you came upon this graph? What did you intend to do?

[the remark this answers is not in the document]
lekcyjna123: I don't have intend to do anything like that. I simply read the code and such example came to my mind. I see now that in such case `condition` will be unsatisfiable too. But this left me wonder what is the exact semantic of `independent`? This is not a simple reverse of `simultaneus` because it can influence transitivity of `simultaneus`.

lekcyjna123 (approved):

coreblocks/transactions/core.py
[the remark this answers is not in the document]
tilk: Its whole function is to influence the transitivity computation, so that the things that can't be simultaneous, are not. But I wouldn't recommend using `independent` in user code - exactly because its semantics is (currently?) sketchy, and it's possible it will disappear completely from future iterations. (I think that, for this reason, `independent` should probably be marked as internal.) But on the upside, the simple (but possibly troublesome) model used now was successfully used to validate the implementation mechanics of simultaneous transactions, and they seem fine.

piotro888 (approved):
Looks good from what I understand

PR #348, opened 2023-05-19
Lazy exceptions

Arusekk:
Possible simple alternative to #254. Does not work (Yet™) but shows the idea.

### DIFF
--- a/coreblocks/frontend/fetch.py
+++ b/coreblocks/frontend/fetch.py
@@ -2,7 +2,7 @@ from amaranth import *
 from coreblocks.utils.fifo import BasicFifo
 from coreblocks.frontend.icache import ICacheInterface
 from ..transactions import def_method, Method, Transaction, TModule
-from ..params import GenParams, FetchLayouts
+from ..params import GenParams, FetchLayouts, DependencyManager, VerifyBranchKey
 
 
 class Fetch(Elaboratable):
@@ -28,10 +28,25 @@ class Fetch(Elaboratable):
         self.icache = icache
         self.cont = cont
 
-        self.verify_branch = Method(i=self.gp.get(FetchLayouts).branch_verify)
+        # no early_verify_branch for now, will be needed to prevent flushing
+        # predictions on fence.i
+        fetch_layouts = self.gp.get(FetchLayouts)
+        self.verify_branch = Method(i=fetch_layouts.branch_verify_in, o=fetch_layouts.branch_verify_out)
+
+        connections = self.gp.get(DependencyManager)
+        connections.add_dependency(VerifyBranchKey(), self.verify_branch)
 
         # PC of the last fetched instruction. For now only used in tests.
         self.pc = Signal(self.gp.isa.xlen)
+
+    def use_side_fx(self, enable_side_fx: Value):
+        self.enable_side_fx = enable_side_fx
+
+    def predict_next_pc(self, m, pc):
+        """
+        The simplest predictor possible.  95%+ accuracy in real-life applications.
+        """
+        return pc + self.gp.isa.ilen_bytes
 
     def elaborate(self, platform):
         m = TModule()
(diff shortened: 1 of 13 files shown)


test/test_core.py
```diff
@@ -240,8 +240,8 @@ def test_randomized(self):
 @parameterized_class(
     ("name", "source_file", "cycle_count", "expected_regvals", "configuration"),
     [
-        ("fibonacci", "fibonacci.asm", 1200, {2: 2971215073}, basic_core_config),
-        ("fibonacci_mem", "fibonacci_mem.asm", 610, {3: 55}, basic_core_config),
+        ("fibonacci", "fibonacci.asm", 4096, {2: 2971215073}, basic_core_config),
```
Arusekk, line 243: @Kristopher38 I can recall your dissatisfaction with 15 seconds on this test case... so now it is more like 30 seconds.

Arusekk:
I had to change BranchResolvedKey to a direct call the other way round, returning info whether a misprediction happens.

Working parts:
- [x] Unrelated tests
- [x] Fibonacci tests 🎊🎉  - 🐰 - 🐇 - ­🐇🐰 - 🐇🐇🐰 - 🐇🐇🐇🐰🐰 - 🐇🐇🐇🐇🐇🐰🐰🐰
- [ ] JB unit tests
- [ ] retirement tests
- [ ] CSR tests
- [ ] speculative pc not assuming only one branch resolution and being actually correct wrt. rob_id

test/test_core.py
Kristopher38: :pensive:

Why though? Does waiting for all instructions from mispredicted path to finish make that big of a difference vs. stalling on branch and waiting for its result?

test/frontend/test_fetch.py
```diff
@@ -122,21 +136,43 @@ def accept_res_mock():
...
+                    yield
+                    yield self.enable_side_fx.eq(1)
+                    skip = True
+                continue
+            else:
+                skip = False
 
             v = yield from self.m.io_out.call()
             self.assertEqual(v["pc"], instr["pc"])
             self.assertEqual(v["data"], instr["data"])
```
Kristopher38, line 173: this duplicates somewhat what's done in the code above it for one branch

test/frontend/test_fetch.py
```diff
@@ -96,18 +100,28 @@ def cache_process():
 
                 # Speculative fetch. Skip, because this instruction shouldn't be executed.
                 if addr != next_pc:
+                    self.instr_queue.append(
+                        {
+                            "data": data,
+                            "pc": addr,
+                            "is_branch": is_branch,
+                            "next_pc": addr + self.gp.isa.ilen_bytes * (1 + is_branch),
```
Kristopher38, line 108: it eludes me why `next_pc` for a branch is addr + 2*ilen, is this particular formula needed or would any `next_pc` that's different from addr + ilen do the job?

test/test_core.py
Arusekk: Because now we have to fully execute a dozen mispredicted instructions on every loop iteration...

test/frontend/test_fetch.py
Arusekk: Pretty sure this calculation was already here when I started.

coreblocks/frontend/fetch.py
```diff
@@ -49,45 +64,61 @@ def elaborate(self, platform):
...
+
+        @def_method(m, self.verify_branch)
+        def _(from_pc: Value, next_pc: Value):
+            '''TODO: also compare rob_id to ensure correct order'''
+            mispredict = Signal()
+
+            with m.If(self.enable_side_fx):
+                # this might not work if we dropped the prediction,
+                # therefore we should maintain LRU or change attitude
+                predicted_next_pc = self.predict_next_pc(m, from_pc)
```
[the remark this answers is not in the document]
Arusekk: We could technically just peek the next pc right in the ROB (problematic if it is not there yet). Or we could pass next_pc from fetcher to decoder and from decoder to all FUs, which is quite a waste of resources.  Or we could make sure that next_pc is filled into the ROB before the next instruction even executes in the first place.  Or we should pass next_pc instead of current pc maybeee... and just cache the result, so that it does not use up precious FIFO space? Not sure.

coreblocks/structs_common/rat.py
```diff
@@ -42,8 +42,10 @@ def elaborate(self, platform):
         m = TModule()
 
         @def_method(m, self.commit)
-        def _(rp_dst: Value, rl_dst: Value):
-            m.d.sync += self.entries[rl_dst].eq(rp_dst)
+        def _(rp_dst: Value, rl_dst: Value, side_fx: Value):
+            with m.If(side_fx):
+                m.d.sync += self.entries[rl_dst].eq(rp_dst)
```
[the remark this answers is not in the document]
Arusekk: This is so that we can retrieve the contents of the old `rp` value for the supplied `rl` without breaking abstraction.  But maybe it should be a separate method, like `rob_peek` vs `rob_retire`.

coreblocks/stages/retirement.py
```diff
@@ -28,26 +29,33 @@ def __init__(
...
-        with Transaction().body(m):
+        with Transaction(name='Precommit').body(m):
             # TODO: do we prefer single precommit call per instruction?
             # If so, the precommit method should send an acknowledge signal here.
             # Just calling once is not enough, because the insn might not be in relevant unit yet.
             rob_entry = self.rob_peek(m)
-            self.precommit(m, rob_id=rob_entry.rob_id)
+            self.precommit(m, rob_id=rob_entry.rob_id, side_fx=self.side_fx)
+
+        with m.If(~self.rob_peek.ready):
```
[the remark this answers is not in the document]
Arusekk: Yes we should probably use tagged instructions somehow in the end.

test/frontend/test_fetch.py
Arusekk: We can break it down into `call_init` then `call_result` then `call_disable`...

...or make the blocking conditional:
```py
call_method = self.m.io_out.call
if instr["speculative"]:
    call_method = self.m.io_out.call_try
else:
    skip = False

v = call_method()
if v:
    assert ...
else:
    self.enable_side_fx
    yield ...
    skip = True
```

PR #349, opened 2023-05-20
Add Dockerfile for RISCV toolchain

xThaid:
So we can have an up-to-date toolchain with the standard library compiled for different extension subsets.

`rv32i-ilp32--a*zifence*zicsr;rv32im-ilp32--a*zifence*zicsr;rv32ic-ilp32--a*zifence*zicsr;rv32imc-ilp32--a*zifence*zicsr;rv32imfc-ilp32f--a*zifence` specifies that we want to build it for 5 architectures (each separated by `;`). `rv32i-ilp32--a*zifence*zicsr;` means we build a lib for arch `rv32i` with ABI `ilp32` and that the lib should be reused for architectures with `a`, `zifence`, `zicsr` extensions added (all combinations).

### DIFF
--- a/docker/riscv-toolchain.Dockerfile
+++ b/docker/riscv-toolchain.Dockerfile
@@ -1,0 +1,23 @@
+FROM ubuntu:22.10
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive \
+    apt-get install -y --no-install-recommends \
+    autoconf automake autotools-dev curl python3 bc \
+    libmpc-dev libmpfr-dev libgmp-dev gawk build-essential \
+    bison flex texinfo gperf libtool patchutils zlib1g-dev \
+    libexpat-dev ninja-build git ca-certificates python-is-python3 && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN git clone https://github.com/riscv/riscv-gnu-toolchain && \
+    cd riscv-gnu-toolchain && \
+    git checkout 2023.05.14
+
+RUN cd riscv-gnu-toolchain && \
+    git submodule update --init --recursive
+
+RUN cd riscv-gnu-toolchain && \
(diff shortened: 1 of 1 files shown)


lekcyjna123:
It looks good, but I wonder if instead of creating new docker a `Verilator.dockerfile` shoundn't be updated. If you check `.github/workflows/cocotb.yml` then you will see, that we install risc-v toolchain inside of verilator docker.

xThaid:
We discussed it in the meeting that we may separate it. It is possible to use this new container image to compile RISC-V code and use verilator container image to simulate the core.

But if more people will vote for putting it in the same dockerfile, I'll be happy to do it.

lekcyjna123 (approved):

docker/riscv-toolchain.Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    git checkout 2023.05.14
+
+RUN cd riscv-gnu-toolchain && \
+    git submodule update --init --recursive
+
+RUN cd riscv-gnu-toolchain && \
+    ./configure --with-multilib-generator="rv32i-ilp32--a*zifence*zicsr;rv32im-ilp32--a*zifence*zicsr;rv32ic-ilp32--a*zifence*zicsr;rv32imc-ilp32--a*zifence*zicsr;rv32imfc-ilp32f--a*zifence" && \
+    make -j$(nproc)
+
+RUN rm -rf riscv-gnu-toolchain
```
tilk, line 23: This does nothing useful. Each `RUN` in a Dockerfile creates a new layer, and the layers are stored separately. A `RUN` which deletes something is just creating a layer which de facto hides some files from the previous layer, it does nothing for the image size. Cleaning temporary files should be done in the same RUN that generated them.

piotro888 (approved):
Looks good, approved after @tilk s change is introduced.
I see that you selected the default `--enable-multilib` options, but without the A extension :+1: 

I was in the middle of doing the same thing, but got stuck with gcc build error, because I tried to build binutils and gcc from source manually.

docker/riscv-toolchain.Dockerfile
xThaid: Fixed.

tilk:
I've built the Docker image and uploaded it.

https://github.com/orgs/kuznia-rdzeni/packages/container/package/riscv-toolchain

PR #350, opened 2023-05-21
Fix typo in test/common.py

lekcyjna123:


tilk (approved):
And that's an uncontroversial change! ;)

xThaid (approved):

PR #352, opened 2023-05-21
CSRUnit clear

piotro888:
@Kristopher38 for your request.

With last assumptions about commiting of interrupt marked instructions it is a very straight forward change.
Tracking of instruction between CSRUnit and Retirement is no longer necessary,

### DIFF
--- a/coreblocks/structs_common/csr.py
+++ b/coreblocks/structs_common/csr.py
@@ -2,6 +2,7 @@ from amaranth import *
 from dataclasses import dataclass
 
 from coreblocks.transactions import Method, def_method, Transaction
+from coreblocks.transactions.core import Priority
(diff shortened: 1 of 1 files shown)


Kristopher38:
```
With last assumptions about commiting of interrupt marked instructions
```

Which ones? We discussed so many variants...

piotro888:
About committing of instruction with Interrupt bit set and triggering interrupt after it. 
If CSR instr is the only one in ROB (and the last one), it cannot be discarded with Interrupt, so it can be safely executed without any later tracking.

lekcyjna123 (approved):

lekcyjna123:
Very simple change, but according to arrangements done on last meeting and written in #351. Approved :)

tilk (approved):

PR #353, opened 2023-05-21
Yet another debug log fix.

lekcyjna123:


coreblocks/transactions/core.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, transactions: Iterable["Transaction"]):
         def rec(transaction: Transaction, source: TransactionBase):
             for method in source.method_uses.keys():
                 if not method.defined:
-                    raise RuntimeError("Trying to use method which is not defined yet")
+                    raise RuntimeError(f"Trying to use method '{method.name}' which is not defined yet")
                 if method in self.methods_by_transaction[transaction]:
                     raise RuntimeError("Method can't be called twice from the same transaction")
```
xThaid, line 70: nit: how about adding the method name here as well?

xThaid (approved):

coreblocks/transactions/core.py
lekcyjna123: Done

tilk (approved):

PR #354, opened 2023-05-21
Popcount

lekcyjna123:
Here is a part of multiport fifo which implements popcount, to make #260 smaller.


test/utils/test_utils.py
```diff
@@ -24,3 +28,44 @@ def test_align_to_power_of_two(self):
...
+
+class PopcountTestCircuit(Elaboratable):
+    def __init__(self, size: int):
+        self.sig_in = Signal(size)
+        self.sig_out = Signal(size)
+
+    def elaborate(self, platform):
+        m = Module()
+
+        self.sig_out = popcount(m, self.sig_in)
```
xThaid, line 41: Not really a bug, but you're overwriting a signal that is initialized in the constructor. Perhaps you wanted to write `m.d.comb += self.sig_out.eq(popcount(m,self.sig_in))`?

coreblocks/utils/utils.py
```diff
@@ -271,6 +272,47 @@ def has_explicit_shape(val: ValueLike):
         yield lhs_val.eq(rhs_val)
 
 
+def popcount(m: Module, s: Signal) -> Signal:
+    """
+    Implementation of popcount algorithm from https://en.wikipedia.org/wiki/Hamming_weight
+    """
+    if s.shape().width > 64:
```
xThaid, line 279: You can just save the width of the signal to a variable (let's say `n`) and then reuse `n` instead of `s.shape().width` and `Signal(n)` instead of `Signal.like(...)`

coreblocks/utils/utils.py
```diff
@@ -271,6 +272,47 @@ def has_explicit_shape(val: ValueLike):
...
+    if s.shape().width > 16:
+        popcount_sigs.append(popcount_sig5 := Signal.like(popcount_sigs[-1]))
+        m.d.comb += popcount_sig5.eq(popcount_sigs[-2] + (popcount_sigs[-2] >> 16))
+    if s.shape().width > 32:
+        popcount_sigs.append(popcount_sig6 := Signal.like(popcount_sigs[-1]))
+        m.d.comb += popcount_sig6.eq(popcount_sigs[-2] + (popcount_sigs[-2] >> 32))
+
+    popcount_sigs.append(popcount_res := Signal.like(popcount_sigs[-1]))
+    m.d.comb += popcount_res.eq(popcount_sigs[-2] & 0x7F)
+    return popcount_res
```
xThaid, line 313: Shouldn't the output be truncated to the correct width?

wojpok (commented):
Looks really useful, definitely ZBB will benefit from this

I have some concerns:
1. It would look better if there was a loop instead of copy-paste
2. Lack of masks past width 4 is weird
I can see why it works, but its not easy to understand and might cause some problems later

coreblocks/utils/utils.py
```diff
@@ -271,6 +272,47 @@ def has_explicit_shape(val: ValueLike):
...
+        m.d.comb += popcount_sig3.eq((popcount_sigs[-2] + (popcount_sigs[-2] >> 4)) & m4)
+    if s.shape().width > 8:
+        popcount_sigs.append(popcount_sig4 := Signal.like(popcount_sigs[-1]))
+        m.d.comb += popcount_sig4.eq(popcount_sigs[-2] + (popcount_sigs[-2] >> 8))
+    if s.shape().width > 16:
+        popcount_sigs.append(popcount_sig5 := Signal.like(popcount_sigs[-1]))
+        m.d.comb += popcount_sig5.eq(popcount_sigs[-2] + (popcount_sigs[-2] >> 16))
+    if s.shape().width > 32:
+        popcount_sigs.append(popcount_sig6 := Signal.like(popcount_sigs[-1]))
+        m.d.comb += popcount_sig6.eq(popcount_sigs[-2] + (popcount_sigs[-2] >> 32))
```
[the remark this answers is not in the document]
lekcyjna123: Folded three last 'if'-s to one for loop. I decided to leave first two 'if'-s outside loop because of the masks.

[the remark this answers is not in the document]
tilk: Is putting the masks in a list or something really that hard? You could have two loops, if merging them into one complicates code.

coreblocks/utils/utils.py
```diff
@@ -271,6 +273,46 @@ def has_explicit_shape(val: ValueLike):
...
+    if width > 2:
+        popcount_sigs.append(Signal(width))
+        m.d.comb += popcount_sigs[-1].eq((popcount_sigs[-2] & m2) + ((popcount_sigs[-2] >> 2) & m2))
+    if width > 4:
+        popcount_sigs.append(Signal(width))
+        m.d.comb += popcount_sigs[-1].eq((popcount_sigs[-2] + (popcount_sigs[-2] >> 4)) & m4)
+    # Because of assumption that this function support at most signals with 64 bit width
+    # masks in below 'if'-s are not needed (64 can be written on 7 bits, and after application of
+    # previous mask we have 8 bits already available)
+    for limit in {8, 16, 32}:
```
tilk, line 306: Why a set? Very weird.

coreblocks/utils/utils.py
```diff
@@ -271,6 +273,35 @@ def has_explicit_shape(val: ValueLike):
...
+    popcount_sigs = []
+    popcount_sigs.append(Signal(width))
+    m.d.comb += popcount_sigs[-1].eq(s)
+
+    limit = 1
+    while width > limit:
+        if width > limit:
+            mask = generate_mask(limit, width)
+            popcount_sigs.append(Signal(width))
+            m.d.comb += popcount_sigs[-1].eq((popcount_sigs[-2] & mask) + ((popcount_sigs[-2] >> limit) & mask))
```
tilk, line 297: Actually, I thought about this again, and I find this implementation to be not really fitting for implementation in a digital circuit. The Wikipedia algorithm is crafted for execution on CPUs: it assumes that you must always operate on whole words. In digital circuits, this is no longer true. In particular, this addition used here is possibly inefficient: there are carry propagation paths where there should be none, and possibly there are many adders which always add zeros. Is Yosys somehow able to optimize this? I doubt it.

If we merge this, it's only with a big honking TODO: optimize and a matching issue.

tilk: Had enough. My implementation incoming.

tilk: https://github.com/kuznia-rdzeni/coreblocks/pull/360

tilk (approved):

xThaid (approved):

wojpok (approved):

PR #355, opened 2023-05-21
Priority ordering proxy

lekcyjna123:
Here is a PR which create a method ordering proxy. It will be next used in MultiportFifo to order calls on transactron level (instead of ordering on coreblock logic level). Both `with condition` and `Connect` were useful in this PR.

This depends on: #304 ~#347~


coreblocks/transactions/lib.py
```diff
@@ -846,3 +855,53 @@ def branch(cond: Optional[ValueLike] = None):
...
+            with condition(m, priority=True) as branch:
+                for k in range(len(self.m_unordered), 0, -1):
+                    for comb_un in itertools.combinations(self._m_unordered_read, k):
+                        with branch(1):  # sic!
+                            branch_trans = TransactionBase.get()
+                            for un, ord in zip(comb_un, self.m_ordered):
+                                trans_mod = ConnectTrans(un, ord)
+                                trans_mod.t_elaborate()
+                                m.submodules += trans_mod
+                                branch_trans.simultaneous(trans_mod.transaction)
```
tilk, line 906: How is this different from just creating multiple transactions?

coreblocks/transactions/lib.py
```diff
@@ -637,16 +639,23 @@ def __init__(self, method1: Method, method2: Method):
         self.method1 = method1
         self.method2 = method2
 
+    def t_elaborate(self):
+        self.transaction = Transaction()
+
```
tilk, line 644: This is hackery. Please, don't.

coreblocks/transactions/lib.py
tilk: If I understand things correctly, what this whole thing does is:

1. It gender-changes the input, so that you provide methods. instead of taking methods. (We didn't do this consequently before, but I'd prefer to externalize gender-changing if possible. In other words, first create an Elaboratable which has the most natural interface for it, and then provide gender-changed variants if possible.)
2. For each subsequence of the input methods, it creates a transaction which connects N first methods from `m_ordered` to the methods from the subsequence. Also, these transactions are ordered.

My remarks are:

1. The use of simultaneous transactions is completely unneeded for the point 2 above. You are creating a lot of work for the transaction manager, even though you could easily express this directly - and it would probably be more readable, too.
2. This thing imposes too much ordering. If I understand correctly, you actually want the largest subsequences to be ordered before the smallest. This is a weaker, non-linear ordering.
3. Creating ConnectTrans just to mess with its insides later is very ugly. I understand you wanted to avoid repeating these four lines inside `ConnectTrans`, but this definitely is a very convoluted way to do this. An alternative would be, for example, to create a function which does this, and use it both in `ConnectTrans` and here. (We currently don't have a simple way to compose transactions like you would want. The way you proposed is weirdly convoluted and certainly not the way to go.)

lekcyjna123:
Moved to #395

Issue #280, opened 2023-04-07
Cannot specify mulltiple RS with different `rs_entries` in `fu_units_config`


piotro888:
Example config:
```python
full_core_config = CoreConfiguration(
    func_units_config=(
        RSBlockComponent([ALUComponent(), JumpComponent()], rs_entries=4),
        RSBlockComponent([MulComponent(mul_unit_type=MulType.SEQUENCE_MUL)], rs_entries=2),
        LSUBlockComponent(),
        ),
)
```

Testing fails on `assign` in full core tests.
`ValueError: Shapes not matching: lhs: unsigned(2) (rec out rs_entry_id) rhs: unsigned(1) (sig o)` in rs select Method return value.
It uses `self.layouts.select_out` which is the same for all RSes, and shouldn't be.

PR #356, opened 2023-05-21
Allow additional arguments in DependentCache

piotro888:
Allows to define keyword-only additional arguments in constructors of cached classes, without changing the old interface.
New cache entry will be created for each set of arguments for given class.
Additional arguments need to be specified each time `get` is called, allowing "parametrized" classes. 

Requirement for fix for RS layout issue (#280), which will depend on passing configurable width of Layout, that should also be accessed with cache.

### DIFF
--- a/coreblocks/params/genparams.py
+++ b/coreblocks/params/genparams.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
-from inspect import signature
-from typing import TypeVar, Type, Protocol, runtime_checkable
+from typing import TypeVar, Type, Protocol, runtime_checkable, Any
 from amaranth.utils import log2_int
 
 from .isa import ISA, gen_isa_string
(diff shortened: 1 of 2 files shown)


coreblocks/params/genparams.py
```diff
@@ -20,18 +19,38 @@
...
     def __init__(self):
-        self._depcache = {}
+        self._depcache: dict[tuple[Any, frozenset[tuple[str, Any]]], Any] = {}
 
-    def get(self, cls: Type[T]) -> T:
-        v = self._depcache.get(cls, None)
+    def get(self, cls: Type[T], **kwargs) -> T:
+        v = self._depcache.get((cls, frozenset(kwargs.items())), None)
         if v is None:
-            sig = signature(cls)
```
tilk, line 29: Why stop using `signature`?

coreblocks/params/genparams.py
```diff
@@ -20,18 +19,38 @@
...
-            if sig.parameters:
-                v = cls(self)
+            positional_count = cls.__init__.__code__.co_argcount
+
+            # first positional arg is `self` field, second may be `DependentCache`
+            if positional_count > 2:
```
tilk, line 46: What is the rationale for the restriction to keyword-only arguments?

coreblocks/params/genparams.py
```diff
@@ -20,18 +19,38 @@
...
+    -------
+    get: T, **kwargs -> T
+        Gets class `cls` from cache. Caches `cls` reference if this is the first call for it.
+        Optionally accepts `kwargs` for additional arguments in `cls` constructor.
+
+    """
+
     def __init__(self):
-        self._depcache = {}
+        self._depcache: dict[tuple[Any, frozenset[tuple[str, Any]]], Any] = {}
```
tilk, line 38: Why isn't the first element of the tuple a `Type`?

coreblocks/params/genparams.py
piotro888: because `co_argcount` allows to quickly get number of positional arguments, using `signature` would require iterating over argument list

coreblocks/params/genparams.py
piotro888: Original interface was to check if constructor needs some parameter, and if true, pass DependentCache there.
With additional arguments we can't determine if we want to pass it or not.
I decided to use keyword only arguments, to separate that additional arguments from optional cache argument. I think it also fits to interface of  `get`, the only way to pass that arguments is (`get(Layout, a=1, b=2)`), which looks like use of keyword only args.

On the other hand I don't think that interface of cache without the first argument is used anywhere in the core, so maybe it can be removed and simplify the code? DependentCache would be always passed as a first argument and then **kwargs, without any argument checking.

coreblocks/params/genparams.py
piotro888: I couldn't find the right type here, thanks

tilk (approved):

lekcyjna123 (approved):

PR #357, opened 2023-05-22
Migration to amaranth.lib.data

### LINKED ISSUE #237: Migrate to Data type
RFC with Data types have been merged, so we can migrate on it.
https://github.com/amaranth-lang/rfcs/pull/1

tilk:
As Records are basically deprecated, I've refactored the code to use amaranth.lib.data instead. Most of the changes are pretty straightforward. A weird one was RS, as it mixed combinational and sequential assignments to the same record. I still left records in Wishbone code.

Closes #237.

~Edit: damn it, the debug signals. I wonder if this can be easily fixed, or maybe I just found a major issue with `amaranth.lib.data`, to be reported?~

~Edit: this looks bad (https://github.com/amaranth-lang/amaranth/issues/790), the migration should probably be postponed.~

GTKW files and VCD dumps seem to work fine, this PR is now ready for review.


coreblocks/structs_common/csr.py
```diff
@@ -222,7 +224,7 @@ def elaborate(self, platform):
 
         current_result = Signal(self.gen_params.isa.xlen)
 
-        instr = Record(self.csr_layouts.rs_data_layout + [("valid", 1)])
+        instr = View(StructLayout({**self.csr_layouts.rs_data_layout.members, "valid": 1}))
```
whitequark, line 227: Note: this will break after https://github.com/amaranth-lang/amaranth/pull/791 is merged.

test/transactions/test_transaction_lib.py
```diff
@@ -239,20 +240,20 @@ def elaborate(self, platform):
 
         layout = data_layout(self.iosize)
 
-        def itransform_rec(m: Module, v: Record) -> Record:
-            s = Record.like(v)
+        def itransform_rec(m: Module, v: View) -> View:
+            s = View(Layout.of(v))
```
whitequark, line 244: We could potentially add `View.like(v)`.

whitequark: Ah no, this would reserve the name `like` which I don't want.

tilk: With the new metaclass-based syntax for instantiating Views backed by a Signal, this is weird, as for some Signal-constructed things `like` will work fine, and for some it will not. (In general, the RFC 15 change looks controversial to me because it breaks assumptions on how instantiating classes work in Python. This looks to me as something that might have a high WTF-factor.)

whitequark: > as for some Signal-constructed things `like` will work fine, and for some it will not

This seems to be a bug / was overlooked while implementing RFC 15.

> In general, the RFC 15 change looks controversial to me because it breaks assumptions on how instantiating classes work in Python. This looks to me as something that might have a high WTF-factor.

Normally yes, and that's listed as a potential drawback. However, Amaranth is a language-in-language, so it gets more WTF-factor budget. The variable namer, for example, breaks assumptions about encapsulation (and is *definitely* WTF-worthy), but living without it would be hard.

I will once more encourage you to participate in the RFC process to make sure your views are represented.

whitequark: I filed https://github.com/amaranth-lang/amaranth/issues/794 to track this case. It looks like you never use `data.Struct` (I'm guessing due to typing issues), so the simple fix we can do right now will not benefit you at all.

whitequark:
I've decided that I'll write an RFC soon that will cover formatting of signals in a way suitable to handling `lib.data` Views.

tilk:
I've finished updating to latest Amaranth, we were even able to remove @Arusekk's enum hack, as `amaranth.lib.enum` does the same job now.

test/transactions/test_transaction_lib.py
whitequark: I have now resolved this issue. You can use `Signal.like` freely on any value-castable object.

whitequark:
There is now support for `lib.data` fields in VCD files: https://github.com/amaranth-lang/amaranth/pull/967

whitequark:
Is there anything that's blocking you here?

tilk:
Unfortunately, generating GTKW files for views still doesn't really work for us. See the following example code:

```python
from amaranth import *
from amaranth.sim import *
from amaranth.lib.data import *

layout = StructLayout({'a': 1, 'b': StructLayout({'c': 3})})

class Foo(Elaboratable):
    def __init__(self):
        self.stuff = Signal(layout)

        m = Module()

        m.d.sync += self.stuff.a.eq(1)
        m.d.sync += self.stuff.b.c.eq(6)

        return m

dut = Foo()
sim = Simulator(dut)

def process():
    for x in range(5):
        yield

sim.add_clock(1e-6)
sim.add_sync_process(process)
with sim.write_vcd("trace.vcd", "trace.gtkw", traces=(dut.stuff, )):
    sim.run()
```

This gives an exception:
```
  File "traces.py", line 28, in <module>
    with sim.write_vcd("trace.vcd", "trace.gtkw", traces=(dut.stuff, )):
  File "/home/tilk/.localpython/lib/python3.11/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/amaranth/sim/pysim.py", line 331, in write_vcd
    vcd_writer = _VCDWriter(self._fragment,
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/amaranth/sim/pysim.py", line 68, in __init__
    for trace_signal in trace._rhs_signals():
                        ^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/amaranth/lib/data.py", line 738, in __getattr__
    raise AttributeError("View of {!r} does not have a field {!r}; "
AttributeError: View of (sig stuff) does not have a field '_rhs_signals'; did you mean one of: 'a', 'b'?
```

Alternatively, if I write `(dut.stuff.a, dut.stuff.b.c)` for the `traces` argument, I get the following in the GTKW file:
```
[dumpfile] "trace.vcd"
[dumpfile_size] 553
[treeopen] top.
@22
bench.top.stuff[3:0]
bench.top.\stuff.a
bench.top.\stuff.b[2:0]
bench.top.\stuff.b.c[2:0]
bench.top.stuff[3:0]
bench.top.\stuff.a
bench.top.\stuff.b[2:0]
bench.top.\stuff.b.c[2:0]
```

In other words, for each field I wanted to view, I got all the signals related to the whole view.

Am I doing something wrong, or is this an Amaranth issue?

whitequark:
@wanda-phi Could you take a look at what's happening here, please?

wanda-phi:
I fixed the problem with adding views to traces; making it work with adding a single field would be considerably harder and involve designing some infrastructure.

whitequark:
@tilk The crash when adding a trace is now fixed upstream and backported to the 0.4.x branch. Is this enough for your purposes? If yes, do you need a 0.4.2 release with the backport?

lekcyjna123:
Thanks for backporting the fix to the `v0.4.x` branch. I prepared already a PR #576 to switch to that version. In local tests there was no regression in behaviour so I hope that we will be able to merge it soon.

tilk:
@whitequark Looks like that's enough for us - I was able to restore our structured GTKW files to include `View` fields. The only difference is that `Record` had names, while `View` does not - which leads to a slightly worse output, as I have to group the `View` fields in unnamed groups. But I think we can live with that. I'll update this PR to master and await feedback from the rest of the team.

wanda-phi:
> @whitequark Looks like that's enough for us - I was able to restore our structured GTKW files to include `View` fields. The only difference is that `Record` had names, while `View` does not - which leads to a slightly worse output, as I have to group the `View` fields in unnamed groups. But I think we can live with that. I'll update this PR to master and await feedback from the rest of the team.

Can you elaborate what you mean here by names?

tilk:
> > @whitequark Looks like that's enough for us - I was able to restore our structured GTKW files to include `View` fields. The only difference is that `Record` had names, while `View` does not - which leads to a slightly worse output, as I have to group the `View` fields in unnamed groups. But I think we can live with that. I'll update this PR to master and await feedback from the rest of the team.
> 
> Can you elaborate what you mean here by names?

@wanda-phi Names, as in the `name` attribute of `Signal` and `Record`.

wanda-phi:
> > > @whitequark Looks like that's enough for us - I was able to restore our structured GTKW files to include `View` fields. The only difference is that `Record` had names, while `View` does not - which leads to a slightly worse output, as I have to group the `View` fields in unnamed groups. But I think we can live with that. I'll update this PR to master and await feedback from the rest of the team.
> > 
> > 
> > Can you elaborate what you mean here by names?
> 
> @wanda-phi Names, as in the `name` attribute of `Signal` and `Record`.

`view.as_value().name`?

tilk:
> > @wanda-phi Names, as in the `name` attribute of `Signal` and `Record`.
> 
> `view.as_value().name`?

Looks like that will do the trick, thank you.

transactron/core.py
```diff
@@ -1162,10 +1165,10 @@ class Method(TransactionBase):
     run: Signal, out
         Signals that the method is called in the current cycle by some
         `Transaction`. Defined by the `TransactionManager`.
-    data_in: Record, out
+    data_in: View, out
```
lekcyjna123, line 1168: Shouldn't be `MethodStruct`?

transactron/core.py
```diff
@@ -1211,13 +1212,21 @@ def __init__(
...
+        self.data_out: MethodStruct = Signal(from_method_layout(o))
         self.nonexclusive = nonexclusive
         self.single_caller = single_caller
         self.validate_arguments: Optional[Callable[..., ValueLike]] = None
         if nonexclusive:
-            assert len(self.data_in) == 0
+            assert len(self.data_in.as_value()) == 0
+
+    @property
+    def layout_in(self):
```
lekcyjna123, line 1224: Return type is missing

transactron/core.py
lekcyjna123, line 1: Naming of `RecordDict` should be corrected. But let left  it for another PR.

transactron/lib/connectors.py
```diff
@@ -46,7 +48,7 @@ def __init__(
             How many stack frames deep the source location is taken from.
             Alternatively, the source location to use instead of the default.
         """
-        self.width = len(Record(layout))
+        self.width = from_method_layout(layout).size
```
lekcyjna123, line 51: Comment for future. Can we introduce our own class `MethodLayout` which will be subclassing the `StructLayout`?

test/structs_common/test_rs.py
```diff
@@ -174,20 +170,21 @@ def simulation_process(self):
 
         # Update second entry first SP, instruction should be not ready
         value_sp1 = 1010
-        self.assertEqual((yield self.m._dut.data[1].rec_ready), 0)
+        # TODO: rec_ready inaccessible now
```
lekcyjna123, line 173: Is this "TODO" comment still valid? I don't see any `rec_ready` here.

tilk: It's a stale comment to be removed.

transactron/lib/connectors.py
tilk: We could, but would it have any advantages?

transactron/core.py
tilk: Return type is self-evident and automatically inferred.

transactron/core.py
tilk: It would probably be more readable like that.

lekcyjna123 (approved):

transactron/lib/buttons.py
```diff
@@ -17,18 +19,18 @@ class ClickIn(Elaboratable):
     ----------
     get: Method
         The method for retrieving data from the input. Accepts an empty
-        argument, returns a `Record`.
+        argument, returns a structure.
     btn: Signal, in
         The button input.
-    dat: Record, in
+    dat: RecordStruct, in
```
xThaid, line 25: Should be `MethodStruct` here and below.

xThaid (approved):

PR #359, opened 2023-05-24
Routings networks

lekcyjna123:
Here are two routing networks:
- AnyToAny network - which is the simplest implementation of routing network and should work as baseline
- Omega network - which has logarithmic depth, but fewer connections.

Depends on #304 and ~#347~


lekcyjna123:
Added as a part of #395

PR #364, opened 2023-05-26
Add support for zmmul extension in the toolchain

xThaid:
Since we don't have the division unit, we need to compile benchmarks with zmmul extension instead of M extension. And for that we need to add support for zmmul extension in our toolchain.


tilk (approved):

wojpok (approved):

tilk:
Pushed the updated Docker image.

PR #369, opened 2023-05-29
ZBB extension

wojpok:
An implementation of ZBB instructions on RV32 architecture.
Encoding needs to be verified carefully


coreblocks/fu/alu.py
```diff
@@ -33,25 +36,70 @@ class Fn(IntFlag):
...
+                (self.Fn.CTZ, OpType.UNARY_BIT_MANIPULATION_3, Funct3.CTZ, Funct7.CTZ),
+            ]
+            * self.zbb_enable
+        )
 
 
 class Alu(Elaboratable):
     def __init__(self, gen_params: GenParams, alu_fn=AluFn()):
         self.zba_enable = alu_fn.zba_enable
+        self.zba_enable = alu_fn.zba_enable
```
piotro888, line 101: did you mean
```suggestion
        self.zbb_enable = alu_fn.zbb_enable
```
?

coreblocks/fu/alu.py
```diff
@@ -86,6 +137,87 @@ def elaborate(self, platform):
                 with OneHotCase(AluFn.Fn.SH3ADD):
                     m.d.comb += self.out.eq((self.in1 << 3) + self.in2)
 
+            if self.zba_enable:
```
piotro888, line 140: also here

coreblocks/fu/alu.py
wojpok: Yup, my bad

coreblocks/params/isa.py
```diff
@@ -64,30 +64,43 @@ class Opcode(BitEnum, width=5):
...
+    SA = SUB = ANDN = ORN = XNOR = 0b0100000
     MULDIV = 0b0000001
     SH1ADD = SH2ADD = SH3ADD = 0b0010000
     BCLR = BEXT = 0b0100100
-    BINV = 0b0110100
-    BSET = 0b0010100
+    BINV = REV8 = 0b0110100
+    BSET = ORCB = 0b0010100
+    MAX = MIN = 0b0000101
+    ROL = ROR = SEXTB = SEXTH = 0b0110000
```
lekcyjna123, line 85: Why you define sextb and sexth both in Func7 and Funct12?

coreblocks/params/isa.py
```diff
@@ -64,30 +64,43 @@ class Opcode(BitEnum, width=5):
...
+    ZEXTH = 0b0000100
+    CPOP = CLZ = CTZ = 0b0011000
```
lekcyjna123, line 87: This CTZ is wrong

coreblocks/params/isa.py
wojpok: It's necessary in order to decode instruction in alu decoder. I cant use use funct12 there

I guess I could add a function that extracts funct7 from funct12 and remove all redundant funct7 definitions, but I am not sure which solution is better

coreblocks/params/isa.py
tilk: You mean, it should be 0110000, right? (You could write that yourself.)

coreblocks/fu/alu.py
```diff
@@ -86,6 +137,87 @@ def elaborate(self, platform):
...
+                with OneHotCase(AluFn.Fn.MAXU):
+                    with m.If(self.in1 >= self.in2):
+                        m.d.comb += self.out.eq(self.in1)
+                    with m.Else():
+                        m.d.comb += self.out.eq(self.in2)
+                with OneHotCase(AluFn.Fn.CPOP):
+                    m.d.comb += self.out.eq(popcount(self.in1))
+                with OneHotCase(AluFn.Fn.CLZ):
+
+                    def iter(step: int, s: Value) -> Value:
```
lekcyjna123, line 171: What is your invariant ins this function? I don't see how this function works. If I would like to get count of leading zeros of 2-bits signal, then I should pass as a step `0` (because line 185 give us `log2(2)-1 = 0`). In result I expected 2 bits result (0 - leading zeros, 1 - leading zero or two leading zeros). But If I correctly see `iter(-1)` will return 1-bit signals, `le` will be 1, so length of `high_bits` will be 1. So both `resh` and `resl` will have 1 bit length so result also will be 1 bit.

I am right? Or have I do an error in my calculations?

test/params/test_configurations.py
```diff
@@ -17,7 +17,9 @@ class ISAStrTest:
 
     TEST_CASES = [
         ISAStrTest(basic_core_config, "rv32i", "rv32", "rv32i"),
-        ISAStrTest(full_core_config, "rv32i_zicsr_zmmul_zba", "rv32_zicsr_zmmul_zba", "rv32i_zicsr_zmmul_zba"),
+        ISAStrTest(
```
lekcyjna123, line 20: Adding this exetension to `docker/riscv-toolchain.Dockerfile` isn't needed?

coreblocks/fu/alu.py
tilk: As an aside, I think that the clz/ctz functions should be defined outside ALU, like e.g. popcount.

coreblocks/frontend/decoder.py
```diff
@@ -175,6 +175,35 @@ def __init__(
...
+        Encoding(Opcode.OP, Funct3.MIN, Funct7.MIN),
+        Encoding(Opcode.OP, Funct3.MINU, Funct7.MIN),
+        Encoding(Opcode.OP, Funct3.ORN, Funct7.ORN),
+        Encoding(Opcode.OP, Funct3.ROL, Funct7.ROL),
+        Encoding(Opcode.OP, Funct3.ROR, Funct7.ROR),
+        Encoding(Opcode.OP_IMM, Funct3.ROR, Funct7.ROR),
+        Encoding(Opcode.OP, Funct3.XNOR, Funct7.XNOR),
+    ],
+    OpType.UNARY_BIT_MANIPULATION_1: [
+        Encoding(Opcode.OP_IMM, Funct3.ORCB, funct12=Funct12.ORCB),
```
piotro888, line 191: This will not work without changes in decoder.
Please add test case to `test_decoder` with example ZBB instruction.

Funct12 is not selected for OpType matching because `OP_IMM` format -> `IntstrType.I` doesn't support Funct12 matching.
It is currently only manually enabled to `SYSTEM;PRIV` instuction with this line  `m.d.comb += self.funct12_v.eq((opcode == Opcode.SYSTEM) & (self.funct3 == Funct3.PRIV))`
Temporary solution could be to add OP_IMM with all of ZBB Fuct3 fields to that line, but for longer term decoding method really needs a refactor...

piotro888 (changes requested):

coreblocks/params/isa.py
wojpok: Yikes, I need to introduce 3 new optypes to solve this

coreblocks/utils/utils.py
```diff
@@ -284,6 +286,73 @@ def popcount(s: Value):
...
+        # otherwise add 1 << (step - 1) to lower value and return
+        result = (upper_mask & upper_value) | (lower_mask & (current_bit | lower_value))
+
+        return result
+
+    value = iter(s, xlen_log)
+
+    # 0 number edge case
+    # if s == 0 then iter() returns value off by 1
+    # this swith negates this effect
```
lekcyjna123, line 320:
```suggestion
    # this switch negates this effect
```

test/utils/test_utils.py
```diff
@@ -73,3 +73,105 @@ def process(self):
...
+@parameterized_class(
+    ("name", "size"),
+    [("size" + str(s), s) for s in range(1, 7)],
+)
+class TestCountLeadingZeros(TestCaseWithSimulator):
+    size: int
+
+    def setUp(self):
+        random.seed(14)
+        self.test_number = 40
```
lekcyjna123, line 104: You will check combinational circuit many times with the same arguments for some size (e.g. size=1), so maybe it will be better to make exhaustive check?

coreblocks/utils/utils.py
```diff
@@ -284,6 +286,73 @@ def popcount(s: Value):
...
+    mask = Repl(s.any(), xlen_log + 1)
+    high_bit = 1 << xlen_log
+
+    result = ((~mask) & high_bit) | (mask & value)
+    return result
+
+
+def count_trailing_zeros(s: Value, xlen_log: int) -> Value:
+    # solution is a mirrored version of CLZs
```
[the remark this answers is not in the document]
tilk: Amaranth `Value` supports slicing. How about `sig[::-1]`?

coreblocks/utils/utils.py
```diff
@@ -284,6 +286,73 @@ def popcount(s: Value):
...
+        lower_value = iter(s[:lower_bound], step - 1)
+
+        # upper mask is on iff there are lit bits in upper half
+        upper_mask = Repl(s[lower_bound:upper_bound].any(), xlen_log + 1)
+        # otherwise lower mask is lit
+        lower_mask = ~upper_mask
+
+        # if there are lit bits in upperhalf - take result directly from recursive value
+        # otherwise add 1 << (step - 1) to lower value and return
+        result = (upper_mask & upper_value) | (lower_mask & (current_bit | lower_value))
```
tilk, line 312: Isn't this just a hand-written multiplexer, and therefore `Mux` could be used?

coreblocks/utils/utils.py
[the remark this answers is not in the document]
tilk: This function looks simpler (no `upper_bound`). So maybe keep ctz as the main implementation, and calculate clz using ctz?

To think about (as this is an optimization): one could use either clz and ctz in the ALU (as it is now), or just use one of them (ctz?), and calculate the other using the same circuit and bit reversal. The second implementation will be better in terms of resource usage, and probably shouldn't hurt fmax.

coreblocks/utils/utils.py
```diff
@@ -284,6 +286,73 @@ def popcount(s: Value):
     return sum_layers[0][0 : bits_for(len(s))]
 
 
+def count_leading_zeros(s: Value, xlen_log: int) -> Value:
```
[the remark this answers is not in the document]
tilk: I think this function should be simple to adapt to work with any bit widths, not only powers of 2; and at the same time, the edge case could be eliminated. Just concatenate a 1 to act as a guard, and don't recurse on empty range. (This will probably be easier to implement on ctz?)

test/utils/test_utils.py
tilk: But this will increase the code complexity for no good reason.

lekcyjna123 (approved):

coreblocks/utils/utils.py
```diff
@@ -284,6 +285,52 @@ def popcount(s: Value):
...
+        result = Mux(s[partition:].any(), upper_value, lower_value | current_bit)
+
+        return result
+
+    try:
+        xlen_log = log2_int(len(s))
+    except ValueError:
+        raise NotImplementedError("CountLeadingZeros - only sizes aligned to power of 2 are supperted")
+
+    xlen_log = int(xlen_log)
```
piotro888, line 313: Conversion to int is no longer needed

coreblocks/fu/alu.py
```diff
@@ -124,9 +241,10 @@ def _(arg):
 
 
 class ALUComponent(FunctionalComponentParams):
-    def __init__(self, zba_enable=False):
+    def __init__(self, zba_enable=False, zbb_enable=False):
         self.zba_enable = zba_enable
-        self.alu_fn = AluFn(zba_enable=zba_enable)
+        self.zba_enable = zbb_enable
```
piotro888, line 246:
```suggestion
        self.zbb_enable = zbb_enable
```

piotro888 (approved):

PR #370, opened 2023-05-30
Precommit

tilk:
This PR introduces precommit to the retirement stage. It signals to functional blocks that an instruction is first to be committed. This is done independent of the completion of the instruction, so that side effects can start processing on precommit. Currently, the intended users are the CSR and LSU functional blocks.

Below is the original description of the PR.

In reference to the discussion in #351, here is a simple, quick-and-dirty proposal for adding precommits to the retirement stage, which allows to stall the retirement stage (e.g. because the instruction needs to perform a side effect to be committed). I imagine that this could be useful for LSU and CSR. This is not the only possible implementation, ~another way would be to call the precommit method once (here, it returns a stall flag), and have the commit method locked until side effects are performed~ edit: locking commit method is a bad idea, as this would basically lock instructions from other FUs as well.

A discussion is welcome.

Edit: slipped a ~metaclass~ `__init_subclass__` to UnifierKey to make the subclass declarations look less noisy, sorry. I can move it to another PR later if nobody objects.

Edit: @piotro888 made me realize the implementation here is not what I wanted, ~I'll prepare a modified one later~ I made necessary modifications.

### DIFF
--- a/coreblocks/params/layouts.py
+++ b/coreblocks/params/layouts.py
@@ -316,12 +316,8 @@ class LSULayouts:
 
         self.rs_update_in = rs_interface.update_in
 
-        self.precommit_in = [
+        self.precommit = [
             ("rob_id", gen_params.rob_entries_bits),
-        ]
-
-        self.precommit_out = [
-            ("stall", 1),
         ]
 
         self.commit = [
--- a/coreblocks/params/layouts.py
+++ b/coreblocks/params/layouts.py
@@ -140,7 +140,7 @@ class ROBLayouts:
             ("done", 1),
         ]
 
-        self.retire_layout = [("rob_data", self.data_layout), ("rob_id", gen_params.rob_entries_bits)]
+        self.peek_layout = [("rob_data", self.data_layout), ("rob_id", gen_params.rob_entries_bits)]
 
 
 class RSInterfaceLayouts:
@@ -315,6 +315,14 @@ class LSULayouts:
         self.rs_select_out = rs_interface.select_out
 
         self.rs_update_in = rs_interface.update_in
+
+        self.precommit_in = [
+            ("rob_id", gen_params.rob_entries_bits),
+        ]
+
+        self.precommit_out = [
+            ("stall", 1),
+        ]
 
         self.commit = [
             ("rob_id", gen_params.rob_entries_bits),
--- a/coreblocks/params/layouts.py
+++ b/coreblocks/params/layouts.py
@@ -168,6 +168,17 @@ class RSInterfaceLayouts:
         self.update_in = [("tag", gen_params.phys_regs_bits), ("value", gen_params.isa.xlen)]
 
 
+class RetirementLayouts:
+    def __init__(self, gen_params: GenParams):
+        self.precommit = [
+            ("rob_id", gen_params.rob_entries_bits),
+        ]
+
+        self.commit = [
+            ("rob_id", gen_params.rob_entries_bits),
+        ]
+
+
 class RSLayouts:
     def __init__(self, gen_params: GenParams):
         rs_interface = gen_params.get(RSInterfaceLayouts)
@@ -316,13 +327,11 @@ class LSULayouts:
 
         self.rs_update_in = rs_interface.update_in
 
-        self.precommit = [
-            ("rob_id", gen_params.rob_entries_bits),
-        ]
+        retirement = gen_params.get(RetirementLayouts)
 
-        self.commit = [
-            ("rob_id", gen_params.rob_entries_bits),
-        ]
+        self.precommit = retirement.precommit
+
+        self.commit = retirement.commit
 
 
 class CSRLayouts:
@@ -359,3 +368,7 @@ class CSRLayouts:
         self.rs_select_out = rs_interface.select_out
 
         self.rs_update_in = rs_interface.update_in
+
+        retirement = gen_params.get(RetirementLayouts)
+
+        self.precommit = retirement.precommit
--- a/coreblocks/params/layouts.py
+++ b/coreblocks/params/layouts.py
@@ -177,10 +177,6 @@ class RetirementLayouts:
             ("rob_id", gen_params.rob_entries_bits),
         ]
 
-        self.commit = [
-            ("rob_id", gen_params.rob_entries_bits),
-        ]
-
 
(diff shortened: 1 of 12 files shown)


Arusekk:
I think you might want to consider [`__init_subclass__`](https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__) to cut down on metaclass magic.

coreblocks/structs_common/rob.py
```diff
@@ -29,11 +30,14 @@ def elaborate(self, platform):
 
         m.d.comb += self.single_entry.eq((start_idx + 1)[0 : len(start_idx)] == end_idx)
 
+        @def_method(m, self.peek, ready=self.data[start_idx].done)
```
tilk, line 33: The `ready` signal here is a mistake. I would like to have peeking avaliable if there is an entry in ROB.

coreblocks/stages/retirement.py
```diff
@@ -30,18 +34,22 @@ def elaborate(self, platform):
         m.submodules.instret_csr = self.instret_csr
 
         with Transaction().body(m):
-            rob_entry = self.rob_retire(m)
+            rob_entry = self.rob_peek(m)
+            stall = self.precommit(m, rob_id=rob_entry.rob_id)
 
-            # set rl_dst -> rp_dst in R-RAT
-            rat_out = self.r_rat_commit(m, rl_dst=rob_entry.rob_data.rl_dst, rp_dst=rob_entry.rob_data.rp_dst)
+            with m.If(~stall.stall):
```
tilk, line 40: It's possible that the `stall` signal is not really needed. If an instruction should go into commit, its result should just be announced.

coreblocks/stages/retirement.py
```diff
@@ -30,18 +34,22 @@ def elaborate(self, platform):
...
+                self.rob_retire(m)
```
Kristopher38, line 41: I had a somewhat different idea in mind and started writing it here but then I realized what I really wanted was superscalar precommit so it falls into the realm of "just an optimization(tm)". What you implemented is sane for the current implementation. We'll have to rethink this once we have an LSU that can handle multiple instructions.

tilk: Superscalar precommit/commit will be something we would like to do one day, but not now.

coreblocks/stages/retirement.py
```diff
@@ -35,21 +35,22 @@ def elaborate(self, platform):
 
         with Transaction().body(m):
             rob_entry = self.rob_peek(m)
-            stall = self.precommit(m, rob_id=rob_entry.rob_id)
+            self.precommit(m, rob_id=rob_entry.rob_id)
```
Kristopher38, line 38: Hmm, calling this more than once per rob id can have either good or bad consequences for implementing the actual handling of precommit in FUs. We shall see.

tilk: This is a decision we can easily change later.

coreblocks/params/layouts.py
```diff
@@ -316,9 +327,11 @@ def __init__(self, gen_params: GenParams):
...
         self.rs_update_in = rs_interface.update_in
 
-        self.commit = [
-            ("rob_id", gen_params.rob_entries_bits),
-        ]
+        retirement = gen_params.get(RetirementLayouts)
+
+        self.precommit = retirement.precommit
+
+        self.commit = retirement.commit
```
piotro888, line 334: Is the `commit` method longer needed?
Shouldn't the LSU STOREs be executed at pre-commit now?

lekcyjna123:
This looks good. If I correctly understand there will be also required changes in LSUDummy to change `get_results` method. There should be a check if instruction cause side effects, if yes we first should get execute of `precommit` for that instruction, else we execute load as soon as we get instruction.

coreblocks/params/layouts.py
tilk: Great question! I think they should. I haven't gotten to modifying LSU yet; it's possible that should be left to a separate PR, as LSU and its test are somewhat complicated.

tilk:
> This looks good. If I correctly understand there will be also required changes in LSUDummy to change `get_results` method. There should be a check if instruction cause side effects, if yes we first should get execute of `precommit` for that instruction, else we execute load as soon as we get instruction.

Yes, we need to add memory regions in the future. For now, it's safe to assume every load/store is a side effect.

coreblocks/stages/retirement.py
tilk: It turns out there is an issue: a ROB entry is allocated before the instruction goes into the relevant functional block. If you just call precommit once at the first possible occasion, the functional block might just miss it.

tilk: It could be possible to go around this by using an acknowledge signal from the precommit. But for now, let's leave this decision for later.

tilk:
As surprisingly little objections were raised, I'm now marking this ready for review, in the interest of getting interrupts out of the current impasse faster.

test/stages/test_retirement.py
```diff
@@ -54,6 +57,7 @@ def setUp(self):
         self.rat_map_q = deque()
         self.submit_q = deque()
         self.rf_free_q = deque()
+        self.lsu_precommit_q = deque()
```
piotro888, line 60: please remove `lsu_` prefix
commit test should be removed

coreblocks/params/layouts.py
```diff
@@ -168,6 +171,17 @@ def __init__(self, gen_params: GenParams):
...
 
+class RetirementLayouts:
+    def __init__(self, gen_params: GenParams):
+        self.precommit = [
+            ("rob_id", gen_params.rob_entries_bits),
+        ]
+
+        self.commit = [
+            ("rob_id", gen_params.rob_entries_bits),
+        ]
```
piotro888, line 182: Commit is removed from Retirement, but the layout is still there

test/stages/test_retirement.py
tilk: Done

coreblocks/params/layouts.py
tilk: Done

piotro888 (approved):
Turned out to be a very elegant change. Great to see that we are moving forward on topic on interrupts

PR #372, opened 2023-05-31
Parametrize rs_entries in RSLayouts

### LINKED ISSUE #280: Cannot specify mulltiple RS with different `rs_entries` in `fu_units_config`

piotro888:
Fixes #280.

rs_entries_bits is no longer common in the core and that caused error when trying to specify configuration with different rs sizes.

Now GenParams compute only maximum size of RS, that is used for RS Selection stage in Scheduler that is common to all RSes (and in generic tests).
Each RS uses internal layout with it own width (also for data storage), that is truncated from generic signal in RS Selection stage in Scheduler.

### DIFF
--- a/coreblocks/params/genparams.py
+++ b/coreblocks/params/genparams.py
@@ -93,7 +93,7 @@ class GenParams(DependentCache):
         # if not optypes_required_by_extensions(self.isa.extensions) <= optypes_supported(func_units_config):
         #     raise Exception(f"Functional unit configuration fo not support all extension required by{isa_str}")
 
-        self.rs_entries = 1
+        self.max_rs_entries = 1
 
         @runtime_checkable
         class HasRSEntries(Protocol):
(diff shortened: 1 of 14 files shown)


coreblocks/params/genparams.py
```diff
@@ -93,19 +93,19 @@ def __init__(self, cfg: CoreConfiguration):
...
 
-        self.rs_entries = 1
+        self.max_rs_entries = 1
 
         @runtime_checkable
         class HasRSEntries(Protocol):
             rs_entries: int
 
         for block in self.func_units_config:
             if isinstance(block, HasRSEntries):
```
tilk, line 103: Aren't functional blocks supposed to follow the "RS interface", and therefore each and every one of them should have some RS entries? If the answer is yes, why not remove the `HasRSEntries` protocol from here, and make `rs_entries` required in `BlockComponentParams`?

piotro888: That's `True`, done!

tilk (approved):

lekcyjna123 (approved):

PR #374, opened 2023-06-01
Merge master into feature/interrupts

Kristopher38:
There are conflict because master history was rewritten I guess? There should be only one conflict on `csr.py` but there are many more. @tilk this would be a lot easier if feature branches weren't protected so I could `git rebase master` on `feature/interrupts` and force push. Could you do that for me? (I don't feel confident resolving conflicts in `transactions/core.py`)


tilk:
> There are conflict because master history was rewritten I guess?

It wasn't, because it can't.

Somehow, from commit "2b793dcdabdaf74df829891b1b963ce0cb9cab44", the hashes are different. Are they different because of the signatures generated by Github? Is this the reason why this craziness happens? I don't know.

Looks like Github doesn't really support protected feature branches all that well. I've removed the restrictions.

PR #375, opened 2023-06-01
Make fetcher ready for discarding multiple instructions

Arusekk:
Split from #348 as it can probably be quite useful on its own anyway.

### DIFF
--- a/coreblocks/frontend/fetch.py
+++ b/coreblocks/frontend/fetch.py
@@ -37,22 +37,27 @@ class Fetch(Elaboratable):
         m = TModule()
 
         m.submodules.fetch_target_queue = self.fetch_target_queue = BasicFifo(
-            layout=[("addr", self.gp.isa.xlen)], depth=2
+            layout=[("addr", self.gp.isa.xlen), ("spin", 1)], depth=2
         )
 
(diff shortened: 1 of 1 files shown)


Kristopher38 (approved):
I guess this handles it better than `discard_next` and could easily be extended to handle deeper speculation depth by extending spin bit length.

coreblocks/frontend/fetch.py
```diff
@@ -66,16 +66,24 @@ def stall():
...
-                pass
-            # TODO: find a better way to fail when there's a fetch error.
-            with m.Elif(res.error == 0):
-                with m.If(is_branch | is_system):
-                    stall()
+            with m.If(spin == target.spin):
+                instr = Signal(self.gp.isa.ilen)
 
-                m.d.sync += self.pc.eq(target.addr)
+                with m.If(res.error == 0):
```
Kristopher38, line 72:
```suggestion
                with m.If(res.error != 0):
```
?

Kristopher38: Actually I'd rather keep `==` and change the order of code underneath the branches

tilk (approved):

PR #377, opened 2023-06-02
ISA String improvements

piotro888:
Provides multiple ISA string improvements
* separates machine mode instructions from I to custom XIntMachinemode and XIntSupervisor extensions
* Adds supervisor mode instructions (I had to revert sfence.vma due to issues and combinational loops in decoding, I will make separate PR with changes to decoder)
* Uses generated ISA string for assembler in core tests
* Improved `_` insertion in ISA string

### DIFF
--- a/coreblocks/params/genparams.py
+++ b/coreblocks/params/genparams.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from typing import TypeVar, Type, Protocol, runtime_checkable, Any
 from amaranth.utils import log2_int
 
-from .isa import ISA, gen_isa_string
+from .isa import ISA, gen_isa_string, Extension
 from .icache_params import ICacheParameters
 from .fu_params import extensions_supported
 from ..peripherals.wishbone import WishboneParameters
@@ -109,3 +109,10 @@ class GenParams(DependentCache):
         self.rob_entries_bits = cfg.rob_entries_bits
         self.rs_entries_bits = (self.rs_entries - 1).bit_length()
         self.start_pc = cfg.start_pc
+
+        toolchain_str_extensions = extensions
+        # Zmmul extension is not commonly available in toolchains yet, substitue it with M superset
+        if Extension.ZMMUL in toolchain_str_extensions:
+            toolchain_str_extensions ^= Extension.ZMMUL
+            toolchain_str_extensions |= Extension.M
+        self._toolchain_isa_str = gen_isa_string(toolchain_str_extensions, cfg.xlen, skip_internal=True)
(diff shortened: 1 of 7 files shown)


piotro888:
Why it doesn't support ZBA :(((
Docker container with latest toolchain could be used here, but assuming that's the latest riscv toolchain on ubuntu (?) that test would not run locally for some people.
Should I revert to manually specifying ISA string in asm test?

Arusekk (approved):

xThaid:
> Why it doesn't support ZBA :((( Docker container with latest toolchain could be used here, but assuming that's the latest riscv toolchain on ubuntu (?) that test would not run locally for some people. Should I revert to manually specifying ISA string in asm test?

I think it is reasonable to compile all assembly files in a separate job and then transfer the artifacts (like it is done in benchmarks and regression tests)

lekcyjna123:
In benchmarks we already have toolchains, which have to be started in docker locally, because of lack of out-of-the-box packages in distributions, so I think that here we can also use docker container.

coreblocks/params/genparams.py
```diff
@@ -109,3 +109,10 @@ class HasRSEntries(Protocol):
         self.rob_entries_bits = cfg.rob_entries_bits
         self.rs_entries_bits = (self.rs_entries - 1).bit_length()
         self.start_pc = cfg.start_pc
+
+        toolchain_str_extensions = extensions
+        # Zmmul extension is not commonly available in toolchains yet, substitue it with M superset
```
lekcyjna123, line 114: This is hacky. I understand that this is for better compatibility but I don't like it :/

lekcyjna123 (approved):

Arusekk:
> In benchmarks we already have toolchains, which have to be started in docker locally, because of lack of out-of-the-box packages in distributions, so I think that here we can also use docker container.

This is different, though. While benchmarks are quite CI-only and it does not matter much *how* they get generated, the tests can and should be as universal to run locally as possible. If no major distro has such a compiler in its latest LTS release, I guess we should stick to keeping compatibility, especially if the cost is as low as a single hack, named and annotated correctly.

coreblocks/params/genparams.py
xThaid: Agree. I think it may be a cause for future bugs. The compiler will think that we support division and will generate div instructions (for example in the standard library). As of now, executing an unsupported instruction (like div) results in the core deadlock

xThaid: And I don't quite get `Zmmul extension is not commonly available in toolchains yet`. `Zmmul` is already in GCC 12

PR #379, opened 2023-06-02
Instruction decoder refactor

piotro888:
Sorry, but I become so frustrated with decoder today (trying to add sfence.vma) and seeing the bug from #369, I had to do it.
 
##  The Problem (skip this section if not interested)
Previous decoder was designed with assumption that `InstrType` from RISCV can be always determined from only `Opcode` and rest of the main decoding happened based on that `InstrType`.
And that's *almost always* true. *almost*. RISCV is not that nice with those exceptions

Problem starts to appear for example with `Opcode=SYSTEM,Funct3=Priv`. These instruction uses Funct12, which is not specified in InstrType in specifaction, resulting in special logic in decoder - `self.funct12_v.eq((opcode == Opcode.SYSTEM) & (self.funct3 == Funct3.PRIV))` (and that's also true for ZBB instructions, which need to be added with all of appropiate opcodes/functs there). Then those instruction also specifically state that rd=0 and rs1=0 - another logic (in match  `& ((self.rd == 0) & (self.rs1 == 0) if enc.opcode == Opcode.SYSTEM and enc.funct3 == Funct3.PRIV else 1)`), and similiar in rs1_v decoding. 
When adding new instructions all of those conditions need to updated or new workaround logic added.

And ultimate problem with `sfence.vma`. It's the instruction with `Opcode.SYSTEM` & `Funct3.PRIV`. But as the only one from that group  (for now), instead of using `I`-like format, uses `R`-like format with Funct7 instead of Funct12, rs2 valid, rs1 valid and rd=0=invalid. Decoding this needed so many workarounds that I gave up. If we want to check for that specific instruction we need to compare Funct7 and set funct7_v for it, but if it is not that instruction then funct7_v needs to be false again, and instructions decoded with Fuct12 instead (comb loop) + all the fun with r*_v overrides.

## What is changed

* Decoding begins with matching instruction to specific `Encoding`.
* Funct3/7/**12** validity is decoded from matched `Encoding`. 
  Usage in matching statements is determined by `enc.* is not None` and after successful match assigned to _v signals.
  This allows to cleanly specify instruction encoded by different Functs for the same Opcode and normal support of Funct12.
* Registers and immediate are decoded on `InstrType` like in the old decoder. For this use case it works very well.
* No more workaround logic for main decoding (yay!).
  All of non standard operations are now handled by setting extra args in `Encoding`. It allows to cleanly override `InstrType` for single `Encoding` (instruction) in `Optype` group (for that sfence and others in future) Options are also added to verify rd/rs1=0 and disable its validity (for ECALL/EBREAK...).
* From my perspective the logic is easier to understand now.

### DIFF
--- a/coreblocks/frontend/decoder.py
+++ b/coreblocks/frontend/decoder.py
@@ -370,7 +370,7 @@ class InstrDecoder(Elaboratable):
 
         for enc in supported_encodings:
             with m.If(
-                (self.opcode == enc.opcode if enc.opcode is not None else 1)
+                (opcode == enc.opcode if enc.opcode is not None else 1)
                 & (self.funct3 == enc.funct3 if enc.funct3 is not None else 1)
                 & (self.funct7 == enc.funct7 if enc.funct7 is not None else 1)
                 & (self.funct12 == enc.funct12 if enc.funct12 is not None else 1)
--- a/coreblocks/frontend/decoder.py
+++ b/coreblocks/frontend/decoder.py
@@ -8,8 +8,6 @@ from coreblocks.params import *
 
 __all__ = ["InstrDecoder"]
 
-from coreblocks.utils import OneHotSwitchDynamic
-
 # Important
 #
 # In order to add new instructions to be decoded by this decoder assuming they do not required additional
@@ -23,10 +21,6 @@ _rd_itypes = [InstrType.R, InstrType.I, InstrType.U, InstrType.J]
 _rs1_itypes = [InstrType.R, InstrType.I, InstrType.S, InstrType.B]
 
 _rs2_itypes = [InstrType.R, InstrType.S, InstrType.B]
-
-_funct3_itypes = [InstrType.R, InstrType.I, InstrType.S, InstrType.B]
-
-_funct7_itypes = [InstrType.R]
 
 
 class Encoding:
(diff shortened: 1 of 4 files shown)


piotro888:
What's interesting Fmax reported by actions increased from current masters 52.88 MHz to 59.18 MHz

coreblocks/frontend/decoder.py
```diff
@@ -471,20 +461,13 @@ def elaborate(self, platform):
                 self.funct3.eq(Funct3.ADD),
                 self.funct3_v.eq(1),
                 self.rs1.eq(0),
+                self.rs1_v.eq(1),
```
piotro888, line 464:
```python
with m.If(opcode == Opcode.LUI):
            m.d.comb += [
                self.opcode.eq(Opcode.OP_IMM),
                self.funct3.eq(Funct3.ADD),
                self.funct3_v.eq(1),
                self.rs1.eq(0),
                self.rs1_v.eq(1),
   ]
with m.Else():
            m.d.comb += self.opcode.eq(opcode)
```
Speaking of optimization - could that statement  cause some additional performance problems? The fact that the test was passing means that all of the fields needed to be recomputed by main logic once again after LUI was changed OP_IMM (1 pass combinational loop ?). Should we worry about it or is it not worth changing? (probably would decrease code cleanliness; or that transformation could be done at entry of FU instead)

piotro888: nevermind, there are two signals - opcode and self.opcode - I used one incorrectly before

wojpok (approved):

coreblocks/frontend/decoder.py
```diff
@@ -365,35 +348,53 @@ def elaborate(self, platform):
...
-            self.rs2_v.eq(reduce(or_, (instruction_type == t for t in _rs2_itypes))),
+        ]
+
+        rd_invalid = Signal()
+        rs1_invalid = Signal()
+
+        m.d.comb += self.optype.eq(OpType.UNKNOWN)
+
+        for enc in supported_encodings:
+            with m.If(
```
lekcyjna123, line 372: This will generate bunch of priority logic. As for now we have ~100 instructions, but in future this will be much greater number. Using Switch here can be hard, but in future maybe we should create something like `ParallelIf` and substitute this `If` here.

lekcyjna123 (approved):

PR #381, opened 2023-06-06
Simultaneous transactions - different model

tilk:
This PR contains a work-in-progress, alternative model for simultaneous transactions, which is more likely to have a sensible semantics. I don't love this code, but it passes the current tests.


tilk:
As this approach is more complicated and possibly doesn't really help with anything, I'm archiving this on my fork and closing.

PR #382, opened 2023-06-06
Change get_result to send_result

tilk:
This pull request, inspired by @Arusekk's deliberations, changes the movement of results from FU to announcement from pull-mode (announcement asks units for results) to push-mode (units announce the results themselves). Such a change allows to:

* Introduce a stage between FU completion and announcement (for example, LSU / branch related actions).
* Remove FIFOs from FUs, which simplifies the code and reduces cycle delays for simple FUs, potentially improving IPC.

This is an experiment, I'm not completely happy with the idea. FIFO removal could be achieved with `Forwarder` or `Connect` (from simultaneous transactions PR), and the stage separation can be internal in the FU.


Arusekk (approved):

tilk:
That was a fun little experiment, and the results (https://github.com/kuznia-rdzeni/coreblocks/actions/runs/5446254887) are:

* IPC is only minimally improved, which probably means inter-instruction dependencies aren't a bottleneck right now.
* Fmax is drastically reduced, which makes sense - this change essentially introduces a lot of forwarding.

I don't know if it is worth it going forward with this. A FIFO external to FUs needs to be added to restore performance. I'm leaving the PR here in case some future development makes this change more desirable.

tilk:
Superseded by #792.

PR #386, opened 2023-06-10
Exception support

### LINKED ISSUE #376: Add fallback FU
We could use a separate functional unit for handling illegal instructions and instruction fetch errors; a special OpType could be used, for example providing two values like 'raise ibus error', 'raise illegal instruction'.  The FU can just set 'exception' flag unconditionally.

piotro888:
Implements exception handling compliant with @Arusekk idea. Closes #376

- [x] Exception FU
- [x] Global exception cause register
- [x] Tests
- [x] mcause CSR
-  Implement exceptions in decoder and functional units (#394)


Arusekk (approved):
Famous last words: LGTM

coreblocks/structs_common/exception.py
```diff
@@ -0,0 +1,106 @@
...
+            with m.Else():
+                m.d.comb += should_write.eq(1)
+
+            with m.If(should_write):
+                m.d.sync += self.rob_id.eq(rob_id)
+                m.d.sync += self.cause.eq(cause)
+
+            m.d.sync += self.valid.eq(1)
+
+        @def_method(m, self.get)
```
lekcyjna123, line 98: Why this method don't use `valid` bit? From outside of `ExceptionCauseRegister` you don't have now any possibility to check if results which you get are correct.

piotro888: This method is never used directly. It is used only in Retirement, when result is used when instruction has `exception` bit in  ROB set, which guarantees previous write to `ExceptionCauseRegister` and validity. This signal is only important for internal logic, and would not be checked by any caller anyway.

test/structs_common/test_exception.py
```diff
@@ -0,0 +1,101 @@
...
+from coreblocks.params import GenParams
+from coreblocks.params.configurations import test_core_config
+from coreblocks.transactions.lib import Adapter
+
+from ..common import *
+
+import random
+
+
+class ExceptionCauseTestCircuit(Elaboratable):
```
lekcyjna123, line 14: Why you create separate class for test circuit? I suppose that using `SimpleTestCircuit` and `ModuleConnector` will be easier. See: https://github.com/kuznia-rdzeni/coreblocks/blob/f5302ce9d42c5fbaecd1bf1dbe32d6ab7bd4681c/test/transactions/test_simultaneous.py#L140-L145

coreblocks/structs_common/exception.py
lekcyjna123: Ok

test/structs_common/test_exception.py
```diff
@@ -0,0 +1,101 @@
...
+
+        return False
+
+    def process_test(self):
+        saved_entry = None
+
+        yield from self.dut.rob_idx.enable()
+        for _ in range(self.cycles):
+            rob_rand = random.randint(0, self.rob_max)
+            yield from self.dut.rob_idx.set_inputs({"start": rob_rand, "end": 0})
```
lekcyjna123, line 75: Why you don't use method mock?

lekcyjna123 (commented):
Overall looks good. I left two small comments about tests

test/structs_common/test_exception.py
piotro888: I didn't know about that, nice

test/structs_common/test_exception.py
piotro888: Fixed, I start to feel more convinced about  this new :) test syntax

lekcyjna123 (approved):

coreblocks/params/isa.py
```diff
@@ -42,10 +42,16 @@ class Opcode(IntEnum, shape=5):
...
-    W = SLT = CSRRS = MULHSU = SH1ADD = CLMULR = 0b010
-    SLTU = CSRRC = MULHU = CLMULH = 0b011
+    JALR = BEQ = B = ADD = SUB = FENCE = PRIV = MUL = MULW = _EINSTRACCESSFAULT = 0b000
+    BNE = (
+        H
+    ) = (
+        SLL
+    ) = (
+        FENCEI
+    ) = CSRRW = MULH = BCLR = BINV = BSET = CLZ = CPOP = CTZ = ROL = SEXTB = SEXTH = CLMUL = _EILLEGALINSTR = 0b001
```
tilk, line 52: What's the matter with the parentheses?

coreblocks/params/layouts.py
```diff
@@ -382,3 +384,11 @@ def __init__(self, gen_params: GenParams):
         retirement = gen_params.get(RetirementLayouts)
 
         self.precommit = retirement.precommit
+
+
+class ExceptionRegisterLayouts:
+    def __init__(self, gen_params: GenParams):
+        self.get = self.report = [
+            ("cause", gen_params.supported_exception_cause_bits),
```
tilk, line 392: Shouldn't the shape of this bit be the `Cause` enum? The redundant `supported_exception_cause_bits` could then be removed. Or am I missing something?

coreblocks/params/isa.py
Arusekk: looks like `black` formatting gone wild

test/fu/test_exception_unit.py
```diff
@@ -0,0 +1,67 @@
...
+    if fn == ExceptionUnitFn.Fn.ECALL:
+        cause = Cause.ENVIRONMENT_CALL_FROM_M
+    if fn == ExceptionUnitFn.Fn.INSTR_ACCESS_FAULT:
+        cause = Cause.INSTRUCTION_ACCESS_FAULT
+    if fn == ExceptionUnitFn.Fn.INSTR_PAGE_FAULT:
+        cause = Cause.INSTRUCTION_PAGE_FAULT
+    if fn == ExceptionUnitFn.Fn.ILLEGAL_INSTRUCTION:
+        cause = Cause.ILLEGAL_INSTRUCTION
+
+    return {"result": 0} | {"exception": cause} if cause is not None else {}
```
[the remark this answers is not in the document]
piotro888: The problem is that I designed it to pass exception Cause instead of 0/1 value, to verify both `exception` bit and `ExceptionCauseRegister` method call in FU tests. `0` is also a valid cause value, so exception case would be checked. I agree it looks unelegant, but I don't have a better idea. `None` can't be passed because of typing of methods return value (maybe we could remove it? that would be usefull here)

coreblocks/params/isa.py
piotro888: It turns out to be a problem with max line length. Is it ok to duplicate `0b001` value and write part of the definition in the next line?

coreblocks/params/layouts.py
piotro888: causes circular dependency :(. Maybe enum could be moved to `params.isa`?

test/fu/test_exception_unit.py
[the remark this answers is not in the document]
Arusekk: Maybe the cause could start with 1? It would have other benefits, such as not needing to pass the exception flag where we already know the cause.

coreblocks/params/isa.py
Arusekk: It should be ok if Python allows that.  I would personally use the infamous backslash line continuation here, but I also don't remember what black does to it.

coreblocks/params/layouts.py
tilk: Looks like a good idea. Generally, moving common constants etc. to a file separate from an implementation is a valid way to avoid circular dependencies.

test/fu/test_exception_unit.py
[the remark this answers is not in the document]
piotro888: > Maybe the cause could start with 1  

Those are cause codes from RISC-V specification, that are later written directly to `mcause` CSR. It is better to not add hardware intermediate step just because of our tests.

> It would have other benefits, such as not needing to pass the exception flag where we already know the cause.

It is currently done that way. Passing `cause` in `"exception"` field is only interface for Generic FU testbench (and if it is present in result dict is checks for `exception` flag). In HW pipeline it is one bit and cause is written separetly to ExceptionCauseRegister.

coreblocks/params/isa.py
piotro888: I used the wrong slash before hmm....
But unfortunately black removes it and does the same thing. Previously proposed solution works fine, but it could become a big problem in enums with `auto()`. I will use it here

test/fu/test_exception_unit.py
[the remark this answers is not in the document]
tilk: In my opinion, this matter isn't very important. The current solution is nice in that in the common case (units without exceptions) has no added complexity.

PR #387, opened 2023-06-11
Remove dummy sync signal

xThaid:
It is just annoying to add the dummy signal everywhere.


wojpok (approved):

test/common.py
```diff
@@ -135,12 +132,26 @@ def debug_signals(self):
...
 class PysimSimulator(Simulator):
     def __init__(self, module: HasElaborate, max_cycles: float = 10e4, add_transaction_module=True, traces_file=None):
-        if add_transaction_module:
-            module = TransactionModule(module)
-
-        super().__init__(module)
+        super().__init__(TestModule(module, add_transaction_module))
 
         clk_period = 1e-6
         self.add_clock(clk_period)
```
tilk, line 157: Maybe try this instead of adding a dummy signal?

```suggestion
        self.add_clock(clk_period, if_exists=True)
```

xThaid: I tried that, but many tests (mainly testing the transaction framework) are implicitly using the clock. That is, some tests don't define any synchronous signals, but they use `TestBenchIO`, which underneath uses `yield`. Using `yield` in non-synchronous process results in 
```TypeError: Received default command from process '/home/jurb/dev/repos/coreblocks/test/transactions/test_transaction_lib.py:419' that was added with add_process(); did you mean to add this process with add_sync_process() instead?```

Arusekk (approved):

PR #388, opened 2023-06-12
Encoding uniqueness

wojpok:
An additional check that verifies whether all encodings are unique


test/frontend/test_decoder.py
```diff
@@ -263,3 +265,34 @@ def test_xintsupervisor(self):
...
+            op_code = int(instr.opcode)
+            funct3 = int(instr.funct3) if instr.funct3 is not None else 0
+            funct7 = int(instr.funct7) if instr.funct7 is not None else funct7_of_funct12(instr.funct12)
+
+            return (op_code, funct3, funct7)
+
+        for ext in _instructions_by_optype:
+            known_codes: set[tuple[int, int, int]] = set()
+
+            for instruction in _instructions_by_optype[ext]:
```
Kristopher38, line 288: One loop instead of two nested ones will suffice:
```Python
for ext, instruction in _instructions_by_optype.items():
```

Kristopher38 (changes requested):

test/frontend/test_decoder.py
tilk: ...are you sure? Because `_instructions_by_optype` is a dict of lists, and he intends to iterate over the elements of the lists.

tilk: One could avoid nested loops by using `itertools.chain`, but then the semantics of the check would change: currently this code checks uniqueness per extension, not across all extensions. (Is this the right thing? Are extensions permitted to have collisions on instruction encodings?)

test/frontend/test_decoder.py
```diff
@@ -263,3 +265,34 @@ def test_xintsupervisor(self):
...
+        def instruction_code(instr: Encoding) -> tuple[int, int, int]:
+            def funct7_of_funct12(code: Optional[Funct12]) -> int:
+                if code is None:
+                    return 0
+                else:
+                    return (int(code) & 0xFE0) >> 5
+
+            op_code = int(instr.opcode)
+            funct3 = int(instr.funct3) if instr.funct3 is not None else 0
+            funct7 = int(instr.funct7) if instr.funct7 is not None else funct7_of_funct12(instr.funct12)
```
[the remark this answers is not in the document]
tilk: How would the match statement help, other than increasing verbosity?

test/frontend/test_decoder.py
wojpok: Code was meant to check whether there are no collisions within OpType and all the unary operations from ZBB are distinguishable but I guess the functionality could be extended

test/frontend/test_decoder.py
```diff
@@ -263,3 +265,34 @@ def test_xintsupervisor(self):
...
+
+            for instruction in _instructions_by_optype[ext]:
+                code = instruction_code(instruction)
+
+                self.assertNotIn(
+                    code,
+                    known_codes,
+                    f"Instruction is not unique: OpType={str(ext)} "
+                    f"Funct3={instruction.funct3} Funct7={instruction.funct7} Funct12={instruction.funct12}",
+                )
```
tilk, line 296: This does not really do the job. If a given field (`funct3` etc.) on the `Encoding` is `None`, this means that it does not matter (works as a wildcard). You don't check for that here, failing to detect some possible collisions. Also, mapping `None` to 0, as you are currently doing, is wrong: 0 is a valid value for `funct3` etc.

wojpok: So changing ints in tuples to optional ints will work or do I need to implement different approach?

tilk: No, it will not. For example, `Encoding(Opcode.OP, Funct3.ADD, Funct7.SUB)` collides with `Encoding(Opcode.OP, Funct3.ADD)` (a purposely wrong encoding), but this approach will not detect this.

test/frontend/test_decoder.py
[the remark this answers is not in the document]
Kristopher38: it might be just my personal preference then, seeing a lot of checks for `None` done with ifs (esp. nested ones like is done here) triggers some defense mechanism in my brain

test/frontend/test_decoder.py
Kristopher38: ah nvm, still needs 2 nested loops but this is more pythonic:
```Python
for ext, instructions in _instructions_by_optype.items():
    for instruction in instructions:
```

test/frontend/test_decoder.py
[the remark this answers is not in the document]
tilk: > seeing a lot of checks for None done with ifs (esp. nested ones like is done here) triggers some defense mechanism in my brain

It's a pretty good defense mechanism, actually, as nonchalant use of optional values (especially in untyped contexts) tends to lead to "fun" bugs.

test/frontend/test_decoder.py
wojpok: So I added a test that verifies uniqueness globally and found out that I made mistake while adding ZBB instructions to decoder

lekcyjna123 (approved):

test/frontend/test_decoder.py
```diff
@@ -263,3 +265,80 @@ def test_xintsupervisor(self):
...
+            funct7 = int(instr.funct7) if instr.funct7 is not None else None
+
+            if instr.funct12 is not None:
+                funct7 = (int(instr.funct12) & 0xFE0) >> 5
+
+            return (funct3, funct7)
+
+        for ext, instructions in _instructions_by_optype.items():
+            for instruction in instructions:
+                known_codes: set[code_type] = set()
```
piotro888, line 336: Shouldn't this set be declared one line above? It is created new for each instruction.

And after that change, this test fails for ex. sll & slli, because we actually don't distinguish between them in core as instructions. We only look if immediate is present and then mux it as the second parameter instead of 0 register and treat it as OP variant (and this is working for all OP_IMM instructions).

The same is true for add and addi, but it is not detected, because of `instr.funct7 is not None else None` and in real core if fuct7 is not defined it is set to 0 and FU decoding **depends** on it to treat addi(op=OP_IMM, f3=ADD) exactly like add(op=OP_IMM, f3=ADD, f7=ADD(**=0**)).

To fix that we would need to check for imm as additional uniqueness factor in this check (because it alters the instrucion execution in that cases). But that complicates things a lot, decoder doesn't even check for imm field validity (FU just takes it if rs2=0. If rs2 is meant to be a valid 0, then imm should be 0 as well) (edit: decoder sets imm to 0 if instruciton type is R, in other cases it is filled, maybe we could just check for R/not R instruction type?)

piotro888 (changes requested):

piotro888 (approved):

Kristopher38 (approved):

PR #389, opened 2023-06-13
Division unit

wojpok:
Simple implementation of long division using method described [here](https://yuhei1-horibe.medium.com/designing-divider-213fbd32beb2).

Currently module is hard coded to calculate everything in 4 cycles, but more flexible design will be implemented later.
Signed operation will be added later as well.


tilk:
You could enable division tests in `scripts/run_tests.sh`.

tilk:
And boom, test failed. The failing one is:

```
TEST_RR_OP( 8, div, -1, -1<<31, 0 );
```

wojpok:
There is really weird edge case with the fact that -INT32_MIN == INT32_MIN. I'll fix this tomorrow

coreblocks/fu/div_unit.py
```diff
@@ -0,0 +1,141 @@
...
+
+        with Transaction().body(m):
+            response = divider.accept(m)
+            params = params_fifo.read(m)
+            result = Mux(params.rem_res, response.remainder, response.quotient)
+            # change sign but only if it was requested and sign is not correct
+            flip_sig = Mux(params.flip_sign, ~result[sign_bit], 0)
+            sign_result = Mux(flip_sig, -result, result)
+
+            result_fifo.write(m, rob_id=params.rob_id, result=sign_result, rp_dst=params.rp_dst)
```
pa000, line 126: I would move the body of this transaction directly into the `accept` method. This makes it slightly more readable and allows you to get rid of `result_fifo`.

coreblocks/fu/div_unit.py
```diff
@@ -0,0 +1,141 @@
...
+        def _():
+            return result_fifo.read(m)
+
+        @def_method(m, self.issue)
+        def _(arg):
+            m.d.comb += decoder.exec_fn.eq(arg.exec_fn)
+            i1, i2 = get_input(arg)
+
+            flip_sign = Signal(1)  # if result is negative number
+            rem_res = Signal(1)  # flag wheather we want result or reminder
```
pa000, line 82:
```suggestion
            rem_res = Signal(1)  # flag whether we want result or reminder
```

coreblocks/fu/divison/long_division.py
```diff
@@ -0,0 +1,200 @@
...
+        Calculated n bits of quotient
+    remainder: Signal
+        Calculated remainder
+    partial_remainder: Signal
+        Calculated partial reminder
+    """
+
+    def __init__(self, count: int, size: int, partial_remainder_count: int = 0):
+        self.size = size
+        self.n = count
```
pa000, line 50: `self.count` would be a more consistent name. Or maybe even `step_count`.

coreblocks/fu/divison/long_division.py
```diff
@@ -0,0 +1,200 @@
...
+
+        dividend = Signal(unsigned(xlen))
+        divisor = Signal(unsigned(xlen))
+
+        quotient = Signal(unsigned(xlen))
+        remainder = Signal(unsigned(xlen))
+
+        stage = Signal(unsigned(xlen_log + 1))
+
+        # reseting
```
pa000, line 157:
```suggestion
        # resetting
```

coreblocks/fu/divison/long_division.py
```diff
@@ -0,0 +1,200 @@
...
+        with m.If(~ready & (stage < self.stages)):
+            special_stage = (self.stages == stage + 1) & self.odd_iteration
+
+            # assigning inputs to recursive divider
+            m.d.comb += divider.divisor.eq(divisor)
+            m.d.comb += divider.dividend.eq(dividend[xlen - self.ipc :])
+            m.d.comb += divider.input_remainder.eq(remainder)
+
+            # dividend is a shift register
+            # so in each iterations upper bits are fed into recursive divider
```
pa000, line 184:
```suggestion
            # so in each iteration upper bits are fed into recursive divider
```

coreblocks/fu/divison/long_division.py
```diff
@@ -0,0 +1,200 @@
...
+            m.d.sync += dividend.eq(dividend << self.ipc)
+
+            # if we are in the last stage and uneven amaount of bits needs to be handled
```
pa000, line 187:
```suggestion
            # if we are in the last stage and uneven amount of bits needs to be handled
```

coreblocks/fu/divison/long_division.py
```diff
@@ -0,0 +1,200 @@
+from amaranth import *
+
+from coreblocks.params import GenParams
+from coreblocks.transactions import *
+from coreblocks.transactions.core import def_method
+from coreblocks.fu.divison.common import DividerBase
+
+"""
+Alogorithm - multi-cycle array divider
```
pa000, line 9:
```suggestion
Algorithm - multi-cycle array divider
```

coreblocks/fu/divison/long_division.py
```diff
@@ -0,0 +1,200 @@
...
+
+        # recursive module
+        m.submodules.rec_div = rec_div = RecursiveDivison(
+            self.n - 1, self.size, partial_remainder_count=self.partial_remainder_count - 1
+        )
+
+        m.d.comb += rec_div.dividend.eq(self.dividend)
+        m.d.comb += rec_div.divisor.eq(self.divisor)
+
+        # Signle step as described in article
```
pa000, line 89:
```suggestion
        # Single step as described in article
```

coreblocks/fu/div_unit.py
Kristopher38: Even
```suggestion
            rem_res = Signal(1)  # flag whether we want result or remainder
```
;)

coreblocks/fu/div_unit.py
wojpok: Currently module follows same structure as multiplication unit and I'm honestly not sure which version would better in the long run

coreblocks/fu/divison/long_division.py
```diff
@@ -0,0 +1,200 @@
...
+            m.d.sync += ready.eq(0)
+
+        # returning results
+        @def_method(m, self.accept, ready=(~ready & (stage == self.stages)))
+        def _(arg):
+            m.d.sync += ready.eq(1)
+            return {"quotient": quotient, "remainder": remainder}
+
+        # performing calculations
+        with m.If(~ready & (stage < self.stages)):
```
lekcyjna123, line 175: Maybe use `stage != self.stages`? This should be cheaper in HW.

lekcyjna123 (changes requested):
I suppose that `docker/riscv-toolchain.Dockerfile` also should be updated.

tilk:
> I suppose that `docker/riscv-toolchain.Dockerfile` also should be updated.

I don't think this should be a requirement for this PR.

lekcyjna123:
> Extending Docker could be done in seperate PR alongside with enabling ZBS and ZBC FUs because full config is in fact not a full config

Ok. But please answer rest of my comments.

lekcyjna123 (approved):

tilk:
I've run the synthesis benchmark (https://github.com/kuznia-rdzeni/coreblocks/actions/runs/5425431940/jobs/9866166080), and it looks like this PR reduces the Fmax to 38.4 MHz. It might be worthwhile to check if reducing "ipc" helps.

coreblocks/fu/divison/long_division.py
```diff
@@ -0,0 +1,207 @@
...
+        self.step_count = step_count
+        self.partial_remainder_count = partial_remainder_count
+
+        self.divisor = Signal(unsigned(size))
+        self.dividend = Signal(unsigned(size))
+        self.input_remainder = Signal(unsigned(size))
+
+        self.quotient = Signal(unsigned(size))
+        self.remainder = Signal(unsigned(size))
+        self.partial_reminder = Signal(unsigned(size))
```
tilk, line 59:
```suggestion
        self.partial_remainder = Signal(unsigned(size))
```

coreblocks/fu/divison/long_division.py
```diff
@@ -0,0 +1,207 @@
...
+
+        dividend = Signal(unsigned(xlen))
+        divisor = Signal(unsigned(xlen))
+
+        quotient = Signal(unsigned(xlen))
+        remainder = Signal(unsigned(xlen))
+
+        stage = Signal(unsigned(xlen_log + 1))
+
+        @def_method(m, self.clear)
```
tilk, line 157: Needs to be last, otherwise, when `issue` is run concurrently with `clear`, `issue` will "win". (Clears are not really tested anywhere yet, but we are doing them in preparation for interrupts.)

coreblocks/fu/divison/long_division.py
```diff
@@ -0,0 +1,207 @@
...
+        def _():
+            m.d.sync += remainder.eq(0)
+            m.d.sync += quotient.eq(0)
```
tilk, line 160: Not really needed, these are not used for ready signals and are cleared on issue.

tilk (approved):

PR #393, opened 2023-06-17
Layout for exceptions

Arusekk:
Part of #386 eventually needed for further work on #348.

Cc: @piotro888


test/fu/test_alu.py
```diff
@@ -285,7 +285,7 @@ def setUp(self):
                     "imm": data2 if data2_is_imm else 0,
                 }
             )
-            self.responses.append({"rob_id": rob_id, "result": result, "rp_dst": rp_dst})
+            self.responses.append({"rob_id": rob_id, "result": result, "rp_dst": rp_dst, "exception": 0})
```
Arusekk, line 288: I'd like to note the whole class here is a leftover, and should be considered for removal. There is a generic FU test for ALU and then this — a similar test that only exercises ADD instructions and duplicates code from the excellent generic implementation made in #340.

test/fu/functional_common.py
```diff
@@ -118,7 +118,7 @@ def setUp(self):
                     "pc": pc,
                 }
             )
-            self.responses.append({"rob_id": rob_id, "rp_dst": rp_dst} | results)
+            self.responses.append({"rob_id": rob_id, "rp_dst": rp_dst, "exception": 0} | results)
```
Arusekk, line 121: Note: this way every fu test can override the exception field, but it defaults to 0 when not specified.

piotro888 (approved):
LGTM; I think that note about exception field overriding in fu test deserves a comment in code

lekcyjna123 (approved):

PR #396, opened 2023-06-26
Extend decoder with part of V instructions.

lekcyjna123:
Hi,

here is a part of V instructions which I plan to merge tomorrow to development branch. Most important change introduced by this commit is existence of new field in output of decoder "RegisterType", which indicates if register is scalar (X) or vector(V) in future this can be used to add support for floating point registers.

I have to add decoding tests before merge. They will be generated using llvm based on `test/asm/v_instr.asm`.


lekcyjna123:
Yes. In case when RegisterType will be `X` register will be renamed by scalar register renaming. In other cases it will be forwarded without changes to FU and in case of vector registers renaming will be done in vector FU.

lekcyjna123:
I have observed that adding decoding of 1/3 vector instructions to our current decoder extends simulation time by factor of 2 (calculated over tests which decodes instructions from extension I). So I started benchmarks for this PR without and with vector extension activated. It looks like my changes haven't big impact on Fmax (59MHz vs 57MHz), but in near future a decoder will need a great rework.

Without V:
https://github.com/kuznia-rdzeni/coreblocks/actions/runs/5391792933/jobs/9789181833
With V:
https://github.com/kuznia-rdzeni/coreblocks/actions/runs/5391851409/jobs/9789330046

PR #398, opened 2023-06-28
DummyLSU: Delay pushing store result until execution

piotro888:
Small change, useful for implementing reporting exceptions on store in #394, exported as independent change for eventual discussion.
This PR unifies LOAD and STORE instruction result pushing/accepting to point where instruction in actually executed. It will be useful to be able to pass pipeline `exception` flags in STORE instructions. Additional benefit is slightly easier to understand logic and it *may* be helpful in implementing IO reads.
Downside - increased dealy on stores as we need to wait for result. Previously they were executing also after instruction was retired? But this is the only way to handle exceptions on stores - to wait until completion. (Additonal different delay between FU->RET is negligible). I don't think that's a big issue for now as dummyLSU will be replaced, but may be useful note for future. Maybe we could specify Exception-less memory regions, where STORES could execute 'detached'?

### DIFF
--- a/coreblocks/lsu/dummyLsu.py
+++ b/coreblocks/lsu/dummyLsu.py
@@ -45,7 +45,6 @@ class LSUDummyInternals(Elaboratable):
         self.get_result_ack = Signal()
         self.result_ready = Signal()
         self.execute_store = Signal()
-        self.store_ready = Signal()
 
     def calculate_addr(self):
         """Calculate Load/Store address as defined by RiscV spec"""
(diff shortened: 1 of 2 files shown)


coreblocks/lsu/dummyLsu.py
lekcyjna123, line 1: Line 167:
```py
            with m.State("StoreWaitForExec"):
                with m.If(self.get_result_ack):
                    m.d.sync += self.result_ready.eq(0)
```
This `If` is no longer needed in my opinion.

tilk (approved):
As usual, functionality and simplicity is better than the little bit of lost performance here. Exception-less regions might be an useful workaround, if we decide to keep a simplistic LSU for generating tiny cores - but that's _an optimization_.

lekcyjna123 (approved):

coreblocks/lsu/dummyLsu.py
piotro888: fixed, thanks!

lekcyjna123:
> Downside - increased dealy on stores as we need to wait for result

I wouldn't care about that now. From one side DummyLSU should be removed and refactored. From the other in normal LSU we will have a store queue, so we will be able to handle loads before committing stores, so they could stay in pipeline for long time, without negative impact for the core.

> Previously they were executing also after instruction was retired? 

Yes this was an artifact of refactor to add precommit support, which have taken infrastructure of old LSU directly.

> Maybe we could specify Exception-less memory regions, where STORES could execute 'detached'?

This is an interesting idea to check while implementing normal LSU.

piotro888:
IPC in benchmarks increased, good, but why??

PR #400, opened 2023-06-29
[RFC] Controversial make dict

lekcyjna123:
I have recently observed that in many places in tests we have pattern:
`{"var1" : var1, "var2" : var2, "var3" : var3 }`
This cause a lot of boilerplate, so I started to think about reducing it. Here is a PoC, which allow to write:
`make_dict(var1, var2, var3)`
In interpreter it works, but after substituting in tests there is a fail (but I should be able to correct this).

So I would like to ask if there is a sense to invest time in it, taking into account that current solution is very hacky?(@Arusekk maybe you can do that better)

Other proposition which I have is to introduce a little bit more costly syntax:
`make_dict("var1", "var2", "var3")`
but such `make_dict` will be a lot easier to implement (get previous frame, iterate over `locals` and get requested values).

What do you think? Is second proposition better? Or maybe both are not acceptable hackery?

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -1,3 +1,6 @@
+import inspect
+import dis
+import re
 import unittest
 import os
 import functools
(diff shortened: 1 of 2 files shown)


Arusekk:
This is so [JavaScripty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) I am both amazed and scared.  I both love its simplicity and hate its implications.  This will get unmaintainable like... immediately.  But if I were to pursue this, I would rather use the method used by Amaranth to infer variable names (iterating the code object pseudo-instructions, but in the reverse order, looking for `len(args)` of `LOAD_FAST` opcodes).

tilk:
I kinda agree with @Arusekk that this is kinda impressive, but this has a potential to have a high WTF factor. Verbosity is somewhat reduced at a cost of increased conceptual complexity.

PR #401, opened 2023-06-30
Scheduler preparation for vector extension

lekcyjna123:
In this PR I prepare scheduler for handling vector extension. I am not sure if this is the most optimal way, but relatively easy to implement and should be scalable over extension F. I extended register identifiers with a type which says in which register file we should look for that register. As for now there are two types "X" and "V". These tags are pushed throw whole scheduler.

Additionally there is added FifoRS and VXRS, so logically the first elements of vector block.

Changes:
- rename "csr" to "imm2" (will be used by both CSR and vectors)
- extend register with type
- add FifoRS
- move mod_incr to standalone utils function
- create generic generate test data functions

_Optimisation_ to be implemented under separate PR:
- standard X FU always send data to X registers so there is no need to pass dst reg type throw them

### DIFF
--- a/coreblocks/frontend/decode.py
+++ b/coreblocks/frontend/decode.py
@@ -60,7 +60,7 @@ class Decode(Elaboratable):
                         "rl_s2_rf": Mux(instr_decoder.rs2_v, instr_decoder.rs2_rf, 0),
                     },
                     "imm": instr_decoder.imm,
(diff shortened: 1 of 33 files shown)


lekcyjna123:
ToDo:
create an issue to optimise RegisterType usage

PR #407, opened 2023-06-30
Synthesize different core versions

### LINKED ISSUE #402: Synthesis benchmark for full core
Currently, only the basic core is measured. We should do this for full core also, maybe synthesize both.

tilk:
Synthesis and verilog generation are currently kinda copy-pastey, and that's fine for now - cleaning this up (maybe unifying the two?) can be done later.

Fixes #402.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -8,6 +8,9 @@ on:
 
 jobs:
   synthesis:
+    strategy:
+      matrix:
+        config: [basic, full]
     name: Synthesis benchmarks
     runs-on: ubuntu-latest
     timeout-minutes: 20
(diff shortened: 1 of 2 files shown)


tilk:
Action run is here: https://github.com/kuznia-rdzeni/coreblocks/actions/runs/5424401768

There is surprisingly little difference between `basic` and `full` configurations, which is weird. However, the carry, RAM and DFF fields are different, and the synthesis time for `full` was longer.

lekcyjna123 (approved):

tilk:
By the way: current configurations look weird. In particular, the ROB is quite large (128 entries) compared to RS sizes (4 and 2 entries).

piotro888 (approved):
Approved.

I also think that core configurations need rethinking (and optimization). Adding additional core configuration between basic and full would be helpful. 
* Tiny core config could probably use RV32E and smaller struct sizes when implemented
* Basic config is good as is - standard, usable rv32i good for baseline benchmarks and simple use cases (probably without priv mode - requires zicsr, but thats discussable)
* Some new config that supplies priv mode and some common set of useful extenstions, to be nice for generic usage
* And current Full config that is helpful for arch tests and testing possibilities
* We could also add some configurations accoring to RISCV Profiles that was stated in some issue.

PR #415, opened 2023-07-02
Change default value for condition nonblocking parameter

lekcyjna123:
I have observed that blocking condition is used by me more often than nonblocking. So here is a proposition to change default value.


tilk:
This is not about what is used more, but what is more surprising. I thought that blocking by default is surprising - if you have a body which contains a condition *and* something else, this unrelated something else might not execute because of the condition.

lekcyjna123:
On the other hand you have methods inside a condition and you expect that whole transaction don't start until methods in one of branch are all ready, but in default behaviour transaction will start even when all methods are inactive and will execute something else which is in transaction body.

tilk:
Indeed, you're right - the matter isn't that simple. Another point for blocking by default is that nonblocking behavior is currently implemented by adding an additional transaction which isn't explicitly written by the programmer.

tilk:
Another thought - is `nonblocking` actually useful? Maybe it should be removed completely.

lekcyjna123:
Till now I used `condition` in 9 places:
- in 8 places I use `nonblocking = False`
- in 1 place value of `nonblocking` doesn't matter

So as for now I don't have any example of nonblocking usefulness. But in my opinion it should stay, because we don't know what we will be needing in future.

tilk:
> So as for now I don't have any example of nonblocking usefulness. But in my opinion it should stay, because we don't know what we will be needing in future.

In case this behavior is needed, one simply adds a default branch. That's what the code does. Removing `nonblocking` doesn't remove any functionality at all, it simply makes this case more verbose (which may be a good thing?).

tilk:
Let's go ahead with this for now.

tilk (approved):

Arusekk (approved):

PR #417, opened 2023-07-02
Vector frontend

lekcyjna123:
This PR will be introducing frontend blocks for vector unit. By default they will be part of vector functional block, so they will be active after vector instruction pass scalar RS.

Added:
- VectorInputVerificator (to check if incoming instruction is correct, or we should raise an exception)
- VectorStatus (which handles changes in vector CSRs and add CSR snapshot to each instruction passed through it)
- SuperscalarFreeRF (support allocating and deallocating more registers at once)
- MultiPriorityEncoder (extension of priority encoder from amaranth to get more outputs in one cycle)
- SuperscalarRATs

Other features:
- added `Now()` command to simulation to get current cycle
- added special argument `_now` to get current time in method mock

### DIFF
--- a/coreblocks/frontend/decoder.py
+++ b/coreblocks/frontend/decoder.py
@@ -354,6 +354,8 @@ _instructions_by_optype = {
     OpType.V_CONTROL: [
         Encoding(Opcode.OP_V, Funct3.OPCFG),
     ],
+    OpType.V_MEMORY: [
+    ],
 }
 
 
--- a/coreblocks/frontend/decoder.py
+++ b/coreblocks/frontend/decoder.py
@@ -354,8 +354,7 @@ _instructions_by_optype = {
     OpType.V_CONTROL: [
         Encoding(Opcode.OP_V, Funct3.OPCFG),
     ],
-    OpType.V_MEMORY: [
-    ],
+    OpType.V_MEMORY: [],
 }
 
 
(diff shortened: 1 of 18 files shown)


tilk:
I've noticed that you make a lot of PRs not intended for the main branch, which are then merged. This confuses me, are we supposed to read them?

tilk (approved):
I was reviewing another PR, wtf.

lekcyjna123:
> I've noticed that you make a lot of PRs not intended for the main branch, which are then merged. This confuses me, are we supposed to read them?

They have three goals:
- first and most important - if there is anyone who want to read my code, this is a possibility for them to do this and analyse code in more or less small and consistent parts - as it was said some time ago it is easier to read smaller parts
- this also documents progression of implementation and present a middle stages before squashing and rebasing to development branch
- it allows me to separate work for subparts and motivates me to deliver a ready code to development branch - with all tests, docstrings and after self review, during which I remove some unneeded parts


So it would be nice if someone would read them, but I know how our review is working and there can pass a lot of time before someone check a PR. Therefore I merge them to feature branch when I need to prepare a depended changes, to don't have merge conflicts. On solving merge conflicts between my own PR's which were created in second part of semester ("to have smaller PRs which will be easy to review") I spend two days, so I don't plan to repeat that error.

lekcyjna123:
Most of code from this PR is ready. I have to:
- fix VStatusUnit test
- remove unneeded comments/prints
- write docs strings
I plan to do above points tomorrow.

PR #421, opened 2023-07-03
Use Zbc and Zbs in full core

tilk:
Added Zbc and Zbs support to the full core configuration, it therefore now has full B extension support.

A problem with handling ISA strings was found - the B extension is not generated, even though it's fully supported.

Rationalizing core configurations is left for later.


piotro888:
Generation problem Fixed in #424

piotro888 (approved):

lekcyjna123 (approved):

PR #425, opened 2023-07-03
Single caller method check

### LINKED ISSUE #416: Add support for one caller methods in typing
Since we can make methods ready based on input data in one caller methods, such usage is being added continuously in new places to make some _optimisations_. As for now we don't have any support for such methods in typing system and they are treated as normals methods. Only docstrings can said that if method will be connected to two or more callers an combinational loop will occur. So to prevent such kind of errors I would like to ask for introducing OneCallerMethod type to our typing system.

tilk:
For @lekcyjna123. For some weird, but kinda sensible use cases, it makes sense to have a method which can be called at most once. In such a case, it's better to check that assumption to avoid surprises later.

Fixes #416.


tilk:
Didn't really work for the intended use case. To be verified, changing to draft for now.

lekcyjna123 (approved):

piotro888 (approved):

PR #427, opened 2023-07-04
Simplified DummyLSU

tilk:
You wanted a simplified DummyLSU? You've got a simplified DummyLSU. Now with only two states in the FSM!

This also slightly increases IPC and reduces the core size.


lekcyjna123 (approved):
Now it looks much better.

piotro888 (approved):
Simpler LSU - simple approve

Very nice changes.
IPC results are interesting, maybe running operations immediately was not that great idea (this time probably because on jump mispredicts after LOAD).  Or is that because of other simplifications in FSM?

PR #430, opened 2023-07-08
Use all available cores when compiling cocotb regression tests with verilator

Kristopher38:
I had to run cocotb tests locally and this bugged me (as I am a spoiled 12-core cpu user). Not sure how many cores github CI runners have available but if it's more than one, then this should also help speed up the CI flow.

### DIFF
--- a/scripts/run_tests.py
+++ b/scripts/run_tests.py
@@ -54,7 +54,8 @@ def load_regression_tests() -> list[str]:
 
 
 def run_regressions_with_cocotb(tests: list[str], traces: bool) -> bool:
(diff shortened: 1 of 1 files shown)


lekcyjna123 (approved):
It can be deliberated if `sched_getaffinity` CPU number should be used to spawn a new process, but it is rather a discussion for SO, so I am approving :)

Kristopher38:
From setaffinity manpage:
```
A child created via fork(2) inherits its parent's CPU affinity
mask.  The affinity mask is preserved across an execve(2).
```
;)

tilk (approved):

PR #432, opened 2023-07-10
Method run debugging

tilk:
This PR adds debug signals to `TransactionManager` which expose parts of transaction scheduling logic. This should help debug situations when it's unclear why a given method doesn't run.

### DIFF
--- a/coreblocks/transactions/core.py
+++ b/coreblocks/transactions/core.py
@@ -800,8 +800,8 @@ class TransactionBase(Owned):
 
     @property
     def owned_name(self):
-        if self.owner is not None:
-            return f"{self.owner.__class__.__name__}.{self.name}"
+        if self.owner is not None and self.owner.__class__.__name__ != self.name:
+            return f"{self.owner.__class__.__name__}_{self.name}"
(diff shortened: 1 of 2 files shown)


Kristopher38 (commented):
I had the look at names generated in traces, this looks much better already. Would it be possible to expand the names a little bit to avoid situations like this?
image

 I was thinking of suffixing these with the name of the module they're in (i.e. using `owner.__module__` like is done [here](https://github.com/kuznia-rdzeni/coreblocks/blob/cf92760286bb09504735b01dbc71b62a7bd1f05b/coreblocks/transactions/core.py#L966)) so they're unique between files (they don't have to be unique in a particular file since you can distinguish which is which by their order of appearance in the file). This would be *super* helpful

tilk:
> I had the look at names generated in traces, this looks much better already. Would it be possible to expand the names a little bit to avoid situations like this?

This FU is not really about naming, but as it introduces `owned_name` for other reasons, might as well use it for signal names.

Kristopher38 (approved):

PR #433, opened 2023-07-10
Improve GenericFunctionalTestUnit

### LINKED ISSUE #341: Improve GenericFunctionalTestUnit
Possible improvements:

* Rename `run_pipeline` as it doesn't have anything to do with pipelines.
* Add a way to test pipelining FUs.
* Rename the class, as its name is really misleading. It should probably be named `TestCaseFU` or something.
* Clean up the `methodName` cruft, it serves no purpose.
* Don't pass weird `seed`s, or remove the `seed` argument altogether.
* A sensible default should be present for `number_of_tests`, instead of changing it for every use. A multiplier could be used so that each operation gets enough testing, no matter the number of `operations`.

tilk:
Generic testing of functional units is nice, but it was kind of smelly in places. This PR refactors the FU tests a little.

Changes:
* Constructor no longer overloaded. Constructor parameters are now attributes. This eliminates the `method_name` weirdness, and also simplifies interoperation with `parameterized_class`.
* `number_of_tests` now specifies the number of tests per operation, which is much more useful - it guarantees that operations are tested equally, and the number does not need to be tweaked.
* `if` on `Fn` changed to `match`.
* Added a generic pipelining test.
* Reduced verbosity in operation lists.
* Improved typing.

Fixes #341.

### DIFF
--- a/test/fu/functional_common.py
+++ b/test/fu/functional_common.py
@@ -1,7 +1,7 @@
 from itertools import product
 import random
 from collections import deque
-from typing import Callable, Generic, TypeVar
+from typing import Generic, TypeVar
 
 from amaranth import Elaboratable, Module
 from amaranth.sim import Passive
@@ -57,12 +57,12 @@ class FunctionalUnitTestCase(TestCaseWithSimulator, Generic[_T]):
     Common test unit for testing functional modules which are using @see{FuncUnitLayouts}.
     For example of usage see @see{MultiplierUnitTest}.
 
-    Parameters
+    Attributes
     ----------
     operations: dict[_T, dict]
         List of operations performed by this unit.
     func_unit: FunctionalComponentParams
-        Class of functional unit to be tested.
+        Unit parameters for the unit instantiated.
     expected: Callable[[int, int, int, int, _T, int], dict[str, int]]
         Function computing expected results
         (input_1, input_2, input_imm, pc, operation_key_from_operations, xlen) -> results dict.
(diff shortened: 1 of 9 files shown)


Kristopher38:
Can I have a request? Not sure if this is in scope for this PR but...

Generic FU tests are fine until you want to test for something nonstandard. Recently I wanted to add some extra processes for branch unit in `feature/interrupts`, sometimes replacing the predefined ones, sometimes reusing them - both cases are useful. I also wanted to retain the input generation. Suffice to say the results [look a bit substandard](https://github.com/kuznia-rdzeni/coreblocks/blob/feature/interrupts/test/fu/functional_common.py#L175) with usage that [can look like this](https://github.com/kuznia-rdzeni/coreblocks/blob/feature/interrupts/test/fu/test_jb_unit.py#L159). I'm not entirely happy with how it turned out but it present the general idea and I'm hoping it can be done better.

tilk:
Moved the process definitions outside of the test running function, this should allow you to just write your own if needed.

piotro888 (approved):
Very nice changes!

The only little concern I have is that `zero_imm` override in unit tests classes looks more like some unused variable from test, but probably nothing could be done about that.

tilk:
I'm still having doubts about the `ops` thing, as it's very verbose, and it essentially replicates what is already written in the unit files. Maybe there is value in such replication - like when a mistake is made in one place, the other copy is still fine. But the verbosity is awful.

Kristopher38 (approved):

PR #438, opened 2023-07-11
Add simultaneous transaction count warning

lekcyjna123:
Today my tests stuck in loop, and it looks like I did an error in using simultaneous transaction, which generated huge number of them (~30 000), so transactron haven't even started simulation. Here I add a warning so that in future it will be easier to find such problems.

Number of transactions from which warning appears was chosen empirically. By 3000 generating take already observable big time, but no such big that it will be killed by user pressing Ctrl-C.

### DIFF
--- a/coreblocks/transactions/core.py
+++ b/coreblocks/transactions/core.py
@@ -9,6 +9,7 @@ from amaranth import *
 from amaranth import tracer
 from itertools import count, chain, filterfalse, product
 from amaranth.hdl.dsl import FSM, _ModuleBuilderDomain
+from warnings import warn
 
 from coreblocks.utils import AssignType, assign, ModuleConnector
 from coreblocks.utils.utils import OneHotSwitchDynamic
@@ -358,6 +359,12 @@ class TransactionManager(Elaboratable):
                 continue
             q.extend(new_group | other_group for other_group in simultaneous if new_group & other_group)
             tr_simultaneous.add(new_group)
+
+        if len(tr_simultaneous) > 3000:
+            warn(
+                f"There is a lot of simultaneous transactions generated: {len(tr_simultaneous)}."
+                + "Elaborating can take hudge time. Are you sure that you didn't do any mistake?"
+            )
 
         # step 3: maximal group selection
         def maximal(group: frozenset[Transaction]):
--- a/coreblocks/transactions/core.py
+++ b/coreblocks/transactions/core.py
@@ -363,7 +363,7 @@ class TransactionManager(Elaboratable):
         if len(tr_simultaneous) > 3000:
             warn(
                 f"There is a lot of simultaneous transactions generated: {len(tr_simultaneous)}."
-                + "Elaborating can take hudge time. Are you sure that you didn't do any mistake?"
+                + "Elaborating can take huge time. Are you sure that you didn't do any mistake?"
             )
 
         # step 3: maximal group selection


tilk:
Could you point out which code was that slow? Was it the maximal group selection step, or something else? A comment should probably be written explaining the performance problem.

Also, the warning message you wrote is very poor English.

lekcyjna123:
time = 0 is on enter to TransactionManager elaborate function
For smaller transaction graph:
```
step 2 0.0009400844573974609
step 3 0.06304478645324707
step 4 0.1275768280029297
step 5 529 0.12812542915344238
end simultaneous 0.2046656608581543
method map created 0.2076737880706787
conflict graph created 1.149038553237915
methods enabled 1.3585214614868164
end elaborate 2.203390121459961
start of simulation 12.820
```

For bigger transaction graph:
```
step 2 0.0008661746978759766
step 3 1.2195680141448975
step 5 3393 6.187478542327881
step 4 6.186314821243286
step 5 3393 6.187478542327881
end simultaneous 6.860440254211426
method map created 6.889794588088989
[end of my patience]
```

tilk:
So it's the conflict graph generation which timed out for you? I guess you have a *lot* of relations. Overused priority `conditions` maybe?

I'm reluctant to accept this, as it looks like it's not the pure number of generated transactions which is the problem. You could get that number of transactions by using some simple design a lot of times, and it would not lead to the performance issue.

lekcyjna123:
> Overused priority conditions maybe?

I have checked for that in first place and I removed all priorities, but this didn't help. My hypothesis is that I overused conditions as whole:
```py
with Transaction():
  with condition() as branch_up:
    with branch_up():
       for _ in range(4):
         with condition() as branch:
            with branch(cond):
              method(m)
            with branch(~cond):
              pass
```
This is a pseudo-code of which I have written. `method` is implemented as condition over 8 branches. So what I get here are 4 transactions each of them is multiplied by 8 (because of `method`). These transactions are under upper condition, so they are connected every with every, so there is 8^4 = 4096 transactions.

tilk:
There is no (all transactions)^2 loop in conflict graph generation. The ones that are there are:

* (all methods)*(transactions calling a given method)^2
* (all relations)*(transactions calling start edge)*(transactions calling end edge)

The total number of transactions isn't a factor in any of those.

lekcyjna123:
But if i correctly remember when we have two conditions (first with branches A and B, and second with C and D) which are started simultaneously, they are glued together before conflict graph generation. So we have 4 transactions which describe every-to-every connections: AC, AD, BC, BD.

```py
with Transaction():
  with condition() as branch_up:
    with branch_up():
       for _ in range(4):
         with condition() as branch: # cond0, ..., cond3
            with branch(cond):
              method(m) # contains a condition over 8 branches (let them be branch_inner0, ..., branch_inner7)
            with branch(~cond):
              pass
```

So in my case, let investigate number of transactions where `branch_inner0` from inside `method` is called. So we can choose one of 4 `condX` to call this branch and 3 other conds are free to choose one of other branches. If we analyse only situation where different inner branches are chosen then we have:
4 * 7 * 6 * 5
And now in conflict_graph generation we have:
(all methods) * (transactions calling a given method)^2 = 8 * (4 * 7 * 6 * 5)^2  = 5,644,800 iterations.

lekcyjna123:
I have found a solution to deal with this problem. As always FIFO comes to rescue. All data which I send or receive from `method` first go through a FIFO. 

As a side note changing `condition` to `m.If` haven't solve the problem.

tilk:
> I printed (all methods) * (transactions calling a given method)^2 from my code, and this gave me: 133,742,500. There is 175 methods so by average there is 874 transactions generated for each method.

Again, you are missing the point. This number is big IN YOUR CASE. In other cases, a large amount of transactions doesn't lead to this number being large, as (in a large, highly parallel circuit) you might have a lot of methods and a lot of transactions, with the number of transactions calling each given method small. Therefore, your check will issue a warning on projects which synthesize just fine.

tilk:
And completely off-topic:

> I have found a solution to deal with this problem. As always FIFO comes to rescue. All data which I send or receive from method first go through a FIFO.

How much work did that require? I have a hunch that the Transactron approach should make changes like this fairly easy. It would be nice to have experience reports confirming this.

lekcyjna123:
> How much work did that require? I have a hunch that the Transactron approach should make changes like this fairly easy. It would be nice to have experience reports confirming this.

Not such much. Current implementation is a little big ugly and I should move it to separate module, but all needed changes fit in 20 lines. 
https://github.com/kuznia-rdzeni/coreblocks/blob/f6bd96867b7f77babb195f652b32c6bbd78a3be3/coreblocks/fu/vector_unit/v_elems_downloader.py#L43

>  Therefore, your check will issue a warning on projects which synthesize just fine.

Ok. So I am closing this PR.

PR #439, opened 2023-07-11
Added tip about `condition` to problem checklist.

lekcyjna123:
To allow other to learn on my errors here I add a tip about using `conditions`.

### DIFF
--- a/docs/Problem-checklist.md
+++ b/docs/Problem-checklist.md
@@ -12,4 +12,7 @@ If something doesn't work and you're puzzled as to why - go through this checkli
 
 5. Check if your code doesn't have any combinational loops - especially if your simulation hangs.
 
+6. Check that you don't use repeatedly any methods/transactions inside `condition` that use `condition` themselves. This leads to exponential complexity.
+
+
 Please extend this list if you spot yourself doing an easy-to-fix mistake.
--- a/docs/Problem-checklist.md
+++ b/docs/Problem-checklist.md
@@ -12,7 +12,7 @@ If something doesn't work and you're puzzled as to why - go through this checkli
 
 5. Check if your code doesn't have any combinational loops - especially if your simulation hangs.
 
-6. Check that you don't use repeatedly any methods/transactions inside `condition` that use `condition` themselves. This leads to exponential complexity.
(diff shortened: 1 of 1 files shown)


docs/Problem-checklist.md
```diff
@@ -12,4 +12,7 @@ If something doesn't work and you're puzzled as to why - go through this checkli
 
 5. Check if your code doesn't have any combinational loops - especially if your simulation hangs.
 
+6. Check that you don't use repeatedly any methods/transactions inside `condition` that use `condition` themselves. This leads to exponential complexity.
```
tilk, line 15: What does it mean to "use a transaction inside condition"?

lekcyjna123: Fixed. I am too used to the fact that methods and transactions have often very similar interface.

tilk:
Well, simultaneous transactions (and `condition`) are a tool which increases expressivity, *at a cost*. They allow to write some things in a simpler and more modular way, but you're not freed from understanding what you are writing. In particular, `condition` is not an `m.If` replacement. This should probably be written more clearly in the docs.

It's nice you are exploring the boundaries of this feature, though.

tilk (approved):

Kristopher38 (commented):
Hmm. Problem checklist was meant to contain simple solutions to problems when the code *doesn't work* (at least that's how I see it). How is the design having exponential complexity relevant here? I strongly feel like this should go to the docs instead (and be in a big red warning banner).

tilk:
> when the code doesn't work (...) How is the design having exponential complexity relevant here?

Because indeed, such code doesn't work, just elaborates endlessly. But yeah, I'm inclined to agree with you because:

* The solution is not super simple, some amount of redesign is needed, although the fix can be just adding another pipeline stage by using a FIFO.
* The `condition` construct is waiting for a serious use case (outside of @lekcyjna123's code), and we don't know how much of an issue this is in practice.
* This potential problem should be learned early when trying to use `condition` (or simultaneous transactions in general), so that one doesn't design something that can't really work.

So yeah, a sentence or two of warning in the docstring of `condition`, and then (later) some words on simultaneous transactions and `condition` in the transactions tutorial.

lekcyjna123:
So I am closing this PR, as not needed. If I have time, I will update condition docstring.

PR #440, opened 2023-07-11
E extension

### LINKED ISSUE #342: Implement RV32E (embedded variant)
The E variant is basically I with half the registers. The change involves, when E is enabled, reducing the physical register numbers by 1 bit, reducing the RATs, and rejecting instructions operating on non-existing registers in the decoder.

piotro888:
Implements RISC-V E extension. Closes #342.

### DIFF
--- a/coreblocks/frontend/decoder.py
+++ b/coreblocks/frontend/decoder.py
@@ -358,8 +358,7 @@ class InstrDecoder(Elaboratable):
 
         extensions = self.gen.isa.extensions
 
-        # We need to support all I instructions in E extension, but this is not a full implication.
-        # E&I coexisting is a different case, that should be handled in ISA class
+        # We need to support all I instructions in E extension, but this is not extension implication
         if Extension.E in extensions:
             extensions |= Extension.I
 
--- a/coreblocks/frontend/decoder.py
+++ b/coreblocks/frontend/decoder.py
@@ -402,10 +402,20 @@ class InstrDecoder(Elaboratable):
             self._extract(20, self.funct12),
         ]
 
+        rd_field = Signal(self.gen.isa.reg_field_bits)
+        rs1_field = Signal(self.gen.isa.reg_field_bits)
+        rs2_field = Signal(self.gen.isa.reg_field_bits)
+
         m.d.comb += [
-            self._extract(7, self.rd),
-            self._extract(15, self.rs1),
-            self._extract(20, self.rs2),
+            self._extract(7, rd_field),
+            self._extract(15, rs1_field),
+            self._extract(20, rs2_field),
+        ]
+
+        m.d.comb += [
+            self.rd.eq(rd_field),
+            self.rs1.eq(rs1_field),
+            self.rs2.eq(rs2_field),
         ]
 
         rd_invalid = Signal()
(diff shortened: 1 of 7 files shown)


lekcyjna123 (approved):

coreblocks/frontend/decoder.py
```diff
@@ -480,6 +485,13 @@ def elaborate(self, platform):
             self._extract(28, self.fm),
         ]
 
+        register_space_invalid = Signal()
+        m.d.comb += register_space_invalid.eq(
+            (self.rd_v & instr[7 + self.gen.isa.reg_cnt_log : 7 + self.gen.isa.reg_field_bits].any())
+            | (self.rs1_v & instr[15 + self.gen.isa.reg_cnt_log : 15 + self.gen.isa.reg_field_bits].any())
+            | (self.rs2_v & instr[20 + self.gen.isa.reg_cnt_log : 20 + self.gen.isa.reg_field_bits].any())
```
tilk, line 492: This looks ugly and error-prone with all that bit number arithmetic. But I'm not certain on how to do it best. One idea would to be to create local signals with `reg_field_bits` width, extract to them, assign `self.rd` etc. from them, and use them here for checking. This unfortunately adds more lines, but should have minimal bit number arithmetic. What do you think?

coreblocks/params/configurations.py
```diff
@@ -89,11 +89,13 @@ def replace(self, **kwargs):
...
 # Minimal core configuration
 tiny_core_config = CoreConfiguration(
+    embedded=True,
     func_units_config=(
         RSBlockComponent([ALUComponent(), ShiftUnitComponent(), JumpComponent()], rs_entries=2),
         LSUBlockComponent(),
     ),
-    rob_entries_bits=6,
+    phys_regs_bits=basic_core_config.phys_regs_bits // 2,
+    rob_entries_bits=basic_core_config.rob_entries_bits // 2,
```
tilk, line 98: Fail to understand the logic behind `// 2`. As it's written, this means that `tiny_core_config` gets only 8 physical registers, clearly not enough to back all 15 RVE registers. The division on `rob_entries_bits` is also suspicious. I understand that currently we have an oversized ROB for current needs, but dividing the bit width by 2 looks weird.

coreblocks/params/isa.py
```diff
@@ -237,10 +237,8 @@ class Extension(enum.IntFlag):
     G = I | M | A | F | D | ZICSR | ZIFENCEI
 
 
-# Extensions which are mutually exclusive
-_extension_exclusive = [
-    [Extension.I, Extension.E],
-]
+# Extensions which are mutually exclusive - not needed for now
+# _extension_exclusive = []
```
tilk, line 241: Will this be needed? If not, better to remove - the old code is in version control anyway.

coreblocks/params/configurations.py
piotro888: uhh I missed that those are bit counts, not entry count. Idea was to divide entry number by 2 because logical register count is divided by 2 too. Changed to `-1`

coreblocks/frontend/decoder.py
piotro888: Updated, avoids duplication and maybe looks nicer?

PR #441, opened 2023-07-11
DummyLSU: add FENCE

### LINKED ISSUE #311: Handle the FENCE instruction
The FENCE instruction is the only (?) thing missing from having full RV32I support. Our current trivial LSU makes the task easy: this instruction can basically be a no-op.

piotro888:
Finally, someone had to do it :)
Closes #311.
Implemented and tested no-op in dummyLsu.

As we now fully support the I extension, I thought of setting `allow_partial_extensions` in core configurations to `False`. But I extension specifies `ECALL` and `EBREAK`, so `ExceptionUnit` is necessary to comply with it. Do we want to put `ExceptionUnit` in `basic_core_config`? What about the tiny one?

### DIFF
--- a/coreblocks/lsu/dummyLsu.py
+++ b/coreblocks/lsu/dummyLsu.py
@@ -219,7 +219,7 @@ class LSUDummy(FuncBlock, Elaboratable):
 
         m.submodules.internal = internal = LSUDummyInternals(self.gen_params, self.bus, current_instr)
 
-        result_ready = internal.result_ready
+        result_ready = internal.result_ready | ((current_instr.exec_fn.op_type == OpType.FENCE) & current_instr.valid)
 
         @def_method(m, self.select, ~reserved)
         def _():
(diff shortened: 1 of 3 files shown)


lekcyjna123:
> Do we want to put ExceptionUnit in basic_core_config? 

I thing that yes, we want to put `ExceptionUnit` to `basic_core_config`. In base we should have all required base functionalities, and exceptions are part of them (e.g. to raise illegal instruction). In tiny core I think that we can skip exception unit.

lekcyjna123 (approved):

piotro888:
Changed. But I'm still not sure about tiny core (microcontroller probably should handle interrupts, so why not excpetions). I think that, if we want exceptions disabled, better way might be to add this FU (to not stall scheduler) and disable exception bit checking, so the infrastructure would optimize away. In that case ECALL and EBREAK would be `I` extenstoin compliant, because they are specified only to do something, ignoring them is also a valid action (raising ILLEGALINSTR is even better action, but that needs exception support again to handle).

Anyway we need to rethink this all core configurations in future, it can be adjusted then (but alternative exception disable in core could be introduced earlier)

tilk (approved):

PR #459, opened 2023-08-13
Set seed for nextpnr

lekcyjna123:
I have found that nextpnr has an option to set a seed for random number generator. I hope that this will do our synthesis results more stable.

### DIFF
--- a/scripts/synthesize.py
+++ b/scripts/synthesize.py
@@ -142,7 +142,7 @@ def synthesize(core_config: CoreConfiguration, platform: str, core: UnitCore):
     resource_builder, module = core(gen_params)
 
     if platform == "ecp5":
-        make_ecp5_platform(resource_builder)().build(module)
+        make_ecp5_platform(resource_builder)().build(module, nextpnr_opts="--seed 0")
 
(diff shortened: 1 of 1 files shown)


lekcyjna123:
Job to run synthesis with the new option:
https://github.com/kuznia-rdzeni/coreblocks/actions/runs/5846546722

piotro888:
Unfortunately it seems that nextpnr runs with fixed seed by default, until --randomize-seed, or --seed option is passed, and none of those was defined in our pipeline call.
See: https://github.com/YosysHQ/nextpnr/blob/679b662a2b870ef49779fd0c3698e6e9a356d9f0/common/kernel/command.cc#L264C5-L264C5 , https://github.com/YosysHQ/nextpnr/blob/679b662a2b870ef49779fd0c3698e6e9a356d9f0/common/kernel/deterministic_rng.h#L37C5-L37C21 , https://github.com/search?q=repo%3AYosysHQ%2Fnextpnr%20rngseed&type=code

lekcyjna123:
Yes, you have right.

Issue #446, opened 2023-07-14
Project cleanup


tilk:
There is a need to move stuff around to make things more intuitive for newcomers (and others). Some problems in directory organization are:

* The `coreblocks/params` directory is a mess. It was supposed to contain things related to parametrization (at the beginning, mostly `GenParams`). Now it also contains:
  * `dependencies` - the dependency manager,
  * `keys` - the dependency keys (which we agreed to have in one place),
  * `isa` and `instr` - RISC-V architecture related things,
  * `optypes` - our internal encoding of instructions, which is entangled with ISA, decoding and FUs.
* The `structs_common` directory somehow contains a CSR unit. The directory was meant to contain implementation of data structures, i.e. things that are shared between components.
* The `coreblocks/transactions/lib` is one big ball of stuff, it should finally be split.
* The `stages` directory is weird, it contains:
  * The announcement module, in a file which is somehow named `backend`,
  * The retirement stage,
  * Abstractions related to functional units/blocks.

There are potentially other issues I can't remember right now, please write them down in comments.

lekcyjna123:
Another issues:
- move whole `coreblocks/transactions` to `transactron` directory so that it will be easier to split it in future to separate repository
- cleanup `utils`
  - fifos use transactional interface so they should be part of `transactions/lib`
  - some modules can be send to amaranth upstream
  - `_typing` is a ball of stuff for types, on the begining it was ment to consist only internal types, but now it has types representing amaranth concepts, which we use in `transactron` interfaces
- `transactions/core` also is huge, it should be splitted (e.g. to definitions used by developers such as `Method`, `Transaction` and transactron internal stuff such as `methodMap`)
- I would suggest to split syntax sugar to separate files so that it will be more clear that it isn't strictly needed to use it
- `test/common.py` need splitting
  - useful functions should go to one place
  - classes used on test case creation (SimpleTestCircuit, TestCaseWithSimulator)
  - testbenchIO
  - syntax sugar
- `ci` and `scripts` directories should be in my opinion merged
- I would suggest splitting `coreblocks/scheduler/scheduler.py` to smaller files
- `_instructions_by_optype` should be moved from `coreblocks/frontend/decoder.py` to separate file (after adding part of vector extension it take more than 270 lines to define all instructions)

lekcyjna123:
[...]

PR #460, opened 2023-09-27
Autumn cleaning part 1 - move transactron

lekcyjna123:
Hi,

The new semester is coming so it is good time to start cleaning our repository. This is the first commit to implement #446. I have moved transactron from coreblocks/transactions to transactron directory to allow in future for the easier separation of transactron to its own project/repository. This move required to update all transactron imports in coreblocks and tests directories.

When this MR will be accepted my next step will by to split transactron/lib.py to smaller parts.


Kristopher38:
If you're already moving this out of the `coreblocks` directory tree, why not go all the way through and split it into a separate submodule?

lekcyjna123:
If you create separate submodule, then you have to crate separate review for transactron and its lib and for coreblocks. Whats more it will be hard to manage dependencies between them. So as long as transactron is developed rapidly I would prefer to leave it in coreblocks.

lekcyjna123:
> There are still dependencies from Transactron to Coreblocks.

Yes this is expected. I wanted to use small steps principle to allow for easy review of code. Here I simply drag and drop transactron code, only with changes which are required to don't have regression. Whole cleaning will be in next PRs.

tilk (approved):

Kristopher38 (approved):
> If you create separate submodule, then you have to crate separate review for transactron and its lib and for coreblocks.

Well this is kind of expected when you work with submodules but I'm not so sure it hinders the work that much.

>Whats more it will be hard to manage dependencies between them.

The point is to have transactron be its own separate lib not dependent on coreblocks though (unless you meant something else).

>So as long as transactron is developed rapidly I would prefer to leave it in coreblocks.

Is it though? @tilk what do you think?

tilk:
> Is it though? @tilk what do you think?

Ultimately, I want it to be a separate lib outside of Coreblocks. I have nothing against a small steps approach: let's first prepare for this separation, and later separate using one simple commit.

PR #461, opened 2023-09-27
Fix pipeline

lekcyjna123:
It looks like we use too old amaranth and our pipeline fails to install it:
https://github.com/kuznia-rdzeni/coreblocks/actions/runs/6323098696/job/17169995647?pr=460

The solution is to update amaranth, but this cause some other issues, which we have to debug.

### DIFF
--- a/coreblocks/params/instr.py
+++ b/coreblocks/params/instr.py
@@ -28,6 +28,9 @@ class RISCVInstr(ABC, ValueCastable):
     def as_value(self):
         return self.pack()
 
+    def shape(self):
+        return self.as_value().shape()
+
 
 class RTypeInstr(RISCVInstr):
     def __init__(
--- a/coreblocks/structs_common/exception.py
+++ b/coreblocks/structs_common/exception.py
@@ -68,11 +68,9 @@ class ExceptionCauseRegister(Elaboratable):
                 m.d.comb += should_write.eq(should_update_prioriy(m, current_cause=self.cause, new_cause=cause))
             with m.Elif(self.valid):
                 rob_start_idx = self.rob_get_indices(m).start
-                diff1 = Signal.like(rob_start_idx)
-                diff2 = Signal.like(rob_start_idx)
-                m.d.comb += diff1.eq((rob_id - rob_start_idx))
-                m.d.comb += diff2.eq((self.rob_id - rob_start_idx))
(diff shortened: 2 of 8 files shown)


Kristopher38:
Have you tried to debug why this actually happened? If so, what have you found out?

btw we're not ready to update amaranth yet, as fixes for https://github.com/amaranth-lang/amaranth/issues/790 are not there yet (unless we bodge our own workaround).

lekcyjna123:
I haven't had time to debug this yet. I will try to do this tomorrow.

> btw we're not ready to update amaranth yet, as fixes for https://github.com/amaranth-lang/amaranth/issues/790 are not there yet (unless we bodge our own workaround).

Hmm. So maybe I will look for a other way to fix this pipeline issue. Maybe downgrade of pip will help.

tilk:
> btw we're not ready to update amaranth yet, as fixes for https://github.com/amaranth-lang/amaranth/issues/790 are not there yet (unless we bodge our own workaround).

I think we can upgrade just fine, we just keep using Records as before. The Amaranth update must be done with care though - implemented RFCs must be read and type stubs updated.

tilk:
Tried to do required changes, but somehow some tests broke. This requires investigation. See https://github.com/kuznia-rdzeni/coreblocks/pull/462

lekcyjna123:
It is interesting that regression using cocotb passed, only unit tests failed. List of failing modules:
- ICache
- CSR
- Exception
- MemoryBank
- Core (both asm and random)

Particularly this last module is fully tested by cocotb, so I would rather say that there can be something in tests.

lekcyjna123:
In exception unit there was a problem, that after update `rob_id - rob_start_idx` had signed(8) shape, what caused wrong comparison results in some cases.

lekcyjna123:
I have created a minimal example which points to problems of Amaranth with simulation memory. This can explain why ICache, MemoryBank and Core fails. Issue has been reported here: https://github.com/amaranth-lang/amaranth/issues/922

tilk:
Found another change in Amaranth's behavior, I believe it to be a bug and reported it: https://github.com/amaranth-lang/amaranth/issues/926

tilk:
I took the liberty to clean up this PR for merging, as it's super-extra-blocking.

piotro888 (approved):

PR #463, opened 2023-10-05
Better debug signals for simple test circuit

tilk:
Changed debug signals for `SimpleTestCircuit` to expose the internal module and show the names of exposed IOs.

Also, cleaned up one ugly line of code.


Kristopher38 (approved):
lgtm

PR #464, opened 2023-10-05
Autumn cleaning part 2 - split transactron/lib.py

lekcyjna123:
As a next step I have chosen to split transactron/lib.py into smaller files. There is no changes, other than required to pass tests, to keep this merge request small. Please feel free to comment about directory names.


Kristopher38:
Bikeshedding time - I'd rename:
`connections` -> `connectors` (noun for more of a physical rather than abstract thing)
`methods` -> `transformers` (those classes share a theme of transforming method inputs/outputs in some way)

Also with this scheme it feels like stuff in `transactions` should land in `connectors` and `transformers`. I think my point of view boils down to a notion of "what something does" (vs. "what something is" is how you classified it). I'm not sure my classification is better or worse, so I'm not dead set on these changes.

It also feels like the overall directory structure is reversed? Internal stuff like `core.py`, `graph.py` and `tracing.py` is in the main `transactron` directory, and the user-facing stuff is in `lib` subdirectory. Shouldn't this be the other way around?

tilk:
I very like @Kristopher38's suggestions.

lekcyjna123:
I pushed patches with @Kristopher38 ideas. Please say if they are according to your expectations.

> It also feels like the overall directory structure is reversed? Internal stuff like core.py, graph.py and tracing.py is in the main transactron directory, and the user-facing stuff is in lib subdirectory. Shouldn't this be the other way around?

This is also a great idea, but let leave it for separate PR. `core.py` also should be split as it has over 1200 lines. After that we can move rest of files to directory `internal`, so the directory tree would look like:
```
transactron/
+ core/
|   [...]
+ lib/
|   [...]
- internal/
    [...]
```

Kristopher38 (approved):
lgtm

> but let leave it for separate PR
alright :)

lekcyjna123:
In orginal code there was a bug in docstrings and API for `ArgumentsToResultsZipper` wasn't generated (see current docs). After move it was detected by sphinx and an error was raised. The newest commit modifed docstring to solve the issue.

tilk (approved):

PR #468, opened 2023-10-06
Fix broken synthesis after cleanup

### LINKED ISSUE #465: Broken synthesis benchmark after cleanup
Merging #460 broke synthesis benchmark workflow.

lekcyjna123:
This fix #465. There was no type analysis activated for `constants` directory, so the error wasn't detected by pyright. Additionally there was one more type error, which is fixed by this MR.

Synthesis job: https://github.com/kuznia-rdzeni/coreblocks/actions/runs/6433713558

### DIFF
--- a/constants/ecp5_platforms.py
+++ b/constants/ecp5_platforms.py
@@ -60,6 +60,7 @@ class PinManager:
             if name in self.pin_bag:
                 self.pin_bag.remove(name)
                 return name
+        raise RuntimeError("Named pins %s not free" % ", ".join(names))
 
 
 ResourceBuilder: TypeAlias = Callable[[PinManager], list[Resource]]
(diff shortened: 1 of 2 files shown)


piotro888 (approved):

constants/ecp5_platforms.py
```diff
@@ -123,9 +123,12 @@ class ECP5BG756Platform(LatticeECP5Platform):
         default_clk = "clk"
         default_rst = "rst"
 
+        clk_pin = pins.named_pin(ecp5_bg756_pclk)
+        if clk_pin is None:
```
Kristopher38, line 127: this is required only to satisfy the typechecker, right?

Kristopher38 (approved):
lgtm

constants/ecp5_platforms.py
tilk: I would much rather throw an exception in the `PinManager`, and keep this code clean.

tilk (approved):

PR #470, opened 2023-10-07
Fix zbs and DecoderManager docstrings

lekcyjna123:
Sphinx raised a warning about zbs unit docstrings and the docs were generated wrongly. Additionally docstrings for `DecoderManager` were miss placed, so they weren't included in API output.

This commit fix both issues.


coreblocks/fu/fu_decoder.py
```diff
@@ -55,62 +55,58 @@ class DecoderManager:
...
-    return : Sequence[tuple]
-        List of implemented instructions, each following format:
-        (IntFlag, OpType, Funct3 (optional), Funct7 (optional))
-
-    """
-
     def get_instructions(self) -> Sequence[tuple]:
-        raise NotImplementedError
+        """
+        Method providing list of valid instruction.
```
tilk, line 60: Don't we usually start the first line immediately after the triple quote?

tilk (approved):

coreblocks/fu/fu_decoder.py
```diff
@@ -55,62 +53,54 @@ class DecoderManager:
     """
     Fn: Type[IntFlag]
```
JumiDeluxe, line 54: Sphinx doesn't generate documentation for this variable when it's comment block is placed above it.

Move it below. Then it generates documentation from the comment.

```suggestion
    """
    Type[IntFlag]

    Enumeration of instructions implemented in given functional unit.
    """
    Fn: Type[IntFlag]
```

JumiDeluxe: I think that `Type[IntFlag]` in comment blocks is not even needed since we have type hint.
Just 
```python
Fn: Type[IntFlag]
    """Enumeration of instructions implemented in given functional unit."""
```
should be enough (works for me).

JumiDeluxe (changes requested):
Builds for me.
Just missing `Fn: Type[IntFlag]` in generated Sphinx documentation, with suggested changes everything is OK

JumiDeluxe (approved):

PR #471, opened 2023-10-07
Autumn cleaning part 3 - split transactron/core.py

lekcyjna123:
Here is proposition to split core.py into smaller parts. Tests don't work yet (particularly I haven't started to solve circular imports yet).  Please comment about split and naming.

Method, Transaction, TransactionBase, TransactionManager are very big classes, so I decided that each of them will be in separate file.


lekcyjna123:
There were needed some changes in transactron code to break cyclic references. Some of them are imports inside functions instead of top level imports. Affected functions:
- MethodMap.transactions_for
- Transaction.__init__
- TransactionBase.context

The alternative is to use absolute names, which will be ugly in my opinion.

transactron/core/transaction.py
```diff
@@ -0,0 +1,111 @@
...
+from .._utils import get_caller_class_name
+
+__all__ = [
+    "Transaction",
+]
+if TYPE_CHECKING:
+    from .manager import TransactionManager
+
+
+class Transaction(transaction_base.TransactionBase):
```
tilk, line 17: This is super-fugly. Can this be avoided? If not, maybe just get `TransactionBase`, `Transaction` and `Method` into a single file?

transactron/core/transaction.py
```diff
@@ -0,0 +1,111 @@
...
+        name: str or None
+            Name hint for this `Transaction`. If `None` (default) the name is
+            inferred from the variable name this `Transaction` is assigned to.
+            If the `Transaction` was not assigned, the name is inferred from
+            the class name where the `Transaction` was constructed.
+        manager: TransactionManager
+            The `TransactionManager` controlling this `Transaction`.
+            If omitted, the manager is received from `TransactionContext`.
+        """
+        from .manager import TransactionContext
```
tilk, line 66: Won't separating the `TransactionContext` from the `manager` file solve this?

transactron/core/transaction_base.py
```diff
@@ -0,0 +1,151 @@
...
+            The `Transaction`\\s or `Method`\\s which, together with this
+            `Transaction` or `Method`, need to be independently considered
+            for execution.
+        """
+        self.independent_list += others
+
+    @contextmanager
+    def context(self, m: "TModule") -> Iterator[Self]:
+        from .method import Method
+        from .transaction import Transaction
```
tilk, line 116: This will be resolved by putting `Method`, `Transaction` and `TransactionBase` in a single file.

transactron/core/relation_database.py
```diff
@@ -0,0 +1,73 @@
...
+        for transaction in transactions:
+            self.methods_by_transaction[transaction] = []
+            rec(transaction, transaction)
+
+    def transactions_for(self, elem: TransactionOrMethod) -> Iterable["Transaction"]:
+        # Here is the only place, where real definition of Transaction is needed. So
+        # we import this class here to break cyclic imports
+        from .transaction import Transaction
+
+        if isinstance(elem, Transaction):
```
tilk, line 58: Don't like this, too.

tilk (commented):
I feel this is too fine-grained, and I very much don't like how cyclic dependencies are handled. This might complicate altering the Transactron core in the future.

transactron/core/transaction_base.py
lekcyjna123: I don't like the idea to put `Method`, `Transaction` and `TransactionBase` to single file. Such file will be big (over 500 lines).

transactron/core/transaction.py
lekcyjna123: Fixed. This was leftover from one of first fix attempts.

transactron/core/transaction_base.py
lekcyjna123: I start to think if our type hints are correct. In `TransactionBase` we make some assumptions about subclasses. We allow only `Transaction` and `Method` subclass, so we are leaking the abstraction. The type hint `TransactionOrMethod` is from the time before creating `TransactionBase`.
https://github.com/kuznia-rdzeni/coreblocks/commit/c87bb3112a437e47afae65752e7efcd9035054d9
So in my opinion we should use `TransactionBase` type instead of `TransactionOrMethod`.

transactron/core/relation_database.py
lekcyjna123: Removed

transactron/core/transaction.py
lekcyjna123: Separated

transactron/core/schedulers.py
```diff
@@ -0,0 +1,88 @@
...
+from .._utils import Graph, GraphCC, Scheduler
+from .method import Method
+from .transaction import Transaction
+from .relation_database import MethodMap
+from .typing import TransactionGraph, TransactionGraphCC, PriorityOrder
+
+__all__ = [
+    "eager_deterministic_cc_scheduler",
+    "trivial_roundrobin_cc_scheduler",
+        ]
```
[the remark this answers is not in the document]
lekcyjna123: Removed

transactron/core/relation_database.py
```diff
@@ -0,0 +1,69 @@
...
+from .method import Method
+from .transaction import Transaction
+from .transaction_base import TransactionBase
+from .typing import TransactionOrMethod
+
+__all__ = [
+        'Priority'
+        ]
+
+
```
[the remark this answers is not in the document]
lekcyjna123: Removed

lekcyjna123:
Second proposition has been just uploaded. `Method`, `Transaction` and `TransactionBase` still are in separate files. But imports from insides functions have been removed. This removal was possible because of:
- Change of type hints from `TransactionOrMethod` to `TransactionBase` (in `MethodMap`, `TransactionBase`, `Relation` and `RelationBase`)
- Moving `TransactionContext` to `creators`

Changes:
- `modules` have been renamed to `creators`
- `sugar` has been removed
- `relation_database` has been removed

transactron/core/transaction_base.py
tilk: > So in my opinion we should use TransactionBase type instead of TransactionOrMethod.

No, this is wrong. `TransactionBase` is an open type, `TransactionOrMethod` is a closed one. Or: if you know that a `TransactionOrMethod` is not a `Transaction`, it must be a `Method`. There is no such guarantee for `TransactionBase`.

lekcyjna123: Yes. There is no such guarantee on type level, but if we need it, then the code is wrongly designed. In base class these should be defined only code, which is common for all its subclasses. So limiting it to only `TransactionOrMethod` suggest, that it wouldn't be common for other subclasses, so it shouldn't be in base class.

tilk:
I think that this looks much better than the initial proposal. Would like to get a second opinion before deciding.

transactron/core/transaction_base.py
tilk: You seem to misunderstand. The transaction manager works on transactions and methods *only*, and the `TransactionOrMethod` type tells the typechecker exactly that. Your change messed this up, see e.g. lines 52-55 in `manager.py`.

transactron/core/manager.py
```diff
@@ -0,0 +1,412 @@
...
+
+        for transaction in transactions:
+            self.methods_by_transaction[transaction] = []
+            rec(transaction, transaction)
+
+    def transactions_for(self, elem: "TransactionBase") -> Iterable["Transaction"]:
+        if isinstance(elem, Transaction):
+            return [elem]
+        else:
+            assert isinstance(elem, Method)
```
tilk, line 55: This is silly. Original type already told the typechecker that this is the only option here.

tilk:
The big problem with this PR is that I see NO WAY to check what actual changes were made to the code, so there is no way to review them.

transactron/core/manager.py
```diff
@@ -0,0 +1,412 @@
...
+        args = defaultdict[Method, list[ValueLike]](list)
+        runs = defaultdict[Method, list[ValueLike]](list)
+
+        for source in method_map.methods_and_transactions:
+            if isinstance(source, Method):
+                run_val = Cat(transaction.grant for transaction in method_map.transactions_by_method[source]).any()
+                run = Signal()
+                m.d.comb += run.eq(run_val)
+            else:
+                assert isinstance(source, Transaction)
```
tilk, line 202: Again, you silently added this `assert`.

lekcyjna123:
> The big problem with this PR is that I see NO WAY to check what actual changes were made to the code, so there is no way to review them.

Yes, it is the problem and the reason why I wanted to have all split to small PR. I have two ideas:
- I can create separate PR in which I will introduce all needed modification without moving files
- You can compare first commit in this review with the newest. I did moves in the first, so all other contain changes which were needed to make the code works.

tilk:
> You can compare first commit in this review with the newest. I did moves in the first, so all other contain changes which were needed to make the code works.

Unfortunately, some of your later commits contain both moves and code changes, e.g. "Remove relation database". As you smuggled weakening of the type guarantees in the most fundamental code in this project there, I'm now becoming paranoid with regard to your changes.

lekcyjna123:
Ok. So I will try to split the changes to separate PR without any moves.

Issue #384, opened 2023-06-07
Python 3.11 issues with VCD dumps


Kristopher38:
In python 3.11 some signals don't appear in VCD output (one clue: perhaps this happens when they're not a class member) or have width truncated to 1 (!). Needs to be investigated.

lekcyjna123:
Not sure if this is correlated but I have observed that giving a name explicit cause that signals appears normally in VCD.

lekcyjna123:
Reproduction:
commit: `07f0907f84c2715aa5f2f13a3aaadb8be24c66d5` on branch: `lekcyjna/vector-permutator-2`
Command: `./scripts/run_tests.py -t test.structs_common.test_scoreboard.TestScoreboard.test_random_1`

https://github.com/kuznia-rdzeni/coreblocks/blob/07f0907f84c2715aa5f2f13a3aaadb8be24c66d5/coreblocks/structs_common/scoreboard.py#L50-L51

Is signals have a name set as an argument, then both appears in VCD in `dut`. And their first non zero values are `0x11A` for `data_forward_valid` and `0x112` for `data_forward`. If `name` arguments from this signal are commented out:
https://github.com/kuznia-rdzeni/coreblocks/blob/8b3536e491c9fbb57c572b53db3a9084bb4259c1/coreblocks/structs_common/scoreboard.py#L50-L51
then in VCD in the same place are signals `data` and `data_forward` (no `data_forward_valid` and `data` was not present when these two signals had its own names). Values in first non zero cycle: `data_forward`: `0x11A`, `data` : `0x112`.

This suggest in my opinion that there are problems with automatic signal name recognitions in amaranth on python 3.11.  Real `data` signal from scoreboard is set synchronously, so it can not have the same value in the same cycle as `data_forward` signal which is set combinational.

tilk:
A minimized example (a short `Elaboratable` without dependencies on Coreblocks code) would allow to report the issue to Amaranth.

(By the way, with the recent release of PyPy with 3.10 support, I tried running Coreblocks tests using PyPy. They ran significantly longer. This probably means that we should stay on CPython and go ahead with the 3.11 upgrade soon.)

Kristopher38:
[...]

PR #472, opened 2023-10-08
Change TransactionBase into a Protocol

lekcyjna123:
Here is a proposition of solution to problem found in #471. In `TransactionBase` we try to enforce closed set of types (`TransactionOrMethod`) in place where python provide us an open set (`TransactionBase`). This open set is a result of inheritance, where there is an assumption, that base class implementation should be common for all derived classes. We workarounded that by asserting if argument is a `Method` or a `Transaction`. But this workaround caused cyclic dependencies in imports after split of core.py.

Proposed solution use `Protocol` instead of inheritance to fully move type checking into static analysis. This change allow us to add type hint "TransactionOrMethod` to `self` in `context` function, which is not possible in plain base class (where type of `self` have to be superclass of its class type).

### DIFF
--- a/transactron/core.py
+++ b/transactron/core.py
@@ -2,7 +2,18 @@ from collections import defaultdict, deque
 from collections.abc import Sequence, Iterable, Callable, Mapping, Iterator
 from contextlib import contextmanager
 from enum import Enum, auto
-from typing import ClassVar, NoReturn, TypeAlias, TypedDict, Union, Optional, Tuple
+from typing import (
+    ClassVar,
+    NoReturn,
+    TypeAlias,
+    TypedDict,
+    Union,
+    Optional,
+    Tuple,
+    TypeVar,
+    Protocol,
+    runtime_checkable,
+)
 from graphlib import TopologicalSorter
 from typing_extensions import Self
 from amaranth import *
@@ -38,6 +49,7 @@ PriorityOrder: TypeAlias = dict["Transaction", int]
 TransactionScheduler: TypeAlias = Callable[["MethodMap", TransactionGraph, TransactionGraphCC, PriorityOrder], Module]
 RecordDict: TypeAlias = ValueLike | Mapping[str, "RecordDict"]
 TransactionOrMethod: TypeAlias = Union["Transaction", "Method"]
+TransactionOrMethodBound = TypeVar("TransactionOrMethodBound", bound=Union["Transaction", "Method"])
 
 
 class Priority(Enum):
@@ -670,15 +682,20 @@ class TModule(ModuleLike, Elaboratable):
         return self.main_module
 
 
-class TransactionBase(Owned):
+@runtime_checkable
+class TransactionBase(Owned, Protocol):
     stack: ClassVar[list[Union["Transaction", "Method"]]] = []
     def_counter: ClassVar[count] = count()
     def_order: int
     defined: bool = False
     name: str
+    method_uses: dict["Method", Tuple[ValueLike, ValueLike]]
+    relations: list[RelationBase]
+    simultaneous_list: list[TransactionOrMethod]
+    independent_list: list[TransactionOrMethod]
 
     def __init__(self):
-        self.method_uses: dict[Method, Tuple[ValueLike, ValueLike]] = dict()
+        self.method_uses: dict["Method", Tuple[ValueLike, ValueLike]] = dict()
         self.relations: list[RelationBase] = []
         self.simultaneous_list: list[TransactionOrMethod] = []
         self.independent_list: list[TransactionOrMethod] = []
@@ -769,9 +786,7 @@ class TransactionBase(Owned):
         self.independent_list += others
 
     @contextmanager
-    def context(self, m: TModule) -> Iterator[Self]:
-        assert isinstance(self, Transaction) or isinstance(self, Method)  # for typing
-
+    def context(self: TransactionOrMethodBound, m: TModule) -> Iterator[TransactionOrMethodBound]:
         parent = TransactionBase.peek()
         if parent is not None:
             parent.schedule_before(self)
--- a/transactron/graph.py
+++ b/transactron/graph.py
@@ -3,15 +3,14 @@ Utilities for extracting dependency graphs from Amaranth designs.
 """
 
 from enum import IntFlag
-from abc import ABC
 from collections import defaultdict
-from typing import Literal, Optional
+from typing import Literal, Optional, Protocol
 
 from amaranth.hdl.ir import Elaboratable, Fragment
 from .tracing import TracingFragment
 
 
-class Owned(ABC):
+class Owned(Protocol):
     name: str
     owner: Optional[Elaboratable]
 


lekcyjna123:
It looks like there is some incompatibility between CI and my local setup. Locally all tests passed. I suppose that this is because of the python 3.11 which I use locally.

lekcyjna123:
From [What's new in python 3.11](https://docs.python.org/3/whatsnew/3.11.html#what-s-new-in-python-3-11):
> The [`__init__()`](https://docs.python.org/3/reference/datamodel.html#object.__init__) method of [Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) subclasses is now preserved. (Contributed by Adrian Garcia Badarasco in [gh-88970](https://github.com/python/cpython/issues/88970).)

This probably explain, why tests on 3.10 fails.

tilk:
This looks like a great reason to move to 3.11. Please do so in a separate PR though.

lekcyjna123:
Yes. Update in this PR is for testing purposes. When all check pass, I will move them to separate MR. I am building now docker images with updated python. Do I need any additional permissions to upload docker images to ghcr.io or should it work without them?

tilk:
We don't have a workflow which builds Docker images. And some permissions will probably be needed, but I don't really remember how this worked.

lekcyjna123:
I have build images, but I don't have access to push them into `kuznia-rdzeni` namespaces. To my private namespace I am able to push.
obraz

robtaylor:
Should just be a matter of declaring the right perms in the work flow file.

Happy to help if needed.

tilk:
@lekcyjna123: You got write permissions for the packages, please don't mess up stuff.

transactron/core.py
```diff
@@ -769,9 +786,7 @@ def _independent(self, *others: TransactionOrMethod) -> None:
         self.independent_list += others
 
     @contextmanager
-    def context(self, m: TModule) -> Iterator[Self]:
-        assert isinstance(self, Transaction) or isinstance(self, Method)  # for typing
-
+    def context(self: TransactionOrMethodBound, m: TModule) -> Iterator[TransactionOrMethodBound]:
```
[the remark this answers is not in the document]
lekcyjna123: Trick is loosely inspired by [PEP 673](https://peps.python.org/pep-0673/). The main difference is that in PEP 673 type used in `self` is bound to the base class to fullfill requirement that self has to be superclass of its type. Using `Protocol` allow us to drop this requirement and this was my idea to use that construct. Base class in inheritance is open type by default, so we needed something else. Protocols as specified in [PEP 544](https://peps.python.org/pep-0544/) are designed to:
> [...] allowing users to write the above code without explicit base classes in the class definition

And non explicit base class was exactly what we needed. Additionally the same PEP specify that protocol can have [default implementation](https://peps.python.org/pep-0544/#protocol-members). So it can be used very similar to the base class in inheritance. There is also a point about trick with `Self` to be used with [protocols](https://peps.python.org/pep-0544/#self-types-in-protocols).

tilk:
Why did you create a new Docker package?

lekcyjna123:
[amaranth-synth-ecp5](https://github.com/orgs/kuznia-rdzeni/packages/container/package/amaranth-synth-ecp5) is renamed `amaranth-synth` this is because of the tag `ecp5` used in the original one. I haven't seen possibility to update it in consistent way. But as I think about it now, then maybe instead of changing package base name I should change tag to `ecp5-2023.10.09`.

riscv-toolchain I uploaded into personal space, because I haven't had permissions to upload into `kuznia-rdzeni` space.

lekcyjna123:
In the newest version I had to use flag `--break-system-packages` in workflows, which don't use python setup. So we should analyse in future how to remove that flag and properly use caches in the rest of our containers.

I see that all tests passed. Tomorrow I will create new PR with python update.

tilk:
> In the newest version I had to use flag `--break-system-packages` in workflows,

Looks smelly. Why did you use it? (You seem quick to hack things, but you rarely explain why.)

Kristopher38:
>So we should analyse in future how to remove that flag

```
python -m venv venv
source venv/bin/activate
```
:)

(This creates a virtual environment in directory `venv` and "activates" it, after that you can install the packages as if they were being installed system-wide, but they get installed into the `venv` directory instead, which you must `source venv/bin/activate` on any new terminal session if you want to use the packages installed there)

Also if you ever want to run python version not available on your distro but don't want to compile from source I recommend [miniconda](https://docs.conda.io/projects/miniconda/en/latest/)

Kristopher38:
> This looks like a great reason to move to 3.11. Please do so in a separate PR though.

Afaik there was a bug on python 3.11 where signal width would get trimmed to 1 bit in vcd dumps. I was supposed to find a minimal example and submit an issue but forgot, not sure if that bug is still present though.

tilk:
> Afaik there was a bug on python 3.11 where signal width would get trimmed to 1 bit in vcd dumps. I was supposed to find a minimal example and submit an issue but forgot, not sure if that bug is still present though.

If it is, it's probably time to have it fixed.

Kristopher38:
> > Afaik there was a bug on python 3.11 where signal width would get trimmed to 1 bit in vcd dumps. I was supposed to find a minimal example and submit an issue but forgot, not sure if that bug is still present though.
> 
> If it is, it's probably time to have it fixed.

This was fixed 5 months ago and we updated amaranth to a version which includes that fix, so it's no longer an issue. Full details [here](https://github.com/kuznia-rdzeni/coreblocks/issues/384#issuecomment-1756057153)

transactron/core.py
```diff
@@ -670,15 +682,20 @@ def elaborate(self, platform):
         return self.main_module
 
 
-class TransactionBase(Owned):
+@runtime_checkable
```
tilk, line 685: Does it really need to be `runtime_checkable`?

tilk: You wanted to point to this line:

https://github.com/kuznia-rdzeni/coreblocks/pull/472/files#diff-c921823c823cc8629523d78c2d028fd8642e477cbb7f8cd5825c4c3f369c9d8fR790

This could be special-cased, but I agree that using `runtime_checkable` is probably cleaner.

tilk (approved):

transactron/core.py
```diff
@@ -670,15 +682,20 @@ def elaborate(self, platform):
...
+class TransactionBase(Owned, Protocol):
     stack: ClassVar[list[Union["Transaction", "Method"]]] = []
     def_counter: ClassVar[count] = count()
     def_order: int
     defined: bool = False
```
Kristopher38, line 690: This was not in scope of this PR but made me curious - this definitely looks like a class variable, but is used elsewhere as if it was an instance variable. I think this should probably be assigned in `__init__`? Or am I missing something here?

transactron/core.py
```diff
@@ -670,15 +682,20 @@ def elaborate(self, platform):
...
     name: str
+    method_uses: dict["Method", Tuple[ValueLike, ValueLike]]
+    relations: list[RelationBase]
+    simultaneous_list: list[TransactionOrMethod]
+    independent_list: list[TransactionOrMethod]
```
Kristopher38, line 695: Do these type definitions have to be both here and in `__init__`?

Kristopher38 (commented):
Looks sensible from my perspective, though I'm not an expert on python typing.

transactron/core.py
lekcyjna123: In Protocols and ABC instance variables are declared as class variables. See https://peps.python.org/pep-0544/#protocol-members

transactron/core.py
lekcyjna123: Yes. Here to define the interfaces, so the fact that we have such instance members and in `__init__` to initialise each of them with new empty list.

Kristopher38 (approved):

PR #473, opened 2023-10-14
Update python to 3.11

lekcyjna123:
As pointed in #472 we need to update python to 3.11 to be allowed to inherit `__init__` in `Protocol`. This pull request update docker files, docker images and documentation.

It introduced venv to jobs which uses custom dockers because in python 3.11 installing packages system wide is disabled by default and `setup-python` don't work with custom docker images. There are reported issues (eg. https://github.com/actions/setup-python/issues/719), but they aren't solved.

I also updated `pip3 install ...` to `python3 -m pip install ...` to always use the same python version (setup-python don't configure pip so it is used system wide instead of uploaded by setup-python).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,12 +49,14 @@ jobs:
       - name: Install dependencies
         run: |
           python3 -m venv venv
-          source venv/bin/activate
+          . venv/bin/activate
           python3 -m pip install --upgrade pip
           python3 -m pip install -r requirements-dev.txt
 
       - name: Generate Verilog
-        run: PYTHONHASHSEED=0 ./scripts/gen_verilog.py --verbose --config full
+        run: |
+          . venv/bin/activate
+          PYTHONHASHSEED=0 ./scripts/gen_verilog.py --verbose --config full
 
       - uses: actions/download-artifact@v3
         with:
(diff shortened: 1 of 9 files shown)


lekcyjna123:
Benchmark job:
https://github.com/kuznia-rdzeni/coreblocks/actions/runs/6517968307

tilk (approved):
Seems to work fine.

.github/workflows/main.yml
```diff
@@ -15,7 +15,7 @@ jobs:
   build-regression-tests:
     name: Build regression tests
     runs-on: ubuntu-latest
-    container: ghcr.io/kuznia-rdzeni/riscv-toolchain:2023.05.14
+    container: ghcr.io/lekcyjna123/riscv-toolchain:2023.10.08_v
```
piotro888, line 18: Can we push the new Docker image tag to our repo?

tilk: Good catch!

lekcyjna123: fixed

Arusekk (approved):
Should work fine :) especially with https://github.com/kuznia-rdzeni/coreblocks/issues/384#issuecomment-1756057153 out of the way.

PR #477, opened 2023-10-17
Autumn cleaning part 4 - split test/common.py

lekcyjna123:
As I have said, I have prepared a split of `test/common.py`.
Some remarks:
- Types which went from `common.py` to `transactron/_utils.py` aren't in final place. Probably they should go into `transactron/core/typing.py`, but I don't want to do merge conflicts and `_utils.py` is on the list to be refactored.
- I would like to know you opinion about `data_layout` I moved it to `_utils.pu`, but now I think that this function should probably stay in `test/common/functions.py`.


test/common/functions.py
```diff
@@ -0,0 +1,28 @@
+from amaranth import *
+from typing import TYPE_CHECKING
+from transactron._utils import RecordValueDict, RecordIntDict
+
+
+if TYPE_CHECKING:
+    from .testbenchio import TestGen
```
tilk, line 7: You could avoid the circular dependency by moving `TestGen` out of `testbenchio`.

transactron/_utils.py
```diff
@@ -150,3 +154,64 @@ def get_caller_class_name(default: Optional[str] = None) -> tuple[Optional[Elabo
         return None, default
     else:
         raise RuntimeError("Not called from a method")
+
+
+def data_layout(val: int) -> LayoutLike:
```
tilk, line 159: Argument could be generalized now to `ShapeLike`.

transactron/_utils.py
```diff
@@ -17,6 +17,10 @@
     "method_def_helper",
 ]
 
+RecordIntDict = Mapping[str, Union[int, "RecordIntDict"]]
+RecordIntDictRet = Mapping[str, Any]  # full typing hard to work with
+RecordValueDict = Mapping[str, Union[ValueLike, "RecordValueDict"]]
```
tilk, line 22: Shouldn't they be in `_typing`? (It's currently under `coreblocks`, but almost all of it belongs in `transactron`).

Also, maybe it's a good opportunity to mark those as type aliases?

test/common/functions.py
```diff
@@ -0,0 +1,29 @@
...
+from amaranth.hdl.ast import Statement
+from amaranth.sim.core import Command
+from typing import TypeVar, Any, Generator, TypeAlias
+from coreblocks.utils._typing import RecordValueDict, RecordIntDict
+
+T = TypeVar("T")
+TestGen: TypeAlias = Generator[Command | Value | Statement | None, Any, T]
+
+
+def set_inputs(values: RecordValueDict, field: Record) -> "TestGen[None]":
```
tilk, line 11: The TestGen return type doesn't need to be a string anymore.

tilk (approved):

Durchbruchswagen (approved):

PR #478, opened 2023-10-19
Data dependent method calling

lekcyjna123:
We have currently concept of single caller methods (#416, #425). Such methods allow us to add condition in `ready` signal which depends from input argument. Such conditions are only supported in single caller methods, because when there are two or more callers there is a risk that we end up with combinational cycle. Goal of this PR is to extend data dependent readiness of methods to multi caller methods.

Current implementation of transactron assume that readiness signal is constant for the whole cycle. So we use it to calculate `runnable` signal for each transaction. When ready signal depend from data we can decided **after** scheduling a transaction to execution that method isn't ready. This cause change in `runnable` and an another transaction can be chosen to execution, which data will be potentially accepted by method, so the `ready` will be high one more time and the whole story will repeat.

Solution which I propose in this PR is to calculate data dependent readiness **before** scheduling the transaction, so that we already know that method will accept data from this transaction. The old `ready` signal is used as global readiness for the whole method for backward compatibility.

Additionally I modified types in `method_uses` to be more strict. Instead of `ValueLike` only `Record` is accepted. This allow to properly type input argument for the `user_ready_function`.

### DIFF
--- a/transactron/core.py
+++ b/transactron/core.py
@@ -63,7 +63,7 @@ class MethodMap:
     def __init__(self, transactions: Iterable["Transaction"]):
         self.methods_by_transaction = dict[Transaction, list[Method]]()
         self.transactions_by_method = defaultdict[Method, list[Transaction]](list)
-        self.arguments_by_method_by_transaction = defaultdict[Transaction, dict[Method, RecordDict]](dict)
+        self.arguments_by_method_by_transaction = defaultdict[Transaction, dict[Method, Record]](dict)
 
         def rec(transaction: Transaction, source: TransactionBase):
             for method, (arg_rec, _) in source.method_uses.items():
@@ -129,9 +129,7 @@ def eager_deterministic_cc_scheduler(
     ccl = list(cc)
     ccl.sort(key=lambda transaction: porder[transaction])
     for k, transaction in enumerate(ccl):
-        for method in method_map.methods_by_transaction[transaction]:
-            method.ready_function(method_map.arguments_by_method_by_transaction[transaction][method])
-        ready = [method.ready for method in method_map.methods_by_transaction[transaction]]
+        ready = [method.ready_function(method_map.arguments_by_method_by_transaction[transaction][method]) for method in method_map.methods_by_transaction[transaction]]
         runnable = Cat(ready).all()
         conflicts = [ccl[j].grant for j in range(k) if ccl[j] in gr[transaction]]
         noconflict = ~Cat(conflicts).any()
@@ -682,7 +680,7 @@ class TransactionBase(Owned):
     name: str
 
     def __init__(self):
-        self.method_uses: dict[Method, Tuple[ValueLike, ValueLike]] = dict()
+        self.method_uses: dict[Method, Tuple[Record, ValueLike]] = dict()
         self.relations: list[RelationBase] = []
         self.simultaneous_list: list[TransactionOrMethod] = []
         self.independent_list: list[TransactionOrMethod] = []
@@ -718,7 +716,7 @@ class TransactionBase(Owned):
         """
         self.relations.append(RelationBase(end=end, priority=Priority.LEFT, conflict=False))
 
-    def use_method(self, method: "Method", arg: ValueLike, enable: ValueLike):
+    def use_method(self, method: "Method", arg: Record, enable: ValueLike):
         if method in self.method_uses:
             raise RuntimeError(f"Method '{method.name}' can't be called twice from the same transaction '{self.name}'")
         self.method_uses[method] = (arg, enable)
@@ -983,6 +981,7 @@ class Method(TransactionBase):
         self.data_out = Record(o)
         self.nonexclusive = nonexclusive
         self.single_caller = single_caller
+        self.user_ready_function : Optional[Callable[[Record], ValueLike]] = None
         if nonexclusive:
             assert len(self.data_in) == 0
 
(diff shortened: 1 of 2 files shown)


tilk:
> Solution which I propose in this PR is to calculate data dependent readiness before scheduling the transaction, so that we already know that method will accept data from this transaction.

I find this explanation very unintuitive. "Before" and "after" doesn't matter for combinatorial logic; only thing that exists are data dependencies. What you are doing, in actuality, is generating separate circuits for the readiness condition *for each calling transaction*. This relaxes the previous restriction on ready signals, but at the cost of a (potentially small) increase in circuit size.

The idea seems sound; I'll review the PR later.

transactron/core.py
```diff
@@ -1076,6 +1090,11 @@ def body(self, m: TModule, *, ready: ValueLike = C(1), out: ValueLike = C(0, 0))
         finally:
             self.defined = True
 
+    def ready_function(self, arg_rec: Record) -> ValueLike:
+        if self.user_ready_function is not None:
+            return self.ready & self.user_ready_function(arg_rec)
+        return self.ready
```
tilk, line 1096: This function is not a part of the `Method`'s public interface. It should probably be single-underscored.

transactron/core.py
```diff
@@ -1022,7 +1028,14 @@ def _(arg):
...
 
     @contextmanager
-    def body(self, m: TModule, *, ready: ValueLike = C(1), out: ValueLike = C(0, 0)) -> Iterator[Record]:
+    def body(
+        self,
+        m: TModule,
+        *,
+        ready: ValueLike = C(1),
+        out: ValueLike = C(0, 0),
+        user_ready_function: Optional[Callable[[Record], ValueLike]] = None,
```
tilk, line 1037: I believe that. for consistency, the ready function should follow the calling conventions of methods: the input record should be split into individual fields in parameters. 

Also, the new argument should be documented.

transactron/core.py
```diff
@@ -63,15 +63,17 @@ class MethodMap:
     def __init__(self, transactions: Iterable["Transaction"]):
         self.methods_by_transaction = dict[Transaction, list[Method]]()
         self.transactions_by_method = defaultdict[Method, list[Transaction]](list)
+        self.arguments_by_method_by_transaction = defaultdict[Transaction, dict[Method, Record]](dict)
```
tilk, line 66: I have mixed feelings about this.

* The name is awful.
* The structure you give here is unnecessary: this isn't used for traversing, and doesn't need to, because the information is already in `methods_by_transaction`. A `dict` from pairs to `Record`s should be enough.

transactron/core.py
```diff
@@ -127,7 +129,10 @@ def eager_deterministic_cc_scheduler(
     ccl = list(cc)
     ccl.sort(key=lambda transaction: porder[transaction])
     for k, transaction in enumerate(ccl):
-        ready = [method.ready for method in method_map.methods_by_transaction[transaction]]
+        ready = [
```
tilk, line 132: I leave this comment here, because I can't put a comment on unchanged lines.

I know that `trivial_roundrobin_cc_scheduler` is currently not usable in general, but it's still here, and should be modified accordingly.

tilk (changes requested):
This change is pretty cool actually, but some cleanups are required.

tilk:
> The old `ready` signal is used as global readiness for the whole method for backward compatibility.

This is not just backward compatibility. Both things have a different performance characteristic: there is only one combinatorial circuit generating the old `ready` signal, but the ready function is instantiated for each transaction calling the method.

transactron/core.py
tilk: Check if this can be done outside of the schedulers.

transactron/core.py
```diff
@@ -1162,7 +1186,9 @@ def debug_signals(self) -> SignalBundle:
         return [self.ready, self.run, self.data_in, self.data_out]
 
 
-def def_method(m: TModule, method: Method, ready: ValueLike = C(1)):
+def def_method(
+    m: TModule, method: Method, ready: ValueLike = C(1), ready_function: Optional[Callable[[Record], ValueLike]] = None
```
tilk, line 1190: The type of `ready_function` is now wrong.

transactron/core.py
```diff
@@ -1060,6 +1073,11 @@ def body(self, m: TModule, *, ready: ValueLike = C(1), out: ValueLike = C(0, 0))
             Data generated by the `Method`, which will be passed to
             the caller (a `Transaction` or another `Method`). Assigned
             combinationally to the `data_out` attribute.
+        ready_function: Optional[Callable[[Record], ValueLike]]
+            Function to instantiate a combinational circuit for each
+            method caller. It should take input arguments and return
+            if the method can be called with those arguments. By default
+            there is no function, so all arguments are accepted.
```
tilk, line 1080: I think the ordering is wrong; the first sentence is about an implementation detail, and the second tells us what the function actually does.

tilk (commented):
Left some minor comments.

transactron/core.py
lekcyjna123: I have changed it into the similar type as taken by `method_def_helper`. I hope, that now this is fine.

transactron/core.py
```diff
@@ -1060,6 +1073,11 @@ def body(self, m: TModule, *, ready: ValueLike = C(1), out: ValueLike = C(0, 0))
             Data generated by the `Method`, which will be passed to
             the caller (a `Transaction` or another `Method`). Assigned
             combinationally to the `data_out` attribute.
+        ready_function: Optional[Callable[[Record], ValueLike]]
```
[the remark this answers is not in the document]
lekcyjna123: Changed to `validate_arguments`

tilk (approved):

tilk:
Merged without a second review.

PR #481, opened 2023-10-22
Fix memory protection in regression tests

piotro888:
I found a bug in #466 - all memory segments were set to all RWX flags. After fixing that, regression tests started to fail.

First bug is simple - wrong flag on data memory (only in cocotb backend)
Second bug is problematic. In `rv32uc-rvc` test there is a `data` and not `.data` (!) section that is embedded into instruction memory. This section is written by rvc tests, but not flagged as writeable in ELF file. I don't have currently solution to it (ignore for this specific test, or maybe data section can be added to linker script?)

### DIFF
--- a/test/regression/cocotb.py
+++ b/test/regression/cocotb.py
@@ -147,7 +147,7 @@ class CocotbSimulation(SimulationBackend):
         instr_wb = WishboneSlave(self.dut, "wb_instr", self.dut.clk, mem_model, is_instr_bus=True)
         cocotb.start_soon(instr_wb.start())
 
-        data_wb = WishboneSlave(self.dut, "wb_data", self.dut.clk, mem_model, is_instr_bus=True)
+        data_wb = WishboneSlave(self.dut, "wb_data", self.dut.clk, mem_model, is_instr_bus=False)
         cocotb.start_soon(data_wb.start())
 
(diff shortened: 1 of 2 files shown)


piotro888:
This data section is [here](https://github.com/riscv-software-src/riscv-tests/blob/d020e2069a9f6a9c0e875f23f0f4aababea9fbf0/isa/rv64uc/rvc.S#L29). Does someone know how this can be added to linker script, so instruction LOAD segment would be also writeable?

piotro888:
I decided to manually exclude tests from write protection. It will be also useful for tests with self-modifying code, which can't be detected for linker script.

piotro888:
@tilk Can you change the https://github.com/orgs/kuznia-rdzeni/packages/container/package/riscv-toolchain package visibility from internal, please? If we are working on forks, it can't be pulled for manual run of CI benchmark workflow on forked branch.

tilk:
Done.

tilk:
> Second bug is problematic. In `rv32uc-rvc` test there is a `data` symbol and not `.data` (!) section that is embedded into instruction memory.

If I understand correctly, this test has no `.data` section, and the `data` symbol is just for padding in the code section. I don't see why this should cause a problem.

lekcyjna123 (approved):

tilk (approved):

PR #486, opened 2023-10-24
Refactor def_helper

tilk:
This PR untangles different use cases of `method_def_helper` and prepares the code for further re-use.

### DIFF
--- a/transactron/_utils.py
+++ b/transactron/_utils.py
@@ -18,7 +18,6 @@ __all__ = [
     "get_caller_class_name",
     "def_helper",
     "method_def_helper",
-    "transformer_helper",
 ]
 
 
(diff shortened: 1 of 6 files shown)


transactron/_utils.py
```diff
@@ -122,24 +125,35 @@ def _graph_ccs(gr: ROGraph[T]) -> list[GraphCC[T]]:
 MethodLayout: TypeAlias = LayoutLike
 
 
-def method_def_helper(method, func: Callable[..., T], arg=None, /, **kwargs) -> T:
+def has_first_param(func: Callable[..., T], name: str, tp: type[U]) -> TypeGuard[Callable[Concatenate[U, ...], T]]:
+    parameters = signature(func).parameters
+    return (
+        len(parameters) > 1
+        and next(iter(parameters.keys())) == name
```
Arusekk, line 132: Would work even shorter:
```suggestion
        and next(iter(parameters)) == name
```
but might be less explicit.

Arusekk (approved):
I have one general comment. Otherwise LGTM

transactron/_utils.py
```diff
@@ -122,24 +125,35 @@ def _graph_ccs(gr: ROGraph[T]) -> list[GraphCC[T]]:
...
     )
-    if (
-        len(parameters) == 1
-        and "arg" in parameters
-        and parameters["arg"].kind in {Parameter.POSITIONAL_OR_KEYWORD, Parameter.POSITIONAL_ONLY}
-        and parameters["arg"].annotation in {Parameter.empty, Record}
-    ):
-        if arg is None:
-            arg = kwargs
+    if len(parameters) == 1 and has_first_param(func, "arg", tp):
```
lekcyjna123, line 143: This condition will be always false. If `len(parameters)==1` will be true then `len(parameters) > 1` in `has_first_param` will be false, so `has_first_param` will return false.

tilk: Good catch. The condition in `has_first_param` should have been `>= 1`.

lekcyjna123 (approved):

PR #489, opened 2023-10-25
Better transformer API

tilk:
This PR utilizes `def_helper` mechanisms to create a nicer API for using transformers. Following changes in interpreting the function arguments of the transformer classes are made:

* The module parameter is now optional.
* Instead of taking a record of method inputs (`arg`) the functions can take individual record fields, just like methods defined via `def_method`.

To do:
* [ ] Update docstrings.

### DIFF
--- a/transactron/_utils.py
+++ b/transactron/_utils.py
@@ -1,11 +1,13 @@
 import itertools
 import sys
+import functools
 from inspect import Parameter, signature
 from typing import Any, Concatenate, Optional, TypeAlias, TypeGuard, TypeVar
 from collections.abc import Callable, Iterable, Mapping
 from amaranth import *
 from coreblocks.utils._typing import LayoutLike
 from coreblocks.utils import OneHotSwitchDynamic
+from .core import Method, TModule
 
 __all__ = [
     "Scheduler",
@@ -17,6 +19,7 @@ __all__ = [
     "get_caller_class_name",
     "def_helper",
     "method_def_helper",
+    "transformer_helper",
 ]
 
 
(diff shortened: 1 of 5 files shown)


transactron/_utils.py
```diff
@@ -134,6 +139,13 @@ def has_first_param(func: Callable[..., T], name: str, tp: type[U]) -> TypeGuard
...
 
 
+def bind_first_param(func: CallableOptParam[U, P, T], name: str, tp: type[U], arg: U) -> Callable[P, T]:
+    if has_first_param(func, name, tp):
+        return functools.partial(func, arg)  # type: ignore
+    else:
+        return func  # type: ignore
+
+
 def def_helper(description, func: Callable[..., T], tp: type[U], arg: U, /, **kwargs) -> T:
```
lekcyjna123, line 149: Shouldn't be `func` annotated as `CallableOptParam`? I have the same question regarding `mock_def_helper` and `method_def_helper`.

lekcyjna123 (commented):
I have mixed feelings. I like the simplicity of interface and the reduction of boilerplate. But we are loosing consistency. Earlier everywhere first argument had to be `TModule`. Now in transformer API it wouldn't be needed, but e.g. in method calls it will be still required.

tilk:
> But we are loosing consistency.

I think we are actually gaining consistency, as:
* Keyword arguments allows to define functions just like method definitions.
* The filter function can now be defined consistently with your `validate_arguments` feature. The validating function cannot get a `TModule` argument, because it's called outside of the transactional circuit (in the transaction scheduler).

Also:

* The `TModule` argument was basically a hack to allow the use of methods as functions. I expect this argument will not be needed in most use cases.
* The presence or lack of the `TModule` argument actually signals something meaningful: it says that the function can do method calls in the context of the transformer. (There is unfortunately no way to guarantee "purity", as a function might use an external `TModule` to do some stuff. I sure hope this use will be avoided though.)

transactron/_utils.py
tilk: It could, but it basically changes nothing because of the `...` parameter specification.

PR #493, opened 2023-10-28
retirement: Disable side effects on exception

Arusekk:
Split from #348 to make review easier and get somewhere in finite time.

In scope of this PR:
- [x] A. precommit methods accept an extra `side_fx` input
- [x] B. LSU respects the `side_fx` input
  - [ ] B'. does it wait for pending reads to finish? no idea! current impl skips processing an instruction immediately as soon as precommit is called; it might as well be utter nonsense, because I can't even reason about it, feedback appreciated
- [x] C. CSR respects the `side_fx` input (simply skips calling `read`/`write` handlers)
- [x] D. when side effects are disabled, FRAT gets partially restored (we might want to revisit the priorities of transactions accessing `frat.rename`; they won't starve for sure, but flushing might get faster, I don't know)
- [x] E. when an exception happens, the side effects get disabled
- [x] F. BONUS: regression tests in assembly that actually break when side_fx.eq(0) is commented out!
  - [ ] F'. tests for whether LSU disables its side fx correctly; feedback appreciated, because we might want to extend the assembly tests facility for testing memory contents (mightn't we?)

Out of scope:
- Unit tests for LSU `side_fx=0` itself without the rest of the core? I hope they are out of scope, otherwise help is desperately needed. Might actually help diagnosing B'.
- Fiddling with the fetcher. It still ingests instructions after this PR, they just get ignored in the retirement phase. In the future we might want to just call `stall()` exported as a method.
- Fixing CSRs so that they actually signal important (i.e. ISA-changing or w/e) side effects through a 'benign exception' instead of stalling the fetcher. But it would be nice to get it once exceptions are quasi-ready.
- Fiddling with the jump/branch unit to signal mispredictions through benign exceptions. If ain't broke, don't fix it (yet) :wink:

### DIFF
--- a/coreblocks/lsu/dummyLsu.py
+++ b/coreblocks/lsu/dummyLsu.py
@@ -219,7 +219,10 @@ class LSUDummy(FuncBlock, Elaboratable):
 
         m.submodules.internal = internal = LSUDummyInternals(self.gen_params, self.bus, current_instr)
 
-        result_ready = internal.result_ready | ((current_instr.exec_fn.op_type == OpType.FENCE) & current_instr.valid)
+        result_ready = Signal()
+        m.d.comb += result_ready.eq(
+            internal.result_ready | ((current_instr.exec_fn.op_type == OpType.FENCE) & current_instr.valid)
+        )
 
         @def_method(m, self.select, ~reserved)
         def _():
@@ -256,8 +259,10 @@ class LSUDummy(FuncBlock, Elaboratable):
             }
 
         @def_method(m, self.precommit)
-        def _(rob_id: Value):
-            with m.If(
+        def _(rob_id: Value, side_fx: Value):
+            with m.If(~side_fx):
+                m.d.comb += result_ready.eq(1)
+            with m.Elif(
                 current_instr.valid & (rob_id == current_instr.rob_id) & (current_instr.exec_fn.op_type != OpType.FENCE)
             ):
                 m.d.comb += internal.execute.eq(1)
--- a/coreblocks/stages/retirement.py
+++ b/coreblocks/stages/retirement.py
@@ -70,6 +70,7 @@ class Retirement(Elaboratable):
             rp_freed = Signal(self.gen_params.phys_regs_bits)
             with m.If(side_fx):
                 m.d.comb += rp_freed.eq(rat_out.old_rp_dst)
+                self.instret_csr.increment(m)
             with m.Else():
                 m.d.comb += rp_freed.eq(rob_entry.rob_data.rp_dst)
                 # free the phys_reg with computed value and restore old reg into FRAT as well
@@ -81,7 +82,5 @@ class Retirement(Elaboratable):
             # put old rp_dst to free RF list
             with m.If(rp_freed):  # don't put rp0 to free list - reserved to no-return instructions
                 self.free_rf_put(m, rp_freed)
-
-            self.instret_csr.increment(m)
 
         return m
--- a/coreblocks/stages/retirement.py
+++ b/coreblocks/stages/retirement.py
@@ -18,7 +18,8 @@ class Retirement(Elaboratable):
         free_rf_put: Method,
         rf_free: Method,
         precommit: Method,
-        exception_cause_get: Method
+        exception_cause_get: Method,
+        frat_rename: Method,
     ):
         self.gen_params = gen_params
         self.rob_peek = rob_peek
@@ -28,6 +29,7 @@ class Retirement(Elaboratable):
         self.rf_free = rf_free
         self.precommit = precommit
         self.exception_cause_get = exception_cause_get
+        self.rename = frat_rename
 
         self.instret_csr = DoubleCounterCSR(gen_params, CSRAddress.INSTRET, CSRAddress.INSTRETH)
 
@@ -36,18 +38,23 @@ class Retirement(Elaboratable):
 
         m.submodules.instret_csr = self.instret_csr
 
+        side_fx = Signal(reset=1)
+
         with Transaction().body(m):
             # TODO: do we prefer single precommit call per instruction?
             # If so, the precommit method should send an acknowledge signal here.
             # Just calling once is not enough, because the insn might not be in relevant unit yet.
             rob_entry = self.rob_peek(m)
-            self.precommit(m, rob_id=rob_entry.rob_id)
+            self.precommit(m, rob_id=rob_entry.rob_id, side_fx=side_fx)
 
         with Transaction().body(m):
             rob_entry = self.rob_retire(m)
 
             # TODO: Trigger InterruptCoordinator (handle exception) when rob_entry.exception is set.
-            with m.If(rob_entry.exception):
+            with m.If(rob_entry.exception & side_fx):
+                m.d.sync += side_fx.eq(0)
+                # TODO: only set mcause/trigger IC if cause is actual exception and not e.g.
+                # misprediction or pipeline flush after some fence.i or changing ISA
                 mcause = self.gen_params.get(DependencyManager).get_dependency(GenericCSRRegistersKey()).mcause
                 cause = self.exception_cause_get(m).cause
                 entry = Signal(self.gen_params.isa.xlen)
@@ -56,13 +63,24 @@ class Retirement(Elaboratable):
                 mcause.write(m, entry)
 
             # set rl_dst -> rp_dst in R-RAT
-            rat_out = self.r_rat_commit(m, rl_dst=rob_entry.rob_data.rl_dst, rp_dst=rob_entry.rob_data.rp_dst)
+            rat_out = self.r_rat_commit(
+                m, rl_dst=rob_entry.rob_data.rl_dst, rp_dst=rob_entry.rob_data.rp_dst, side_fx=side_fx
+            )
 
-            self.rf_free(m, rat_out.old_rp_dst)
+            rp_freed = Signal(self.gen_params.phys_regs_bits)
+            with m.If(side_fx):
+                m.d.comb += rp_freed.eq(rat_out.old_rp_dst)
+            with m.Else():
+                m.d.comb += rp_freed.eq(rob_entry.rob_data.rp_dst)
+                # free the phys_reg with computed value and restore old reg into FRAT as well
+                # TODO: are method priorities enough?
+                self.rename(m, rl_s1=0, rl_s2=0, rl_dst=rob_entry.rob_data.rl_dst, rp_dst=rat_out.old_rp_dst)
+
+            self.rf_free(m, rp_freed)
 
             # put old rp_dst to free RF list
-            with m.If(rat_out.old_rp_dst):  # don't put rp0 to free list - reserved to no-return instructions
-                self.free_rf_put(m, rat_out.old_rp_dst)
+            with m.If(rp_freed):  # don't put rp0 to free list - reserved to no-return instructions
+                self.free_rf_put(m, rp_freed)
 
             self.instret_csr.increment(m)
 
--- a/coreblocks/core.py
+++ b/coreblocks/core.py
@@ -136,6 +136,7 @@ class Core(Elaboratable):
             rf_free=rf.free,
             precommit=self.func_blocks_unifier.get_extra_method(InstructionPrecommitKey()),
             exception_cause_get=self.exception_cause_register.get,
+            frat_rename=frat.rename,
         )
 
         m.submodules.csr_generic = self.csr_generic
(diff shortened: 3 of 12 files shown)


tilk:
> does it wait for pending reads to finish? no idea! current impl skips processing an instruction immediately as soon as precommit is called; it might as well be utter nonsense, because I can't even reason about it, feedback appreciated.

I believe that:

* By setting `result_ready` in `LSUDummy` to 1, you allow `get_result` to run - but, as you are doing an combinational assignment, this only works if `get_result` is called in the same cycle!
* Even though side effects were cancelled, if the read ends with an error, the exception will still end up reported. But the instruction was already committed!
* We will need support for memory mapped I/O regions SOON. Reads to MMIO need to wait for precommit, just like writes. Otherwise we lose our fun LiteX demo, as LiteX-UART reads are side-effecting.

And, unrelated to your changes:

* When I think about it, communication of `execute` seems to have the same problem as yours with `get_result`: if `bus.request` is not ready, the signal will be missed. Clearly, we need a simple way to safely communicate conditions between transactions. Maybe we already have it - this feels like a job for a `Forwarder`.

@lekcyjna123, @piotro888

Arusekk:
> this only works if get_result is called in the same cycle!

Actually precommit is a special method and therefore executes every cycle until the instruction gets committed, so I am pretty sure it also works. I also don't quite like it, we should probably make it fit transactron ecosystem better.

> Even though side effects were cancelled, if the read ends with an error, the exception will still end up reported. But the instruction was already committed!

Right, this has a worse implication: if there are two failed reads in the pipeline, the fresh one takes precedence. A simple and versatile fix would be to ignore exception causes right in retirement stage if `side_fx` was set to 0 already.
- [ ] TODO(self): implement that.

> We will need support for memory mapped I/O regions SOON. Reads to MMIO need to wait for precommit, just like writes. Otherwise we lose our fun LiteX demo, as LiteX-UART reads are side-effecting.

This is actually counter-intuitively easier and already correct: `side_fx` can only be disabled on instruction commit boundary, so precommit either gets it disabled all the time, or enabled all the time: they can just be skipped entirely (and this is the case). I'm only worried by reads that have already started, because they can no longer be 'skipped'. And I'm afraid they currently *cannot be skipped at all* and must fully execute, because we don't have access to side_fx in exec stage, only in precommit (unless exec happens during precommit). (Is it a bad thing though? We've already given up on optimizing miss penalty at the expense of simplicity.) I'm still not sure if letting them finish execution is handled correctly.

This brings me to an idea of *even dumber* dummyLSU: assume all L/S is MMIO & only do it during precommit (horrible performance, but its code would be clean as never before; sounds like a Good First Issue for educational purposes). This would of course impair throughput, which is not nice, so it could only be useful as an extra option for debugging, not a replacement.

coreblocks/lsu/dummyLsu.py
```diff
@@ -256,8 +259,10 @@ def _():
             }
 
         @def_method(m, self.precommit)
-        def _(rob_id: Value):
-            with m.If(
+        def _(rob_id: Value, side_fx: Value):
+            with m.If(~side_fx):
+                m.d.comb += result_ready.eq(1)
```
lekcyjna123, line 264: Instead of overwriting results of `result_ready` maybe it will be better to add a next `or` to the `result_ready` expression? This will create a simpler verilog. See:
```v
module top(in2, in3, out1, in1);
  reg \$auto$verilog_backend.cc:2097:dump_module$1  = 0;
  wire \$1 ;
  input in1;
  wire in1;
  input in2;
  wire in2;
  input in3;
  wire in3;
  output out1;
  reg out1;
  assign \$1  = in1 | in2;
  always @* begin
    if (\$auto$verilog_backend.cc:2097:dump_module$1 ) begin end
    out1 = \$1 ;
    casez (in3)
      1'h1:
          out1 = 1'h1;
    endcase
  end
endmodule
```
and:
```v
module top(in2, in3, out1, in1);
  wire \$1 ;
  wire \$3 ;
  input in1;
  wire in1;
  input in2;
  wire in2;
  input in3;
  wire in3;
  output out1;
  wire out1;
  assign \$1  = in1 | in2;
  assign \$3  = \$1  | in3;
  assign out1 = \$3 ;
endmodule

```

lekcyjna123:
If read started an execution it have to be fully executed, to don't left garbage in the data bus. I see two ways to take care about this:
- add clearing of the bus - after we execute all instruction, whose side effects should be skipped, we can clear the bus buffers to remove garbage
- check in LSU if we started execution (`internal.instr_ready == 1`) if yes we have to wait till the end of instruction. If not we can block the `update` and block the instruction execution by setting `rp_s1` and `rp_s2` to non zero. Execution of `get_result` after blocking execution is safe.

coreblocks/stages/retirement.py
```diff
@@ -56,13 +63,24 @@ def elaborate(self, platform):
...
 
             # set rl_dst -> rp_dst in R-RAT
-            rat_out = self.r_rat_commit(m, rl_dst=rob_entry.rob_data.rl_dst, rp_dst=rob_entry.rob_data.rp_dst)
+            rat_out = self.r_rat_commit(
+                m, rl_dst=rob_entry.rob_data.rl_dst, rp_dst=rob_entry.rob_data.rp_dst, side_fx=side_fx
+            )
+
+            rp_freed = Signal(self.gen_params.phys_regs_bits)
+            with m.If(side_fx):
+                m.d.comb += rp_freed.eq(rat_out.old_rp_dst)
```
lekcyjna123, line 72: This can be assigned in `av_comb`.

tilk:
> Actually precommit is a special method and therefore executes every cycle until the instruction gets committed, so I am pretty sure it also works. I also don't quite like it, we should probably make it fit transactron ecosystem better.

I believe you are right - this is probably just fine.

> I'm only worried by reads that have already started, because they can no longer be 'skipped'. And I'm afraid they currently cannot be skipped at all and must fully execute, because we don't have access to side_fx in exec stage, only in precommit (unless exec happens during precommit).

This is exactly the problem I'm talking about - the code as it is written allows MMIO reads to go out to the bus before we know they should actually happen. And, as I said, there is a simple way to correct this: just make the MMIO reads (but not memory reads) wait for precommit, just like all writes do currently.

coreblocks/lsu/dummyLsu.py
```diff
@@ -256,8 +259,10 @@ def _():
             }
 
         @def_method(m, self.precommit)
-        def _(rob_id: Value):
-            with m.If(
+        def _(rob_id: Value, side_fx: Value):
+            with m.If(~side_fx):
```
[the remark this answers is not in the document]
Arusekk: Would it though? The first condition must consider side_fx anyway, so it would effectively check side_fx twice.

coreblocks/lsu/dummyLsu.py
Arusekk: Nice, how did you get that?

tilk: We're not optimizing the Verilog output, we are optimizing readability first, performance next. Performance is obviously the same for both, so the only question is, what is more readable.

coreblocks/stages/retirement.py
```diff
@@ -56,13 +63,24 @@ def elaborate(self, platform):
...
-            self.rf_free(m, rat_out.old_rp_dst)
+            self.rf_free(m, rp_freed)
 
             # put old rp_dst to free RF list
-            with m.If(rat_out.old_rp_dst):  # don't put rp0 to free list - reserved to no-return instructions
-                self.free_rf_put(m, rat_out.old_rp_dst)
+            with m.If(rp_freed):  # don't put rp0 to free list - reserved to no-return instructions
+                self.free_rf_put(m, rp_freed)
 
             self.instret_csr.increment(m)
```
piotro888, line 85: `instret` CSR should only be incremented on `side_fx` instructions (and without `rob_entry.exception` instruction too)

piotro888: and that is additional source of possible side effects and should be considered in future FUs

lekcyjna123:
I checked reads and it wouldn't work:
- start a read (send a request on the bus)
- precommit with `side_fx==0` cause `result_ready==1`
- this cause execution of `get_result` which indicates `internal.get_result_ack==1` and `reserved==0`
- `internal.get_result_ack==1`  cause change of state machine in LSUDummyInternals to `Start`
- `reserved ==0` so we can `select` and `insert` new instruction
- new load is inserted and passed to execution
- data from first load are ready on the bus

lekcyjna123:
Short and dirty fix:
```py
        @def_method(m, self.precommit)
        def _(rob_id: Value, side_fx: Value):
            with m.If(~side_fx):
                m.d.comb += result_ready.eq(internal.fsm.state=="Start")
```
and
```py

        instr_ready = (
            (self.current_instr.rp_s1 == 0)
            & (self.current_instr.rp_s2 == 0)
            & self.current_instr.valid
            & ~self.result_ready
            & self.side_fx
        )
```
So instr is ready only if `side_fx==1` and we can not run `get_ready` method as long as the state machine is not in "Start" state/

tilk:
> So instr is ready only if `side_fx==1` and we can not run `get_ready` method as long as the state machine is not in "Start" state/

Isn't this equivalent to removing the `is_load` condition in the `Start` state? Also, there is no `get_ready` method, there is `get_result`, but it depends only on `result_ready`.

Geez, DummyLSU is still a giant liability.

piotro888:
Shouldn't we also check 
```python 
with m.If(~side_fx & current_instr.valid & (rob_id == current_instr.rob_id)):
    m.d.comb += result_ready.eq(internal.fsm.state=="Start")
```
?
Otherwise `get_result` would be ready on each `precommit` call (to different FUs!) and output illegal entries to pipeline in each cycle. Or am I missing something?

tilk:
I'm assuming the following invariant: once a precommit is sent for a given `rob_id`, the value of `side_fx` can't suddenly change while waiting for commit.

I see two solutions: one quick and dirty (and probably fine for now), and one which requires some additional thought.

The simple one: just remove the `| is_load` condition from the if in the `Start` state. This ensures that, if `side_fx` is `False`, no requests are sent nor are any exceptions reported, and everything works just fine. Performance is affected, but we can fix it later. (If this option allows @Arusekk to continue working faster, I'm all for it.)

The complex one: vary the behavior depending on the current state of processing the request. If the request was not sent yet, ensure that it will not be sent (possibly by modifying the condition in the `Start` state). If the request was sent, let it run to completion (so don't assign `result_ready` locally, wait until it's set by the internals module). I don't know if something needs to be done about exception reporting: this would be encountering an exception while processing a previous exception, we need some strategy for that, as this will not be the only place where something like this could happen. I would much prefer having some general solution instead of special-casing in each module which can report exceptions. @piotro888, what are your thoughts?

piotro888:
> I don't know if something needs to be done about exception reporting: this would be encountering an exception while processing a previous exception, we need some strategy for that, as this will not be the only place where something like this could happen.

I think it will work just fine. `ExceptionCauseRegister` is only updated for 'older' instructions (based on `rob_id`), so it will be updated to match the correct (first retired) instruction. If multiple exceptions will happen, we only know about it globally (via `side_fx`) after instruction with exception reaches the retirement - so this is correct, and would not change other behaviour.

Only other special effect of exceptions is setting of `exception` bit on, but it should be easily ignored in retirement if `side_fx` are disabled (and it looks like it is already done)


And I noticed one thing to remember for next part of exceptions integration - we should call `clear()` on `ExceptionCauseRegister` after flush (at the same moment when restoring fetch) to correctly register next exceptions and ignore all that happened when `side_fx` were disabled.

> The complex one:

I'm obviously for the complex solution, and I don't think it will be that difficult to implement. But obviously, as always, this can be done later in other PR (with simple solution workaround for now). And I think it is the right thing to do now - to not delay this PR and exception development.

lekcyjna123:
> The complex one: vary the behavior depending on the current state of processing the request. If the request was not sent yet, ensure that it will not be sent (possibly by modifying the condition in the Start state). If the request was sent, let it run to completion (so don't assign result_ready locally, wait until it's set by the internals module). 

This was an idea behind my changes proposition. Adding `If` in `precommit` to set `result_ready` only on `fsm.state=="Start"` should guarantee that every started request will run to completion. Adding `self.side_fx` to `instr_ready` guarantee that no new request will be started if side effects are turned off.

> Isn't this equivalent to removing the is_load condition in the Start state? 

No, it isn't. I my proposition we still can run loads before `precommiting` the load. We can speculate about value of `side_fx` based on the earlier instructions, if it is `1` then we start a load request immediately. Of course our speculation can fail, so in such case we simply wait till the request will be processed by memory.

piotro888 (commented):
Left comments from meeting. Otherwise looks good

tilk:
Made some changes on a [branch](https://github.com/kuznia-rdzeni/coreblocks/tree/tilk/lazy-exc-2). In particular, I switched to the new, not-yet-merged LSU.

tilk:
This was merged through #512.

Issue #476, opened 2023-10-16
Optimise build regression workflow


lekcyjna123:
In buildregression workflow we always build binaries and store them as artifacts. But binaries are usualy the same as in other jobs, so instead of artifacts a cache should be used.

PR #495, opened 2023-10-30
Cache regression tests in workflow

JumiDeluxe:
Targets https://github.com/kuznia-rdzeni/coreblocks/issues/476

Replaced building `test/external/riscv-tests` directory contents in every run, it is now done on cache miss.
They are now stored using workflow cache instead of artifacts.


.github/workflows/main.yml
```diff
@@ -22,14 +22,23 @@ jobs:
...
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-regression-tests
         with:
-          name: "riscv-tests"
-          path: |
-            test/external/riscv-tests/test-*
+          path: test/external/riscv-tests/test-*
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{
+            hashFiles('**/riscv-tests/test-*') }}
```
lekcyjna123, line 33: We should add to the key `job.container.id` to rebuild cache after docker change

lekcyjna123 (commented):
Overall looks good. Only one change is needed.

lekcyjna123 (approved):

.github/workflows/main.yml
JumiDeluxe: Added the key, also changed cache name to match previous ones in this repo (https://github.com/kuznia-rdzeni/coreblocks/actions/caches).
It is now `${{ env.cache-name }}-${{ runner.os }}-${{ job.container.id }}-`

.github/workflows/main.yml
```diff
@@ -22,14 +22,23 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
```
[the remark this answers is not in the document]
JumiDeluxe: Good point, I've added this cache before looking into this repo existing caches, it's true that this could cause problem in this case, also for future cases.

For now restore keys are:
```yaml
restore-keys: |
            ${{ env.cache-name }}-${{ runner.os }}-${{ job.container.id }}-
            ${{ env.cache-name }}-${{ runner.os }}-
```

It is not surprising that including only the first restore key entry caused jobs to fail, as test running jobs executed in container with different ID than the ones building/caching them.

piotro888:
This was not working correctly because `Build regression tests` and `Run regression tests` uses different docker images. Currently, in second workflow, cache id is partially matched and re-uploaded with second docker id.

Also the last part of cache id (`hashFiles`) doesn't seem to work at all: from logs ```Cache saved with key: cache-regression-tests-Linux-d2ea333ee3c65f5c8876f565d3f8c7a5f3a7669f4cc2a67cd8a23c1f1fd36c63-```

.github/workflows/main.yml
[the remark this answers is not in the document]
piotro888: Maybe in the second action we should depend only on `${{ env.cache-name }}-${{ runner.os }}-` `restore-keys` to download the test? Exact key could never be matched, because we don't know the tests file hash - it will always fall back to restore key - which is the intended behaviour -> download the latest test files from current branch or master.

In this case some always invalid `key:` could be used. Then `fileHash` would also not crash the CI.

The `${{ job.container.id }}-` should not be used here, because it is independent docker image, that doesn't change the uploaded files. 

What do you think?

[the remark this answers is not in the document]
JumiDeluxe: I've changed `fileHash` to calculate contents of `test/external/riscv-tests` directory, it was empty before because it just tried to calculate hash from files that didn't exist before build. And after the change it crashed in fetching phase because it tried to follow `encoding.h`, which is a symlink to a submodule that wasn't fetched in that job.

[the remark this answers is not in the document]
piotro888: I'm afraid it is still incorrect :(

Build action after compilation pushes cache with id: ...-(riscv-toolchain docker id)-(hash of directory with asm test)
Then Run action downloads it, and pushes it again to second cache  ...-(verilator docker id)-(hash of directory with only asm tests). This is the id that would be checked next time in Run action.

I don't know if it is the intended design, in my opinion pushing two caches is not necessary; and in Run action only last/correct cache from Build needs to be downloaded with the same key (and all following problems avoided) (we don't care about ventilator docker - because it is not used to build tests and not outputting anything to cache)

In any case, the current solution would not work when riscv-toolchain changes, because the second cache would still hit (id is not changed in any way)!

I'm not entirely sure about it, but  it looks like if tests are updated, then Run action is a cache miss, but according to current priority it would select  `${{ env.cache-name }}-${{ runner.os }}-${{ job.container.id }}-` prefix first - so the same previous Run cache with different file hash, and not update to new build files.

There is also a warning in logs: `
Run regression tests
Unexpected input(s) 'submodules', valid inputs are ['path', 'key', 'restore-keys', 'upload-chunk-size', 'enableCrossOsArchive', 'fail-on-cache-miss', 'lookup-only']
`
`submodules` should be in checkout stage, that also explains why file hashes are different in both actions.

lekcyjna123:
Please check https://github.com/JumiDeluxe/coreblocks/pull/1 where I added proposition to fix problems reported by @piotro888

lekcyjna123 (approved):

tilk:
I'm sorry this is that late, but I have a little objection. Artifacts do have one feature that caches don't have: they can be downloaded. This is somewhat useful here, as one can get the compiled tests without the need to have the toolchain installed locally. Is there any way to keep the cache, but still be able to download the files if needed?

.github/workflows/main.yml
```diff
@@ -58,10 +69,16 @@ jobs:
...
           PYTHONHASHSEED=0 ./scripts/gen_verilog.py --verbose --config full
 
-      - uses: actions/download-artifact@v3
+      - uses: actions/cache@v3
+        env:
+          cache-name: cache-regression-tests
         with:
-          name: "riscv-tests"
-          path: test/external/riscv-tests
+          submodules: recursive
```
piotro888, line 76:
```suggestion
```
Please remove this line. It is ignored and produces warning in logs (already specified in actions/checkout)

.github/workflows/main.yml
```diff
@@ -58,10 +69,16 @@ jobs:
...
+          path: test/external/riscv-tests/test-*
+          key: ${{ env.cache-name }}-${{ runner.os }}-${{
+            needs.build-regression-tests.outputs.container_id }}-${{ hashFiles('**/test/external/riscv-tests') }}
+          restore-keys: |
+            ${{ env.cache-name }}-${{ runner.os }}-${{ needs.build-regression-tests.outputs.container_id }}-
```
piotro888, line 81:
```suggestion
          fail-on-cache-miss: true
```
I think that failing on cache miss would be helpful here.
"Run" action always runs after "Build" action, so the cache with built test files must be already present (and we want to use that exact cache).

This would also prevent reuploading the cache (and all weird problems with it).

In this case `restore-keys` would not be needed.

piotro888 (changes requested):
As discussed on meeting, it still doesn't work correctly.
The `hashFiles` part of `key` is different in Build than Run action when fetching cache.

@lekcyjna123 found the probable reason for it - when fetching submodules `.git` directory produces different hash each time.

This behaviour causes Run action to firstly fetch the cache from partial match via restore key and then **reupload** it with its own id at end of action (because  of cache miss). This could cause fetching the outdated Run cache (that was reuploaded with Run id) next time, instead of Build cache (that could change and be updated).

Solution for it would be to exclude `.git` directory form `hasFiles` and the keys should always fully match (using `**` at end of path was suggested on meeting to ignore dot-dirs (to verify) )

JumiDeluxe:
> I'm sorry this is this late, but I have a little objection. Artifacts do have one feature that caches don't have: they can be downloaded. This is somewhat useful here, as one can get the compiled tests without the need to have the toolchain installed locally. Is there any way to keep the cache, but still be able to download the files if needed?

I've added conditional step that uploads artifacts, it is executed on cache miss, after build

```yaml
      - if: ${{ steps.cache-regression.outputs.cache-hit != 'true' }}
        run: cd test/external/riscv-tests && make

      - if: ${{ steps.cache-regression.outputs.cache-hit != 'true' }}
        name: Upload riscv-tests
        uses: actions/upload-artifact@v3
        with:
          path: test/external/riscv-tests
```

JumiDeluxe:
```
git ls-remote https://github.com/riscv-software-src/riscv-tests.git HEAD
bd0a19c136927eaa3b7296a591a896c141affb6b        HEAD
```

.github/workflows/main.yml
```diff
@@ -58,10 +87,18 @@ jobs:
...
-      - uses: actions/download-artifact@v3
+      - uses: actions/cache@v3
+        env:
+          cache-name: cache-regression-tests
         with:
-          name: "riscv-tests"
-          path: test/external/riscv-tests
+          path: test/external/riscv-tests/test-*
+          run: |
+            echo "RISCV_COMMIT=$(echo git ls-remote https://github.com/riscv-software-src/riscv-tests.git HEAD | awk '{ print $1 }')" >> $GITHUB_ENV
```
piotro888, line 96: We want to check commit id of our submodule (which is fixed), and not the last commit of riscv-tests repo.
@tilk found that this commit id is stored in `.git/modules/test/external/riscv-tests/riscv-tests/HEAD` file in coreblocks dir. Appending this single file to hashFiles should be enough?

JumiDeluxe: Yes, added it to hashing function and removed `git ls-remote` hash extraction

piotro888 (approved):

PR #499, opened 2023-10-31
Implement LSU clear

Kristopher38:
This PR implements a proper clear method in `dummyLSU` with proper handling in `LSUDummyInternals`. As with most changes, bulk of the code is tests. Test code is a bit copy-paste'y as I didn't go for writing randomized tests (@lekcyjna123 already went through that route and it turned out to be a nightmare) and went with explicitly writing a sequence of steps that are supposed to test different conditions:
- calling clear after instruction has been inserted into the LSU but before bus request was sent
- calling clear after bus request was sent but before response arrived
- calling clear after bus response arrived but wasn't read with `get_result` yet

This merges into `feature/interrupts` since that way I won't have to rebase against it (if it was merged into `master`).

### DIFF
--- a/coreblocks/lsu/dummyLsu.py
+++ b/coreblocks/lsu/dummyLsu.py
@@ -153,14 +153,19 @@ class LSUDummyInternals(Elaboratable):
 
                     m.d.sync += self.loadedData.eq(self.postprocess_load_data(m, fetched.data, addr))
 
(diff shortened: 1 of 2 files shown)


lekcyjna123 (approved):
It looks good and surprisingly simple :) Only one remark, to didn't forget about that in future, but it is an optimisation ;)

PR #500, opened 2023-11-02
Fix build after #466

tilk:
PR #466 by @piotro888 removed zero-padding of segments by mistake, leading to the benchmark build failing. This PR restores the padding (and fixes a typo).


piotro888 (approved):
Sorry, I completely missed the padding, and I was sure that I checked the Benchmark CI (but it turns out, it was for another PR...)

lekcyjna123 (approved):

tilk:
Well, still doesn't work. Looks like the data segments need some padding, too.

PR #502, opened 2023-11-03
Fix build, for real this time

tilk:
Always run the affected workflow before merging. Always run the affected workflow before merging. Always run the affected workflow before merging. Always run the affected workflow before merging.

Summary of the comments: `UnalignedFetcher` sometimes fetches one additional cache line beyond the ones that contain the code. Not really a big problem, but the core freezes on the bus error that can then happen. To work around this issue, an additional cache line is added to code segments.

### DIFF
--- a/test/regression/memory.py
+++ b/test/regression/memory.py
@@ -165,9 +165,10 @@ def load_segment(segment: Segment, *, disable_write_protection: bool = False) ->
     if flags_raw & P_FLAGS.PF_X:
         # align instruction section to full icache lines
         align_bits = config.icache_block_size_bits
+        # workaround for fetching/stalling issue
+        align_bits += 1
     else:
-        # align to memory words
-        align_bits = log2_int(config.xlen // 8)
+        align_bits = 0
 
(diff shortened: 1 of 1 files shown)


tilk:
Looks like this try was misguided. I objdumped the failing tests, and it looks like the core is still somehow trying to fetch data just outside the code segment. In both test programs, the last instruction is compressed `ret`, which is very close to the cache line boundary. I believe it's the `UnalignedFetcher` which makes the additional, unneeded request.

tilk:
I think that the right way to proceed is to stop crashing the simulation when accessing stuff outside of the defined memory ranges. The `UnalignedFetcher` does some speculative prefetching, and prefetching outside of a memory range is not a failure.

tilk:
This didn't help, either. The core freezes. Looks like fixing this properly requires some proper debugging. I think that, for now, we should just work around the issue.

tilk:
Finally, benchmarks work again. The temporary workaround is to add one more cache line to the code segment.

lekcyjna123 (approved):

piotro888 (approved):

PR #395, opened 2023-06-23
[WIP] One vector PR to rule them all

lekcyjna123:
Here is a PR to merge branch with implementation of vector extension. As for now I am going to merge here all already open PR's related with vector extension.

PR #504, opened 2023-11-04
Add `condition` based MetodFilter

lekcyjna123:
This MR is the first fragment of #395 to port to the trunk. Here I extend `MethodFilter` with possibility to use `condition` under the hood to ignore situations, where method is not ready and condition is false. This has the caveats that resulting filter is `single_caller` so both versions should coexistence.

### DIFF
--- a/transactron/lib/transformers.py
+++ b/transactron/lib/transformers.py
@@ -5,6 +5,7 @@ from typing import Optional
 from collections.abc import Callable
 from coreblocks.utils import ValueLike, assign, AssignType
 from .connectors import Forwarder, ManyToOneConnectTrans, ConnectTrans
+from .simultaneous import condition
 
 __all__ = [
     "MethodTransformer",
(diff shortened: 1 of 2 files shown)


lekcyjna123:
A remark to remember is that the usage of `validate_arguments` is not possible in `MethodFilter`, because this feature cause blocking of method, if the condition is not fulfilled, what cause the blocking of the calling transaction.

tilk (approved):

piotro888 (approved):
Approved, but maybe if it doesn't cause any logic penalty this:
```python
               with condition(m, nonblocking=True) as branch:
                    with branch(cond):
                        m.d.comb += ret.eq(self.target(m, arg))
```
would be cleaner? (`nonblocking=True` to automatically generate empty default condition - equivalent to ~cond)

transactron/lib/transformers.py
```diff
@@ -150,8 +160,17 @@ def elaborate(self, platform):
 
         @def_method(m, self.method)
         def _(arg):
-            with m.If(self.condition(m, arg)):
-                m.d.comb += ret.eq(self.target(m, arg))
+            if self.use_condition:
+                cond = Signal()
+                m.d.top_comb += cond.eq(self.condition(m, arg))
+                with condition(m, nonblocking=False) as branch:
```
tilk, line 166: The conditions are mutually exclusive - maybe `priority=False` should be used?

Or remove the second branch and use `nonblocking=True`, as @piotro888 suggested. Right now you use both exclusivity and transaction priorities - do you see any advantage of this, or is this accidental?

tilk: I'm waiting with merge until this is resolved.

tilk (changes requested):
Address the review comment.

transactron/lib/transformers.py
lekcyjna123: We can not use `nonblocking=True` here. The `condition` class doesn't use two state logic. We have:
- True and ready
- False and ready
- Not ready
If the `cond=1` and `target` is not ready, then with `nonblocking=True` the transaction will execute and so the whole method will execute, which will be wrong because the data instead of being forwarded will be dropped.

I added `priority=False`, this shouldn't impact the logic.

tilk (approved):

PR #509, opened 2023-11-11
Extend simulation infrastructure with `Now()`

lekcyjna123:
In some places it is useful to have direct access to the current cycle number. Till now we could manually count cycles, but this required using `call_try` everywhere. This commit ports simulation framework extension from #395. There is added new class `Now()` which can be yielded during testing. In return a current cycle will be returned.

This feature is needed to extend and port `CondVar` from #395
https://github.com/kuznia-rdzeni/coreblocks/blob/ddb21b50c9cbc46d154436b3a48a8cdbe323ceeb/test/common.py#L429-L433
Current `CondVar` notification requires to wait a cycle on `notify`. With usage of `Now()` it should be possible to omit this waiting.


test/common/infrastructure.py
```diff
@@ -99,6 +100,40 @@ def elaborate(self, platform) -> HasElaborate:
...
+
+
+class SyncProcessWrapper:
+    def __init__(self, f):
+        self.org_process = f
+        self.current_cycle = 0
+
+    def _wrapping_function(self):
+        response = None
+        org_corutine = self.org_process()
```
tilk, line 118: coroutine

test/common/infrastructure.py
```diff
@@ -99,6 +100,40 @@ def elaborate(self, platform) -> HasElaborate:
         return m
 
 
+class CoreblockCommand:
```
tilk, line 103: CoreblocksCommand
ABC

test/common/functions.py
```diff
@@ -1,11 +1,16 @@
 from amaranth import *
 from amaranth.hdl.ast import Statement
 from amaranth.sim.core import Command
-from typing import TypeVar, Any, Generator, TypeAlias
+from typing import TypeVar, Any, Generator, TypeAlias, TYPE_CHECKING, Union
 from transactron.utils._typing import RecordValueDict, RecordIntDict
 
+
+if TYPE_CHECKING:
+    from .infrastructure import CoreblocksCommand
```
Kristopher38, line 9: Could you elaborate why this needs to be conditioned on typechecking?

Kristopher38 (approved):
LGTM, left one non-critical question

test/common/functions.py
lekcyjna123: Because in `infrastructure.py` file I import symbols from `functions.py`. So without this `if` there is a circular dependency.

tilk (approved):
Let's finally merge some testing extensions.

PR #510, opened 2023-11-12
Implement physical memory attributes (PMA) for memory-mapped IO (MMIO)

kindlermikolaj:
DRAFT: 
This PR adds basic physical memory attributes implementation, currently with only one attribute (MMIO, that should be defined in core configuration). PMAChecker(Elaboratable)  with method `ask` is used to fetch all attributes for given memory address.


coreblocks/lsu/dummyLsu.py
```diff
@@ -121,15 +123,21 @@ def elaborate(self, platform):
...
 
         with m.FSM("Start"):
             with m.State("Start"):
-                with m.If(instr_ready & (self.execute | is_load)):
+                with Transaction().body(m):
+                    mmio_flag = self.pma.ask(m, in_addr=addr)["mmio"]
+                    m.d.sync += mmio.eq(mmio_flag)
+                    m.next = "Load"
+
+            with m.State("Load"):
```
tilk, line 139: The additional cycle increases latency and decreases throughput. PMAs should work purely combinationally (and they do!) and not introduce any delay cycles.

coreblocks/params/configurations.py
```diff
@@ -59,6 +59,8 @@ class CoreConfiguration:
         Allow partial support of extensions.
     _implied_extensions: Extenstion
         Bit flag specifing enabled extenstions that are not specified by func_units_config. Used in internal tests.
+    mmio : [(int,, int)]
```
tilk, line 62: Suggestion: `[PMARegion]`, where `PMARegion` is a dataclass with three fields: `start`, `end` and `mmio`.

coreblocks/lsu/pma.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def elaborate(self, platform) -> HasElaborate:
+        m = TModule()
+
+        for attr_name in self.attrs:
+            m.submodules[attr_name] = self.attrs[attr_name]
+
+        @def_method(m, self.ask)
+        def _(in_addr: Value):
+            out_signals = dict()
```
tilk, line 117: For each region, check if `addr` is in the region, if it is, output value for the region is taken from PMAs, if not, zero. At the end, "or" everything.

coreblocks/lsu/dummyLsu.py
tilk: Maybe use nested transactions?

tilk: Disregard the previous comment, see new DummyLSU in #511.

kindlermikolaj: Ok, I think that the best solution is to check PMAs during insert. Then, only LSUDummy needs to have access to PMAChecker. MMIO flag should be part of `current_instr` record, since it is required for every instruction.

tilk:
The refactored LSU is now in master. I think it's better for you to stop working on the outdated code and rebase to master ASAP.

coreblocks/lsu/pma.py
```diff
@@ -0,0 +1,134 @@
...
+
+    ATTR_WIDTH = 1
+
+    def __init__(self, gen_params: GenParams) -> None:
+        super().__init__()
+
+        self.attr_width = PMAChecker.ATTR_WIDTH
+        self.ask = Method(i=[("in_addr", gen_params.isa.xlen)], o=gen_params.get(PMALayouts).pma_attrs_layout)
+        self.attrs = dict()
+        self.attrs["mmio"] = PhysMemAttr(gen_params)
```
tilk, line 110: Why are you insisting on mangaging ranges per attribute instead of attributes per range? It's not that each attribute will have completely different set of ranges, a typical situation is that different ranges differ on a lot of attributes.

kindlermikolaj: I'm not insisting on it, I just did PMARegion, refactoring PMAChecker is next on list.

coreblocks/lsu/pma.py
```diff
@@ -0,0 +1,134 @@
...
+    Attributes
+    ----------
+    in_addr : Signal, in
+        Instructs to return attribute of particular memory address.
+    out : Signal, out
+        Response to inquiry about atribute of `in_addr` memory address.
+    """
+
+    @staticmethod
+    def merge_intervals(acc: tuple[tuple[int, int], list], el: tuple[int, int]):
```
tilk, line 31: There is no need to do any merging. You can assume here that the input data is correct and don't worry about optimality. If in doubt, keep things simple.

kindlermikolaj: Done

kindlermikolaj:
One issue I arrived at:
Unless there is a way to assert that method is not ready, it is currently impossible to test if PMAs work because such test would freeze. One solution would be to add a function to AdapterTrans that returns readiness status of method.

Kristopher38:
There's a `call_try` method in the adapter that will call the method if it's ready (and return it's returned data) or return None (iirc) if it's not ready.

coreblocks/params/configurations.py
```diff
@@ -59,6 +62,8 @@ class CoreConfiguration:
...
+        Ranges of memory mapped input output.
```
piotro888, line 66: Update the docstring

coreblocks/lsu/pma.py
```diff
@@ -0,0 +1,76 @@
...
+        Defines beginning of region.
+    end : int
+        Defines end of region.
+    attrs : Record
+        Attributes for given region. Record should follow adequate layout from `PMALayouts`.
+    """
+
+    start: int
+    end: int
+    attrs: Record
```
piotro888, line 27: I think that using single fields like `mmio` of type `Bool` fits better here.
It allows to document each field here, specify default values and the configuration is cleaner.
`Record` could be constructed in elaboratable.

coreblocks/lsu/dummyLsu.py
```diff
@@ -227,6 +238,10 @@ def _():
         def _(rs_data: Record, rs_entry_id: Value):
             m.d.sync += assign(current_instr, rs_data)
             m.d.sync += valid.eq(1)
+            with Transaction().body(m):
+                addr = calculate_addr(m, self.gen_params.isa.xlen, current_instr)
+                mmio_flag = pma.ask(m, addr=addr)["mmio"]
+                m.d.sync += current_instr.mmio.eq(mmio_flag)
```
piotro888, line 244: This won't work here. You use `current_instr.s1_val` in `calculate_addr`, but at this stage this field may not be valid here.

In FUs (RSes) `insert` is called first for each instruction, and register values may be already computed or not. If register value is not yet known, it will be supplied later via `update` call. See `instr_ready` definition.

You need to check that after `instr_valid` is set. Which can be only after `insert` or `insert`+`update`.

Proposed solutions:
1) Move this transaction one level up (not under any method), with run condition of `instr_ready`. The result signal should be updated both in comb domain to affect issue condtion, which also depends on instr_ready. Because of combinational dependency your transaction must be `scheduled_before` the `issue` transaction.
2) Execute `pma.ask` always under `issue` transaction, and execute rest of the code there under `If` of pma result

Maybe someone else sees some better solutions?

coreblocks/lsu/dummyLsu.py
```diff
@@ -7,10 +7,17 @@
...
 
+from coreblocks.lsu.pma import PMAChecker
 
 __all__ = ["LSUDummy", "LSUBlockComponent"]
 
 
+def calculate_addr(m: ModuleLike, xlen: int, current_instr: Record):
+    addr = Signal(xlen)
+    m.d.comb += addr.eq(current_instr.s1_val + current_instr.imm)
+    return addr
```
piotro888, line 18: I don't think this function is needed, it is already done inline in line 262

coreblocks/params/layouts.py
```diff
@@ -520,6 +521,8 @@ def __init__(self, gen_params: GenParams):
             },
         )
 
+        data_layout = layout_superset(base_data_layout, [("mmio", 1)])
```
piotro888, line 524: Why special `layout_superset` is needed? You could just simply do list concatenation.

piotro888 (changes requested):

coreblocks/lsu/pma.py
tilk: Our dataclasses are typically used at the meta-level, for generation. Records are for the circuit level only.

coreblocks/params/configurations.py
```diff
@@ -80,6 +85,8 @@ class CoreConfiguration:
 
     _implied_extensions: Extension = Extension(0)
 
+    pma: List[PMARegion] = field(default_factory=list)
```
tilk, line 88: `List` is deprecated. Use `list`.

transactron/utils/utils.py
```diff
@@ -337,6 +339,11 @@ def layout_subset(layout: LayoutList, *, fields: set[str]) -> LayoutList:
     return [item for item in layout if item[0] in fields]
 
 
+def layout_superset(base_layout: LayoutList, add_layout: LayoutList) -> LayoutList:
+    fields = [key for (key, _) in base_layout]
+    return base_layout + [item for item in add_layout if item[0] not in fields]
```
tilk, line 344: Useless complication. If needed. just concatenate.

coreblocks/params/layouts.py
tilk: I wouldn't extend the RS structure, I would use a new one and avoid concatenation entriely.

coreblocks/lsu/dummyLsu.py
```diff
@@ -227,6 +238,10 @@ def _():
         def _(rs_data: Record, rs_entry_id: Value):
             m.d.sync += assign(current_instr, rs_data)
             m.d.sync += valid.eq(1)
+            with Transaction().body(m):
+                addr = calculate_addr(m, self.gen_params.isa.xlen, current_instr)
+                mmio_flag = pma.ask(m, addr=addr)["mmio"]
```
tilk, line 243: Would keep the flags in the record, and use the fields when needed.

coreblocks/lsu/dummyLsu.py
tilk: A better solution would be to not use the transactional interface in PMA checking. There is no reason to use it, this is a simple combinational circuit. Also, there is a disadvantage for doing so: if the PMA checker is required in more than one place, then the PMA methods will be blocked by competing transactions. It's better to just instantiate the whole circuit again.

kindlermikolaj: how about something like this:
`with Transaction().body(m, request=~(current_instr.rp_s1 | current_instr.rp_s2)): 
   ...
  mmio_flag = pma.ask(m, addr=addr)["mmio"]
    ...
`
And then similar but without request condition in insert? That way we call it within method & we call it only once per instruction

coreblocks/params/layouts.py
kindlermikolaj: The rationale behind `layout_superset` is that there is currently no instance of list concatenation in layouts, but there is `layout_subset` so having similar function that does the opposite makes sense (at least to me).

coreblocks/params/configurations.py
kindlermikolaj: perhaps `Iterable` would be more adequate?

coreblocks/lsu/dummyLsu.py
tilk: Your suggestion is basically @piotro888's number 1. We can discuss this during the meeting, but I now believe that the transactional interface is introducing unnecessary complexity here while solving no issues at all.

coreblocks/params/configurations.py
tilk: No, `Iterable` is fine for function interfaces. Data structures should be concrete.

coreblocks/lsu/dummyLsu.py
piotro888: I also think now that instantiating a module is better choice here. Transactional interface is not worth the introduced complexity

coreblocks/params/configurations.py
kindlermikolaj: ok, done

coreblocks/lsu/dummyLsu.py
```diff
@@ -217,6 +219,16 @@ def elaborate(self, platform):
...
 
+        addr = Signal(self.gen_params.isa.xlen)
+        m.d.comb += addr.eq(current_instr.s1_val + current_instr.imm)
+
+        pmas = Record(self.gen_params.get(PMALayouts).pma_attrs_layout)
+        m.d.comb += pma_checker.addr.eq(addr)
+        m.d.comb += pmas.eq(pma_checker.result)
+
+        is_mmio = Signal()
+        m.d.comb += is_mmio.eq(pmas["mmio"])
```
tilk, line 230: Combinational `Signal` has two main roles:

* Being used together with Amaranth's control structures (`m.If` etc.)
* De-duplicating expressions (as Amaranth works on trees instead of DAGs, unfortunately).

This signal is used for none of the above purposes, so a regular Python variable is enough:

```suggestion
        is_mmio = pmas["mmio"]
```

But actually, the RHS is so short and descriptive I would just consider inlining the definition at use site.

coreblocks/lsu/dummyLsu.py
```diff
@@ -217,6 +219,16 @@ def elaborate(self, platform):
         instr_is_load = Signal()
         m.d.comb += instr_is_load.eq(current_instr.exec_fn.op_type == OpType.LOAD)
 
+        addr = Signal(self.gen_params.isa.xlen)
+        m.d.comb += addr.eq(current_instr.s1_val + current_instr.imm)
+
+        pmas = Record(self.gen_params.get(PMALayouts).pma_attrs_layout)
+        m.d.comb += pma_checker.addr.eq(addr)
+        m.d.comb += pmas.eq(pma_checker.result)
```
tilk, line 227: The attribute `pma_checker.result` is already a `Record`, you gain nothing by this. Consider just:

```suggestion
        m.d.comb += pma_checker.addr.eq(addr)
        pmas = pma_checker.result
```

Also, in case when creating a new Record is actually needed, check out `Record.like`.

coreblocks/lsu/pma.py
```diff
@@ -0,0 +1,77 @@
...
+    Attributes
+    ----------
+
+    addr : Signal
+        Memory address, for which PMAs are requested.
+
+    result : Record
+        PMAs for given address.
+
+    """
```
tilk, line 46: There are some superfluous newlines here.

```suggestion
    """
    Implementation of physical memory attributes checker. It may or may not be a part of LSU.
    `PMAChecker` exposes method `ask` that can be used to request information about given memory
    address. Result of `ask` method contains bit vector of all defined PMAs.

    Attributes
    ----------
    addr : Signal
        Memory address, for which PMAs are requested.
    result : Record
        PMAs for given address.
    """
```

coreblocks/lsu/pma.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def __init__(self, gen_params: GenParams) -> None:
+        super().__init__()
+
```
tilk, line 50: Elaboratable does not have a constructor. This is not needed, and even [Amaranth's docs](https://amaranth-lang.org/docs/amaranth/latest/start.html) show this.

coreblocks/lsu/pma.py
```diff
@@ -0,0 +1,77 @@
...
+        outputs = Array([Record(self.attr_layout) for _ in self.segments])
+
+        # zero output if addr not in region, propagate value if addr in region
+        for i, segment in enumerate(self.segments):
+            start = segment.start
+            end = segment.end
+
+            # check if addr in region
+            with m.If((self.addr >= start).bool() & (self.addr <= end).bool()):
```
tilk, line 68: These `.bool` don't do anything.

```suggestion
            with m.If((self.addr >= start) & (self.addr <= end)):
```

coreblocks/lsu/pma.py
```diff
@@ -0,0 +1,77 @@
...
+                m.d.comb += outputs[i].eq(segment.mmio)
+            with m.Else():
+                m.d.comb += outputs[i].eq(0x0)
```
tilk, line 71: This is not needed, the default is 0.

```suggestion
```

coreblocks/lsu/pma.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # OR all outputs
+        for output in outputs:
+            m.d.comb += self.result.eq(self.result | output)
```
tilk, line 75: A nitpick - could be written as:

```suggestion
        m.d.comb += self.result.eq(reduce(or_, outputs))
```

I know that Pythonists prefer iteration, but [even Guido considers using reduce with associative operators okay](https://www.artima.com/weblogs/viewpost.jsp?thread=98196).

tilk (commented):
Looks like a pretty nice PR, the only major thing missing is a test.

coreblocks/lsu/pma.py
kindlermikolaj: I didn't know that `reduce` can be applied in `Signal.eq()`

coreblocks/lsu/pma.py
```diff
@@ -0,0 +1,70 @@
...
+        # poor man's interval list
+        self.segments = gen_params.pma
+        self.attr_layout = gen_params.get(PMALayouts).pma_attrs_layout
+        self.result = Record(self.attr_layout)
+        self.addr = Signal(gen_params.isa.xlen)
+
+    def elaborate(self, platform) -> HasElaborate:
+        m = TModule()
+
+        outputs = Array([Record(self.attr_layout) for _ in self.segments])
```
lekcyjna123, line 56: This don't have to be `Array`. Please see that in line 65 you index `outputs` with the python integer `i` not with signal. So all iterations will be done on generation time, so there is no need to include selection logic for runtime.
```suggestion
        outputs = [Record(self.attr_layout) for _ in self.segments]
```

test/lsu/test_pma.py
```diff
@@ -0,0 +1,128 @@
...
+
+        m.submodules.exception_report = self.exception_report = TestbenchIO(
+            Adapter(i=self.gen.get(ExceptionRegisterLayouts).report)
+        )
+
+        self.gen.get(DependencyManager).add_dependency(ExceptionReportKey(), self.exception_report.adapter.iface)
+
+        m.submodules.func_unit = func_unit = LSUDummy(self.gen, self.bus)
+
+        m.submodules.select_mock = self.select = TestbenchIO(AdapterTrans(func_unit.select))
```
lekcyjna123, line 67: Why you don't use `SimpleTestCircuit`?

lekcyjna123 (approved):
Logic looks ok. But I would rather see the newer syntax for TestCircuits, not legacy one (can be fixed in separate PR).

PR #511, opened 2023-11-14
DummyLSU refactored again

### LINKED ISSUE #496: Refactor DummyLSU to avoid direct combinational communication
There are potential timing issues related to the use of combinational signals to communicate the DummyLSU with the bus control internals. These can probably be resolved by including the communication in the transaction framework. Probably `Forwarder` will be useful, as it retains the combinational characteristic while playing nice with transactions.

tilk:
I finally cleaned up this ungodly mess of wires which the DummyLSU was. There is now a clear separation of concerns: The `LSURequesterWB` handles Wishbone, and `DummyLSU` handles the LSU logic. Use of a transactional interface removed a lot of weird combinational communication.

Also:

* Disregarding typos, this worked on the first try.
* The IPC performance improved by another 7%.
* Separating Wishbone logic paves the way for adding a data cache.

Closes #496.


coreblocks/lsu/dummyLsu.py
```diff
@@ -241,26 +236,48 @@ def _(reg_id: Value, reg_val: Value):
...
+                store=~instr_is_load,
+            )
+            with m.If(res["exception"]):
+                results.write(m, data=0, exception=res["exception"], cause=res["cause"])
+
+        with Transaction().body(m, request=instr_ready & ~issued & instr_is_fence):
+            m.d.sync += issued.eq(1)
+            results.write(m, data=0, exception=0, cause=0)
+
+        with Transaction().body(m, request=instr_ready):
```
lekcyjna123, line 255: Why `request=instr_ready` is needed? This transaction will be ready only if `requester.accept` will be ready, so first an `requester.issue` has been started which is already guarded by `instr_ready`.

lekcyjna123 (approved):
Now it looks much better. Whats more there are no changes in unit tests, so this is an example that our testing strategy is right.

coreblocks/lsu/dummyLsu.py
tilk: You're right that this check is superfluous: if `accept` was ready and `instr_ready` was false, this would violate an (unspoken) invariant on the `issue`/`accept` interfaces of the requester, so we're in an undefined state and anything can happen.

coreblocks/lsu/dummyLsu.py
```diff
@@ -241,26 +236,48 @@ def _(reg_id: Value, reg_val: Value):
                 m.d.sync += current_instr.s2_val.eq(reg_val)
                 m.d.sync += current_instr.rp_s2.eq(0)
 
-        @def_method(m, self.get_result, result_ready)
+        with Transaction().body(m, request=instr_ready & ~issued & ~instr_is_fence & (execute | instr_is_load)):
```
Kristopher38, line 239: Leaving a comment on why such a `request` condition is necessary would be helpful (I get what it does after reading rest of the code but not at first glance).

coreblocks/lsu/dummyLsu.py
```diff
@@ -241,26 +236,48 @@ def _(reg_id: Value, reg_val: Value):
...
+                m,
+                addr=current_instr.s1_val + current_instr.imm,
+                data=current_instr.s2_val,
+                funct3=current_instr.exec_fn.funct3,
+                store=~instr_is_load,
+            )
+            with m.If(res["exception"]):
+                results.write(m, data=0, exception=res["exception"], cause=res["cause"])
+
+        with Transaction().body(m, request=instr_ready & ~issued & instr_is_fence):
```
Kristopher38, line 251: Same as above, simple "Handles fence as a no-op" comment would be sufficient.

Kristopher38 (changes requested):
This is surprisingly very readable!
May I suggest adding a clear method for use by interrupts, or would you rather see it as a separate PR? Testing it could either wait for when we get to merge `feature/interrupts`, or alternatively tests could be taken right now from here: https://github.com/kuznia-rdzeni/coreblocks/commit/2e1326a13cdef40ba88089d5ffc340fd519af08f

tilk:
Would prefer merging this ASAP and do extensions later.

Kristopher38 (approved):

PR #512, opened 2023-11-15
retirement: Disable side effects on exception (proposed corrections)

tilk:
This is an updated version of #493, created to discuss possible solutions to the problems in that PR. This branch is intended to be merged into #493 by @Arusekk.

Changes introduced:

* New LSU from #511.
* Combinational side_fx signal in retirement, to fix the issue reported by @piotro888.
* FRAT rename calls in retirement moved to a separate transaction, to avoid mutual locking of retirement and renaming during normal core operation. Verified in benchmarks that there is no loss in IPC.

### DIFF
--- a/coreblocks/core.py
+++ b/coreblocks/core.py
@@ -136,6 +136,7 @@ class Core(Elaboratable):
             rf_free=rf.free,
             precommit=self.func_blocks_unifier.get_extra_method(InstructionPrecommitKey()),
             exception_cause_get=self.exception_cause_register.get,
+            frat_rename=frat.rename,
         )
 
         m.submodules.csr_generic = self.csr_generic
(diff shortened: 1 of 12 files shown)


piotro888 (approved):
Looks good to me. LSU and layouts.py changes integrated very nicely here :)
Is there a reason why it is a draft PR?

piotro888:
> FRAT rename calls in retirement moved to a separate transaction, to avoid mutual locking of retirement and renaming during normal core operation. Verified in benchmarks that there is no loss in IPC.

I have a (unrelated to this PR) side question looking at this solution. 
We call `ExceptionCauseRegister.report` method from multiple FUs in that way:
```python
            with m.If(res["exception"]):
                self.report(m, rob_id=current_instr.rob_id, cause=res["cause"])
```
Doesn't it cause conflicts in parent Transactions and delay different FUs step from working at the same time? Or is it the If also guarding the conflict condition? I'm asking only about case where there are no exceptions reported.
If exceptions are reported those calls even should block each other - to guarantee that when instruction with exception reaches retirement, then ExceptionCauseRegister is properly set. (so if this is a problem for normal operation, same solution with Forwarder can't be used there)

tilk:
Yes, the use of `report` here can possibly cause problems when `report` will be used in more places. As this doesn't affect the core too much right now, this issue is left for the future.

lekcyjna123 (approved):

PR #454, opened 2023-07-27
Update synthesis docs

lekcyjna123:
I have found today that our synthesis page in documentation is a little bit out-of-date, so I updated it by adding information how to use our pre-build docker image.

PR #515, opened 2023-11-19
Update dockerfiles

lekcyjna123:
While updating #454 I have found that we don't have venv in dockers, so it have to be manually installed before running the regression tests by hand. Here is a PR which update dockers to contain this package. Additionally I integrated changes added by @piotro888 in #513 to update dockers only once.

(Dockers are being currently uploaded)

### DIFF
--- a/docker/riscv-toolchain.Dockerfile
+++ b/docker/riscv-toolchain.Dockerfile
@@ -3,15 +3,40 @@ FROM ubuntu:23.04
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive \
     apt-get install -y --no-install-recommends \
-    autoconf automake autotools-dev curl python3 bc lsb-release \
+    autoconf automake autotools-dev curl python3.11 python3.11-venv python3-pip bc lsb-release \
     libmpc-dev libmpfr-dev libgmp-dev gawk build-essential \
-    bison flex texinfo gperf libtool patchutils zlib1g-dev \
-    libexpat-dev ninja-build git ca-certificates python-is-python3 && \
+    bison flex texinfo gperf libtool patchutils zlib1g-dev device-tree-compiler \
+    libexpat-dev ninja-build git ca-certificates python-is-python3 \ 
+    libssl-dev libbz2-dev libreadline-dev libsqlite3-dev libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev && \
     rm -rf /var/lib/apt/lists/*
 
-RUN git clone https://github.com/riscv/riscv-gnu-toolchain && \
+RUN git clone --shallow-since=2023.05.01 https://github.com/riscv/riscv-gnu-toolchain && \
     cd riscv-gnu-toolchain && \
     git checkout 2023.05.14 && \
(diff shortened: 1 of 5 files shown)


docker/riscv-toolchain.Dockerfile
```diff
@@ -3,15 +3,40 @@ FROM ubuntu:23.04
...
     DEBIAN_FRONTEND=noninteractive \
     apt-get install -y --no-install-recommends \
-    autoconf automake autotools-dev curl python3 bc lsb-release \
+    autoconf automake autotools-dev curl python3.11 python3.11-venv python3-pip bc lsb-release \
     libmpc-dev libmpfr-dev libgmp-dev gawk build-essential \
-    bison flex texinfo gperf libtool patchutils zlib1g-dev \
-    libexpat-dev ninja-build git ca-certificates python-is-python3 && \
+    bison flex texinfo gperf libtool patchutils zlib1g-dev device-tree-compiler \
+    libexpat-dev ninja-build git ca-certificates python-is-python3 \ 
+    libssl-dev libbz2-dev libreadline-dev libsqlite3-dev libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev && \
```
piotro888, line 10: Regarding your comment https://github.com/kuznia-rdzeni/coreblocks/pull/513#discussion_r1398237648 and this change - I don't think that running `apt-get` only in the first `RUN` is that better. Opposite has its benefits - if I wanted to add another tool like spike, with its dependencies I only created next layer that is entirely dedicated to the tool. I didn't modify previous layers, so I didn't need to rebuild the entire container (cached versions of first two steps were used) or upload first two layers again. I would be for this option, as it is useful for updating and developing containers.
Are there some benefits of installing everything in first step, that I am missing?

lekcyjna123: There is a benefit that you don't call `apt update` multiple times, so you generate lower traffic for Debian servers.

piotro888: But on the other hand, when your are adding new tool with new package dependencies as a new step, instead of installing all of the packages again at rebuild, you use cached previous docker layers and only download the newly added packages, which creates even less traffic.

xThaid: Note that there is `rm -rf /var/lib/apt/lists/*` in the end of the command. This will make the resulting docker image smaller (by noticeable amount), as all the files that we are going to delete will be ephemeral (and won't exist in the resulting image). If we split that into two `RUN` instructions, the contents of `/var/lib/apt/lists/` would be saved as part of the first layer (each instruction is a new layer).

docker/riscv-toolchain.Dockerfile
```diff
@@ -3,15 +3,40 @@ FROM ubuntu:23.04
...
+    cd .. && \
+    rm -rf build/
+
+RUN git clone --depth=1 https://github.com/pyenv/pyenv.git .pyenv && \
+    export PATH=/.pyenv/bin:$PATH && \
+    export PYENV_ROOT=/root/.pyenv && \
+    eval "$(pyenv init --path)" && \
+    pyenv install 3.6.15 && \
+    pyenv global 3.6.15 && \
+    python -V && \
```
piotro888, line 37: please remove this line

piotro888 (approved):
Okay, lets merge it. I will merge it to my PR after this one is on master

Issue #514, opened 2023-11-17
Merge improved BasicFIFO


tilk:
The proof-of-concept vector core contains a slightly improved BasicFIFO implementation. Merging it might improve synthesis performance of the core.

PR #516, opened 2023-11-19
BasicFifo tweaks

lekcyjna123:
As requested in #514, here is a commit to merge my modifications to BasicFifo. 

Changes:
- In Fifo on the end of the previous cycle we now what data will be read in the next cycle (based on the fact if we read something in the previous cycle or not), so we can use "sync" memory instead of "comb".
- There is no need to compare for less/greater, simple difference is enough.
- Minimal change in semantic - `clear` method don't longer conflicts with `read` and `write` so it is possible that we will `read` data while clearing fifo. This caused that a `Settle` was needed in tests.
- use `top_comb` where possible

I have started a synthesis job in:
https://github.com/lekcyjna123/coreblocks/actions/runs/6921193516
But comparing to the current master there is slightly degradation in Fmax (by 1.5 MHz). I am not sure if this is a caused by noise or my commit (in vector core the usage of registers was so big that maybe it was cheaper - from PoV of routing - to use memory).


lekcyjna123:
Another synthesis job:
https://github.com/lekcyjna123/coreblocks/actions/runs/6921508344

transactron/utils/fifo.py
```diff
@@ -78,26 +75,27 @@ def mod_incr(sig: Value, mod: int) -> Value:
...
         def _() -> ValueLike:
-            m.d.sync += self.read_idx.eq(mod_incr(self.read_idx, self.depth))
+            m.d.sync += self.read_idx.eq(next_read_idx)
             return self.head
 
         @def_method(m, self.clear)
         def _() -> None:
             m.d.sync += self.read_idx.eq(0)
             m.d.sync += self.write_idx.eq(0)
+            m.d.sync += self.level.eq(0)
```
piotro888, line 98: Why is level assigned two times for clear method? (first in 
```python
with m.If(self.clear.run):
            m.d.sync += self.level.eq(0)
```
).

Please remove one

piotro888 (approved):
Apart from one issue, LGTM

transactron/utils/fifo.py
```diff
@@ -51,25 +52,21 @@ def __init__(self, layout: MethodLayout, depth: int) -> None:
...
-                return sig + 1
-            return Mux(sig == mod - 1, 0, sig + 1)
-
         m = TModule()
 
-        m.submodules.buff_rdport = self.buff_rdport = self.buff.read_port(
-            domain="comb", transparent=True
-        )  # FWFT behaviour
+        next_read_idx = Signal.like(self.read_idx)
+        m.d.top_comb += next_read_idx.eq(mod_incr(self.read_idx, self.depth))
```
tilk, line 63: No reason to use `top_comb` at toplevel! More noise, no gain.

```suggestion
        m.d.comb += next_read_idx.eq(mod_incr(self.read_idx, self.depth))
```

tilk:
Surprisingly, the change in performance numbers doesn't look significant. In particular, there is no change in the number of RAM blocks synthesized.

transactron/utils/fifo.py
```diff
@@ -51,25 +52,21 @@ def __init__(self, layout: MethodLayout, depth: int) -> None:
...
+
+        m.submodules.buff_rdport = self.buff_rdport = self.buff.read_port(domain="sync", transparent=True)
         m.submodules.buff_wrport = self.buff_wrport = self.buff.write_port()
 
-        m.d.comb += self.read_ready.eq(self.level > 0)
-        m.d.comb += self.write_ready.eq(self.level < self.depth)
+        m.d.top_comb += self.read_ready.eq(self.level != 0)
+        m.d.top_comb += self.write_ready.eq(self.level != self.depth)
```
tilk, line 69:
```suggestion
        m.d.comb += self.read_ready.eq(self.level != 0)
        m.d.comb += self.write_ready.eq(self.level != self.depth)
```

transactron/utils/fifo.py
```diff
@@ -78,20 +75,20 @@ def mod_incr(sig: Value, mod: int) -> Value:
         with m.If(self.clear.run):
             m.d.sync += self.level.eq(0)
 
-        m.d.comb += self.buff_rdport.addr.eq(self.read_idx)
-        m.d.comb += self.head.eq(self.buff_rdport.data)
+        m.d.top_comb += self.buff_rdport.addr.eq(Mux(self.read.run, next_read_idx, self.read_idx))
+        m.d.top_comb += self.head.eq(self.buff_rdport.data)
```
tilk, line 79:
```suggestion
        m.d.comb += self.buff_rdport.addr.eq(Mux(self.read.run, next_read_idx, self.read_idx))
        m.d.comb += self.head.eq(self.buff_rdport.data)
```

tilk (approved):

PR #522, opened 2023-11-24
Run linker on internal asm tests

piotro888:
Linker run was somehow missing for our internal assembly tests.
I created basic linker flow (with slightly modified script from our `riscv-tests` env).

It is needed to run asm test that I written for testing exceptions.

`la` (load address) instruction is example that doesn't work without linking.
`la` translates to:
before:
```
10:   00000317                auipc   t1,0x0
14:   00030313                mv      t1,t1
```
after:
```
10000010:       00000317                auipc   t1,0x0
10000014:       02430313                addi    t1,t1,36 # 10000034 <exception_handler>
```

### DIFF
--- a/test/asm/link.ld
+++ b/test/asm/link.ld
@@ -1,0 +1,26 @@
+OUTPUT_ARCH( "riscv" )
+
+start = 0x0;
+
+MEMORY
+{
+    data (wxa!ri) : ORIGIN = 0x00000000, LENGTH = 1K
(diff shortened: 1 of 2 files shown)


tilk:
If I understand things correctly, the data and BSS segments are currently not used, as is the memory map. I believe that the linker script can be simplified, so that it does not suggest that these features work.

piotro888:
I updated it to minimal that makes sense for me. I also kept the .bss support (but removed .data), because it is already supported in test hardware (just uninitialized memory space) and may be useful for writing tests.

PR #523, opened 2023-11-24
Exceptions handling

piotro888:
Full exception handling implementation based on @Arusekk method :tada: 

Implements instruction counting, fetch stalling/resuming, {`mepc`, `mtvec`} CSR support, extension to `ExceptionCauseRegister`
Tested on asm test.

Depends on #522

### DIFF
--- a/coreblocks/core.py
+++ b/coreblocks/core.py
@@ -136,6 +136,7 @@ class Core(Elaboratable):
             rf_free=rf.free,
             precommit=self.func_blocks_unifier.get_extra_method(InstructionPrecommitKey()),
             exception_cause_get=self.exception_cause_register.get,
+            frat_rename=frat.rename,
         )
 
         m.submodules.csr_generic = self.csr_generic
--- a/coreblocks/core.py
+++ b/coreblocks/core.py
@@ -147,7 +147,7 @@ class Core(Elaboratable):
             exception_cause_get=self.exception_cause_register.get,
             frat_rename=frat.rename,
             fetch_continue=self.fetch.verify_branch,
-            fetch_stop=self.fetch.stop,
+            fetch_stall=self.fetch.stall_exception,
             instr_decrement=self.core_counter.decrement,
         )
 
--- a/coreblocks/core.py
+++ b/coreblocks/core.py
@@ -2,6 +2,7 @@ from amaranth import *
 
 from coreblocks.params.dependencies import DependencyManager
 from coreblocks.stages.func_blocks_unifier import FuncBlocksUnifier
+from coreblocks.structs_common.instr_counter import CoreInstructionCounter
 from transactron.core import Transaction, TModule
 from transactron.lib import FIFO, ConnectTrans
 from coreblocks.params.layouts import *
@@ -20,6 +21,7 @@ from coreblocks.stages.retirement import Retirement
 from coreblocks.frontend.icache import ICache, SimpleWBCacheRefiller, ICacheBypass
 from coreblocks.peripherals.wishbone import WishboneMaster, WishboneBus
 from coreblocks.frontend.fetch import Fetch, UnalignedFetch
+from transactron.lib.transformers import MethodProduct
 from transactron.utils.fifo import BasicFifo
 
 __all__ = ["Core"]
@@ -41,6 +43,8 @@ class Core(Elaboratable):
             self.gen_params.get(SchedulerLayouts).free_rf_layout, 2**self.gen_params.phys_regs_bits
         )
 
+        self.core_counter = CoreInstructionCounter(self.gen_params)
+
         cache_layouts = self.gen_params.get(ICacheLayouts)
         if gen_params.icache_params.enable:
             self.icache_refiller = SimpleWBCacheRefiller(
@@ -50,10 +54,12 @@ class Core(Elaboratable):
         else:
             self.icache = ICacheBypass(cache_layouts, gen_params.icache_params, self.wb_master_instr)
 
+        self.fetch_continue = MethodProduct([self.fifo_fetch.write, self.core_counter.increment])
+
         if Extension.C in gen_params.isa.extensions:
-            self.fetch = UnalignedFetch(self.gen_params, self.icache, self.fifo_fetch.write)
+            self.fetch = UnalignedFetch(self.gen_params, self.icache, self.fetch_continue.method)
         else:
-            self.fetch = Fetch(self.gen_params, self.icache, self.fifo_fetch.write)
+            self.fetch = Fetch(self.gen_params, self.icache, self.fetch_continue.method)
 
         self.FRAT = FRAT(gen_params=self.gen_params)
         self.RRAT = RRAT(gen_params=self.gen_params)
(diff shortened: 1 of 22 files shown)


coreblocks/structs_common/exception.py
```diff
@@ -61,7 +70,7 @@ def elaborate(self, platform):
         m = TModule()
 
         @def_method(m, self.report)
-        def _(cause, rob_id):
+        def _(cause, rob_id, pc):
```
Arusekk, line 73: Why set the `pc` here, though? I think it would have been better (to discuss) if we set it to the previous `pc` right when jumping to the handler, and not in (I guess) every speculative instruction reporting exceptions.

coreblocks/frontend/fetch.py
```diff
@@ -44,16 +45,23 @@ def elaborate(self, platform):
         speculative_pc = Signal(self.gp.isa.xlen, reset=self.gp.start_pc)
 
         stalled = Signal()
+        stalled_unsafe = Signal()
```
Arusekk, line 48: Ideally to be simplified out in future PRs, but I get it, still necessary :)

coreblocks/frontend/fetch.py
```diff
@@ -213,9 +231,16 @@ def elaborate(self, platform) -> TModule:
                 self.cont(m, instr=instr, pc=current_pc, access_fault=cache_resp.error, rvc=is_rvc)
 
         @def_method(m, self.verify_branch, ready=(stalled & ~flushing))
-        def _(from_pc: Value, next_pc: Value):
+        def _(from_pc: Value, next_pc: Value, resume_from_exception: Value):
```
Arusekk, line 234: Hope that it works as intended, in my head verify_branch was always slightly misused and I was never quite able to put my finger on it.

test/asm/link.ld
```diff
@@ -0,0 +1,26 @@
...
+PHDRS
+{
+  text PT_LOAD;
+  data_init PT_LOAD;
+  data PT_NULL;
+}
+
+SECTIONS
+{
+  .text.init : { *(.text.init) } >text AT>text :text
```
Arusekk, line 20: Is `.init` necessary? Is the ld script from somewhere or just hand-crafted? It slowly starts to look like a platform description 👀

test/structs_common/test_exception.py
```diff
@@ -75,7 +75,8 @@ def process_test():
 
                 cause = random.choice(list(ExceptionCause))
                 report_rob = random.randint(0, self.rob_max)
-                report_arg = {"cause": cause, "rob_id": report_rob}
+                report_pc = random.randint(0, 2**self.gp.isa.xlen - 1)
```
Arusekk, line 78: My favourite nitpick :)
```suggestion
                report_pc = random.randrange(2**self.gp.isa.xlen)
```

coreblocks/core.py
```diff
@@ -50,10 +54,12 @@ def __init__(self, *, gen_params: GenParams, wb_instr_bus: WishboneBus, wb_data_
         else:
             self.icache = ICacheBypass(cache_layouts, gen_params.icache_params, self.wb_master_instr)
 
+        self.fetch_continue = MethodProduct([self.fifo_fetch.write, self.core_counter.increment])
```
tilk, line 57: Note to self: we need a simple interface to the method transformers.

coreblocks/structs_common/exception.py
piotro888: This is true, `mepc` is actually set in the Retirement [here](https://github.com/kuznia-rdzeni/coreblocks/blob/6990d5f9656ee503a3072b0bf2f4c40b4ecb033a/coreblocks/stages/retirement.py#L82). This field is used to forward (and accumulate) pc field to retirement, because it is not accessible there [in ROB] (just like exception cause).

coreblocks/structs_common/instr_counter.py
```diff
@@ -0,0 +1,36 @@
...
+        m = TModule()
+
+        with m.If(self.increment.run & ~self.decrement.run):
+            m.d.sync += self.count.eq(self.count + 1)
+
+        with m.If(self.decrement.run & ~self.increment.run):
+            m.d.sync += self.count.eq(self.count - 1)
+
+        @def_method(m, self.increment)
+        def _(instr, rvc, access_fault, pc):
```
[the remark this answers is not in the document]
piotro888: It is specified to use `MethodProduct` cleanly, so this is some tradeoff of clarity. Maybe we could add some option to `MethodProduct` to connect only common specified arguments?

coreblocks/frontend/fetch.py
piotro888: Yes, it is intended to cleanly eliminate `stalled_unsafe` in the future

coreblocks/frontend/fetch.py
piotro888: It should, because it provides the same interface and logic for previous use cases, and adds additional change for new the new one. When we are stalled because of exception, all effects from "normal" calls are ignored, because `stalled` is still true and other paremeters will be overwritten by later call (and `flushing` logic is aware of currently processed instruction count).

For now I think it is the best way to avoid code and logic duplication, and also has benefits in jump prediction - when flushing, we still can save taken jump paths and ignore the verify call jumping to handler for prediction.

coreblocks/structs_common/instr_counter.py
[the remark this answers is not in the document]
tilk: Maybe so. Another way would be to use a `MethodTransformer` to clear the arguments. This would become pretty verbose, but I just added #525 to make this kind of use less verbose.

test/asm/link.ld
piotro888: Addressed in #522

coreblocks/structs_common/instr_counter.py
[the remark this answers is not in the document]
piotro888: changed to used the `MethodMap` solution to drop arguments

tilk (approved):
No tests, but we agreed to leave that to the moment when the microarchitecture stabilizes again.

coreblocks/structs_common/instr_counter.py
```diff
@@ -0,0 +1,46 @@
...
+        with m.If(self.decrement.run & ~self.increment.run):
+            m.d.sync += count.eq(count - 1)
+
+        @def_method(m, self.increment)
+        def _():
+            pass
+
+        @def_method(m, self.decrement)
+        def _():
+            return count == 1 & ~self.increment.run
```
lekcyjna123, line 44: Wrong operator precedence. You have written `count == (1 & ~self.increment.run)`.

```suggestion
            return (count == 1) & ~self.increment.run
```

coreblocks/structs_common/instr_counter.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, gp: GenParams):
+        self.gp = gp
+
+        self.increment = Method()
+        self.decrement = Method(o=gp.get(CoreInstructionCounterLayouts).decrement)
+
+    def elaborate(self, platform) -> TModule:
+        m = TModule()
+
+        count = Signal(self.gp.rob_entries_bits + 1)
```
lekcyjna123, line 30: What if someone will create a rob with 1 entry? Then we will have ~4 instructions before ROB, which will cause that the counter will overrun. (I know, this is a very weird edge case :D)

```suggestion
        # Make sure that we have enough bits in counter to count all instructions before ROB
        count = Signal(max(4,self.gp.rob_entries_bits + 1))
```

piotro888: done

lekcyjna123 (approved):

PR #525, opened 2023-11-27
Simpler way of using method transformations

tilk:
This PR allows the use of `MethodTransformer` etc. to be less wordy.


transactron/lib/transformers.py
```diff
@@ -17,7 +18,33 @@
...
+        ----------
+        m: Module or TModule
+            The module to which this combiner is added as a submodule.
+        """
+        m.submodules += self
+        return self.method
+
+
+class MethodMap(Combiner, Elaboratable):
     """Method transformer.
```
piotro888, line 48: Update docstring

piotro888 (approved):
Nice!

PR #533, opened 2023-12-03
Fix riscv-test job dependencies

lekcyjna123:
I didn't want to block #513, so the issue which I found during review is corrected here. In #513 generation of verilog was moved to separate job. `riscv-tests` use artifact of that job, but the dependencies weren't updated.


tilk:
While you're at it, you could also correct the spelling of "Synthethise full core". ;)

piotro888 (approved):
Oh, it must have disappeared during resolving merge conflict, because it was committed

tilk (approved):

PR #534, opened 2023-12-06
Clear ExcepionCauseRegister after flushing core

piotro888:
Oops.
It worked on test because exceptions were close together, and `rob_id` didn't do the full overflow cycle to break the instruction age comparison


tilk (commented):
Bugs were to be expected, as the testing was minimal.

Is it possible, in a relatively short time, to extend the assembly test so that this issue is catched by it?

piotro888:
> Is it possible, in a relatively short time, to extend the assembly test so that this issue is catched by it?

Done, it turns out that the bug was even more specific - because of constatly changing rob tail id - in this case it only was broken when new exception had exactly the same rob idx as the previous one.

Very good that it was found now and not during debugging OS run :)

tilk (approved):

PR #537, opened 2023-12-15
Refactor runnable signal of transactions

tilk:
This PR refactors the `runnable` signal from transaction schedulers into the transaction manager and exposes the signal in the `Transaction` class. This is a prerequisite for implementing a transaction profiler.


lekcyjna123 (approved):

tilk:
I'm bypass-merging because this is a simple change which is unlikely to have problems, and there is no value gained from another review.

PR #538, opened 2023-12-15
Assertions in hardware

tilk:
This PR adds a function named `assertion` which allows to define hardware-level ("run-time") assertions. Failed assertion will fail a test.

TODO:
* [x] Write a test
* [x] Add docstrings
* [x] Handle assertions in regression tests too (pysim engine)
* [x] Assertions with source locations (to help easily find the failed assertion)
* [x] Make independent from Coreblocks (requires #544)


lekcyjna123:
Maybe move the `gp` to `gen_params` and other naming changes to separate PR?

tilk:
> Maybe move the `gp` to `gen_params` and other naming changes to separate PR?

These are crucial for this to work, but you're right, this PR would be easier to review this way.

test/common/coreblocks.py
```diff
@@ -0,0 +1,34 @@
...
+__all__ = ["CoreblocksTestCaseWithSimulator", "make_assert_handler"]
+
+
+def make_assert_handler(gen_params: GenParams, my_assert: Callable[[int, str], Any]):
+    def assert_handler():
+        yield Passive()
+        while True:
+            if (yield assert_bit(gen_params)):
+                for v, (n, i) in assert_bits(gen_params):
+                    my_assert((yield v), f"Assertion at {n}:{i}")
```
lekcyjna123, line 20: Is it possible, that the assertion will be False? If any of `v` is False, then `assert_bit` is False, so we don't take the `if` and we will skip the condition checking.

lekcyjna123 (commented):
Now I see that we should probably move `DependencyManager` to transactron, as assertions aren't something coreblocks specific and they need the manager.

test/common/coreblocks.py
tilk: There is a negation missing in the condition. (Why didn't you say so?)

tilk:
> Now I see that we should probably move `DependencyManager` to transactron, as assertions aren't something coreblocks specific and they need the manager.

No, they're not, but they should be syntactically simple to use, and currently the easiest way to do so is to use GenParams.

test/common/coreblocks.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+__all__ = ["CoreblocksTestCaseWithSimulator", "make_assert_handler"]
+
+
+def make_assert_handler(gen_params: GenParams, my_assert: Callable[[int, str], Any]):
+    def assert_handler():
+        yield Passive()
+        while True:
+            yield Delay(5e-7)  # Shorter than clock cycle
```
lekcyjna123, line 17: Take length of clock cycle as an argument and calculate `Delay` by removing epsilon from it.

How this handler should be ordered regarding to TransactionProfiles? I think that order of this two things shouldn't matter. I am right?

test/common/coreblocks.py
```diff
@@ -0,0 +1,34 @@
...
+            if not (yield assert_bit(gen_params)):
+                for v, (n, i) in assert_bits(gen_params):
+                    my_assert((yield v), f"Assertion at {n}:{i}")
+            yield
+
+    return assert_handler
+
+
+class CoreblocksTestCaseWithSimulator(TestCaseWithSimulator):
```
lekcyjna123, line 26: After moving DependencyManager to transactron it is not needed to have this handler in coreblocks specific file. We can made it part of `TestCaseWithSimulator`.

test/utils/test_assertion.py
lekcyjna123, line 1: I would propose to put it into `test/transactron/utils/test_assertion.py`

test/common/coreblocks.py
tilk: But then the field containing the `DependencyManager` should no longer be named `gen_params`, which will provoke an awkward change in all the Coreblocks tests.

lekcyjna123 (approved):
Overall it looks great. Only two small comments left.

transactron/utils/assertion.py
```diff
@@ -0,0 +1,70 @@
...
+__all__ = ["AssertKey", "assertion", "assert_bit", "assert_bits"]
+
+
+@dataclass(frozen=True)
+class AssertKey(ListKey[tuple[Value, SrcLoc]]):
+    pass
+
+
+def assertion(value: Value, *, src_loc_at: int = 0):
+    """Signal if assertion false.
```
piotro888, line 19: This title may be misleading - if signal/condition is true then assertion is held, not false.

transactron/utils/assertion.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Parameters
+    ----------
+    dependencies : DependentCache | DependencyManager
+        The assertion feature uses the `DependencyManager` to store
+        assertions. If `DependentCache` is passed instead, `DependencyManager`
+        is requested from the cache.
+    """
+    if isinstance(dependencies, DependentCache):
+        dependencies = dependencies.get(DependencyManager)
```
piotro888, line 53: Is passing `DependentCache` useful? `assertion` always adds to `DependencyContext`

piotro888 (approved):
LGTM

transactron/utils/assertion.py
tilk: It was a leftover from previous iterations.

transactron/utils/assertion.py
tilk: Rewrote the docstring a bit.

PR #539, opened 2023-12-15
Method brancher

lekcyjna123:
As requested by @piotro888 here is the port of `MethodBrancher` from #395. I had to write a test for the Brancher, so I have decided to port also `MethodMock`, which simplify syntax. Instead of creating low level Adapter and TestbenchIO after that (and a long chain of accessors to get a real method), we can create a MethodMock which presents high level abstraction.

### DIFF
--- a/test/common/sugar.py
+++ b/test/common/sugar.py
@@ -2,10 +2,29 @@ import functools
 from typing import Callable, Any, Optional
 from .testbenchio import TestbenchIO, TestGen
 from transactron.utils._typing import RecordIntDict
+from transactron._utils import MethodLayout
+from transactron.lib.adapters import Adapter
+from amaranth import *
+
+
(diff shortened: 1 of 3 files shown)


test/common/sugar.py
```diff
@@ -2,10 +2,29 @@
 from typing import Callable, Any, Optional
 from .testbenchio import TestbenchIO, TestGen
 from transactron.utils._typing import RecordIntDict
+from transactron._utils import MethodLayout
+from transactron.lib.adapters import Adapter
+from amaranth import *
+
+
+class MethodMock(Elaboratable):
```
tilk, line 10: This functionality should be moved to another PR, and all method mocks should be updated to use this.

tilk:
After merging #551, this is no longer needed.

PR #540, opened 2023-12-15
Simplify adding mock processes

lekcyjna123:
When writing tests with mock processes I have regularly the problem that I forgot to add such processes to simulation via `sim.add_sync_process`. This cause weird issues and burns time to find the bug. Because of that I would like to propose a solution which will be adding such processes automatically to simulation.

It is done using:
- AutoRegisterMockMetaclass which collects all mock processes in a list
- def_method_mock which marks functions as mock process
- TestCaseWithSimulator which adds automaticaly the processes from list to simulation

Sadly there are some cons of such solution:
- Typing is not obvious (I haven't checked yet if there are some materials in internet about typing with metaclasses)
- This works only if the method mock is defined on the class level (it was the reason for refactor of `test/structs_common/test_exception.py`)

But when the mocks are defined on the class level then using this solution is strightforward (see `test/transactions/test_transaction_lib.py`).


tilk:
Metaclasses generally smell, as they don't compose well. Can't this be handled better by `__init_subclass__`?

lekcyjna123:
No. Because I want to have consistency between `def_method` and `def_method_mock`. In normal `Elaboratables` we allow to have function names all without meaning (e.g. `_`):

```py
@def_method(m, method1)
def _(arg):
   ...
@def_method(m, method2)
def _(arg):
   ...
```

But if we try to do the same with method mocks implemented by `__init_subclass__` or class decorators, then each method mock function have to have unique name (otherwise it will be overwritten). Using metaclasses we allow to overwriting the functions because they are captured in the moment of the function creation and not after creation of all functions.

tilk:
> No. Because I want to have consistency between `def_method` and `def_method_mock`. In normal `Elaboratables` we allow to have function names without meaning (e.g. `_`):

But that's because in `def_method`, the method is already named. In mocks, this is not the case. Also, we don't have consistency here already, as `def_method` can't be used at class level.

I generally dislike metaclass solutions, as they tend to obfuscate things and can make refactoring and redesigns harder. But things making people waste less time are great, so if this can't be done any other way, then I'll live with this.

Is it sensible to apply the metaclass to `TestCaseWithSimulator` directly, instead of its subclasses?  You can forget applying the metaclass as easily as adding mock processes, also exposing metaclass attribute to end users is bad style.

lekcyjna123:
> Is it sensible to apply the metaclass to TestCaseWithSimulator directly, instead of its subclasses? You can forget applying the metaclass as easily as adding mock processes, also exposing metaclass attribute to end users is bad style.

Yes. That is my goal. I wanted to get first the acceptance for the idea, before I will do the refactor of all the tests which use `def_method_mock`.

lekcyjna123:
TODO: Refactor old tests to use new solution.

lekcyjna123:
I have thought one more time about the idea and I decided to drop metaclasses implementation and use instead of that introspection. This caused that mocks processes can not have `_` name, but it simplifies a lot and as @tilk stated:
> But that's because in def_method, the method is already named. In mocks, this is not the case. 

I added also support for auto adding mocks which are defined in function scope. The only situation which left to be handled manually is usage of generators (e.g. `test/scheduler/test_scheduler.py`) but this is fine for me.

While removing manual mock addition to simulator I have found a lot of tests which are inconsistent with our current way of writing tests, so I did refactor where it was needed and justified.

tilk:
You changed local mocks to class-level in some tests. Was it necessary? Can the `add_local_mocks` function you wrote handle this?

lekcyjna123:
> You changed local mocks to class-level in some tests. Was it necessary? Can the add_local_mocks function you wrote handle this?

I have first changed that and after that I implemented the `add_local_mocks`. I have thought about reverting the changes to these tests, but I have decided to leave them as is, because this makes (in my opinion) tests more consistent as class methods are now the prefered method (or I am wrong?).

tilk:
> because this makes (in my opinion) tests more consistent as class methods are now the prefered method (or I am wrong?).

I'm still conflicted about this, but this seems to be just a matter of style, and I agree that consistency is valuable.

tilk (approved):

test/common/infrastructure.py
```diff
@@ -39,7 +40,10 @@ def __init__(self, dut: _T_HasElaborate):
         self._io: dict[str, _T_nested_collection[TestbenchIO]] = {}
 
     def __getattr__(self, name: str) -> Any:
-        return self._io[name]
+        try:
+            return self._io[name]
+        except KeyError:
+            raise AttributeError()
```
piotro888, line 46: This is already way better from previous errors, but maybe some message would be useful? Like `No mock for '{name}'` or something better

piotro888:
I'm also not sure about class level tests, but I think it makes sense to do that - If we have multiple tests in class, class-level processes could be used, `test_` methods could override `self.` variables that would control processes or if completely different behaviour for some process is needed it could be declared locally for each `test_` and rest of mocks automatically shared.

piotro888 (approved):

test/common/infrastructure.py
tilk: I'll wait with merging until this minor change is made.

test/transactions/test_transaction_lib.py
```diff
@@ -569,75 +564,74 @@ def method_process():
...
+        self.serialized_data = deque()
+        self.port_data = [deque() for _ in range(self.port_count)]
 
-        got_request = False
+        self.got_request = False
 
-        def random_wait(rand: int):
-            yield from self.tick(random.randrange(rand) + 1)
+    def random_wait(self, rand: int):
+        yield from self.tick(random.randrange(rand) + 1)
```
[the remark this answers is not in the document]
lekcyjna123: To be addressed in different PR. #564 have been created to track that point

PR #542, opened 2023-12-16
Source locations for transactions and methods

tilk:
Storing source locations for transactions and methods allows to disambiguate them in Transactron's debugging output.

TODO:
- [x] Choose adequate `src_loc_at` for Transactron library and other reusable Elaboratables.

### DIFF
--- a/transactron/lib/simultaneous.py
+++ b/transactron/lib/simultaneous.py
@@ -1,6 +1,6 @@
 from amaranth import *
 
-from ..utils import get_src_loc
+from ..utils import get_src_loc, SrcLoc
 from ..core import *
 from ..core import TransactionBase
 from contextlib import contextmanager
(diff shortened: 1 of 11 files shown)


transactron/lib/simultaneous.py
```diff
@@ -59,6 +61,7 @@ def condition(m: TModule, *, nonblocking: bool = False, priority: bool = True):
     this = TransactionBase.get()
     transactions = list[Transaction]()
     last = False
+    src_loc = get_src_loc(0)
```
piotro888, line 64: Default here should be `1`. Depth of `0` is `/usr/lib/python3.11/contextlib.py`.

Also if all blocks allow providing custom `src_loc`, why not add it also here?

piotro888 (changes requested):
I found one issue with depth, otherwise LGTM

piotro888 (approved):

transactron/lib/simultaneous.py
tilk: Did another change, now the source locations come from the `branch` calls, which allows to identify them.

lekcyjna123 (approved):

PR #543, opened 2023-12-17
Icache code split

kindlermikolaj:
This PR extracts code that is useful for implementation of data cache into separate python module `coreblocks.cache` with interfaces moved to `coreblocks.cache.iface` and `SimpleWBCacheRefiller` move to `coreblocks.cache.refiller`.


test/frontend/test_icache.py
lekcyjna123, line 1: Maybe you should move the Refiller test to the separate file? Similar as with production code?

coreblocks/frontend/icache.py
piotro888, line 1: Please move to cache/

coreblocks/cache/__init__.py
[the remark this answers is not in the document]
tilk: Or maybe not? Caches can be their own thing. Grouping both the LSU stuff and the frontend stuff in a single directory is probably too much.

test/frontend/test_icache.py
piotro888, line 1: test should be moved to test/cache to keep up with main directory structure

coreblocks/cache/refiller.py
```diff
@@ -0,0 +1,75 @@
...
+from amaranth import Cat, Elaboratable, Repl, Signal
+from amaranth.utils import log2_int
+
+from .iface import CacheRefillerInterface
+
+__all__ = ["SimpleWBCacheRefiller"]
+
+
+class SimpleWBCacheRefiller(Elaboratable, CacheRefillerInterface):
+    def __init__(self, layouts: ICacheLayouts, params: ICacheParameters, wb_master: WishboneMaster):
```
piotro888, line 16: Note for future PR: Cache refiller is now generalized, start_refill and acccept_refill method layouts should be moved out of ICacheLayouts to separate class. ICacheParameters also look like general for all caches?

piotro888 (approved):

test/frontend/test_icache.py
lekcyjna123: @kindlermikolaj could you apply this comment?

xThaid:
I am not convinced about making a generic cache class that will fit all. A generic cache for the data cache and the lower level caches - yes. The data cache and the instruction cache - not sure. 

The key differences are that the core itself never writes to the instruction cache, so it doesn't have to care about write policies and so on. Also, I will be soon working on the branch prediction and as a part of that work, I will want to make the I$ return two lines at a time. If the data cache needs to fetch multiple lines - again, not sure.

Another problem is that the instruction cache/data cache may have some specific quirks implemented to work smoothly with IFU/LSU and maintaining a generic version for both can be burdensome.

test/frontend/test_icache.py
kindlermikolaj: done

lekcyjna123 (approved):

lekcyjna123:
In my opinion it is good that we will have a common cache interface, because this will allow for interchangeability. There wouldn't be needed any tweaks to make the same implementation work once as ICache and next as DCache. This will allow us for easy changing the building blocks.

> the core itself never writes to the instruction cache, so it doesn't have to care about write policies and so on

But the core never should care in my opinion about write policies. This is something which cache should handle autonomously.

>  I will want to make the I$ return two lines at a time

Please see, that current interface define only methods, not they layouts. So it should be possible to use the same interface but to return different number of lines depending on the implementation.

> Another problem is that the instruction cache/data cache may have some specific quirks implemented to work smoothly with IFU/LSU and maintaining a generic version for both can be burdensome.

Can you point already some example?

PR #545, opened 2023-12-18
Fix performance problems after #532

piotro888:
#532 introduced loss in Fmax and IPC.
Both problems are fixed and explained in comments.
For other details see discussion at end of comments in #532

### DIFF
--- a/coreblocks/stages/retirement.py
+++ b/coreblocks/stages/retirement.py
@@ -106,51 +106,50 @@ class Retirement(Elaboratable):
 
                     commit = Signal()
 
-                    with condition(m, priority=False) as cond:
-                        with cond(rob_entry.exception):
-                            self.fetch_stall(m)
+                    with m.If(rob_entry.exception):
+                        self.fetch_stall(m)
 
-                            cause_register = self.exception_cause_get(m)
+                        cause_register = self.exception_cause_get(m)
 
-                            cause_entry = Signal(self.gen_params.isa.xlen)
+                        cause_entry = Signal(self.gen_params.isa.xlen)
 
-                            with m.If(cause_register.cause == ExceptionCause._COREBLOCKS_ASYNC_INTERRUPT):
-                                # Async interrupts are inserted only by JumpBranchUnit and conditionally by MRET and CSR
-                                # The PC field is set to address of instruction to resume from interrupt (e.g. for jumps
-                                # it is a jump result).
-                                # Instruction that reported interrupt is the last one that is commited.
-                                m.d.av_comb += commit.eq(1)
-
-                                # TODO: set correct interrupt id from InterruptController
-                                # Set MSB - the Interrupt bit
-                                m.d.av_comb += cause_entry.eq(1 << (self.gen_params.isa.xlen - 1))
-                            with m.Else():
-                                # RISC-V synchronous exceptions - don't retire instruction that caused exception,
-                                # and later resume from it.
-                                # Value of ExceptionCauseRegister pc field is the instruction address.
-                                m.d.av_comb += commit.eq(0)
-
-                                m.d.av_comb += cause_entry.eq(cause_register.cause)
-
-                            m_csr.mcause.write(m, cause_entry)
-                            m_csr.mepc.write(m, cause_register.pc)
-                            self.trap_entry(m)
-
-                            with m.If(core_empty):
-                                m.next = "TRAP_RESUME"
-                            with m.Else():
-                                m.next = "TRAP_FLUSH"
-
-                        # Not using default condition, because we want to block if branch is not ready
-                        with cond(~rob_entry.exception):
-                            # Normally retire all non-trap instructions
+                        with m.If(cause_register.cause == ExceptionCause._COREBLOCKS_ASYNC_INTERRUPT):
+                            # Async interrupts are inserted only by JumpBranchUnit and conditionally by MRET and CSR
+                            # The PC field is set to address of instruction to resume from interrupt (e.g. for jumps
+                            # it is a jump result).
+                            # Instruction that reported interrupt is the last one that is commited.
                             m.d.av_comb += commit.eq(1)
 
-                    # Methods cannot be called multiple times from the same Transaction >:(
+                            # TODO: set correct interrupt id from InterruptController
+                            # Set MSB - the Interrupt bit
+                            m.d.av_comb += cause_entry.eq(1 << (self.gen_params.isa.xlen - 1))
+                        with m.Else():
+                            # RISC-V synchronous exceptions - don't retire instruction that caused exception,
+                            # and later resume from it.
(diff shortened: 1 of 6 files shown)


coreblocks/structs_common/exception.py
```diff
@@ -54,7 +56,15 @@ def __init__(self, gp: GenParams, rob_get_indices: Method):
         self.pc = Signal(gp.isa.xlen)
         self.valid = Signal()
 
-        self.report = Method(i=gp.get(ExceptionRegisterLayouts).report)
+        self.layouts = gp.get(ExceptionRegisterLayouts)
+
+        # Break long combinational paths from single-cycle FUs
+        # Insertion of FIFO is fine, because self.report is always ready, and will delay report by
```
tilk, line 62: This is a single FIFO for all FUs. What will happen when two FUs report an exception in the same clock cycle? Did that work previously?

lekcyjna123 (approved):

coreblocks/structs_common/exception.py
piotro888: It is the same behaviour as last time. Both FUs cannot report exception at the same time, and only one will proceed with its `Transaction`.
This is a performance problem, because calls to report are under Ifs so all FUs reporting instruction are mutually exclusive at some stage.
As we discussed on the meeting, it may be solved via call under `condition` (btw, does it fix that -- will it block parent transction if we want to call report but it is not ready??).
Optimizing to allow reporting multiple exceptions at one cycle is not worth it (we have large flush/resume penalty, so one additional blocked cycle wouldn't change too much).
Anyway it is topic for another PR.

tilk: I'm still wondering about the cycle delays.

* The delay from e.g. CSR to ROB change is in best case only one cycle. CSR's `get_result` goes through `FuncBlocksUnifier`, which uses a `Collector`, which uses a `Forwarder`, which finally connects to the transaction in `ResultAnnouncement`. The same transaction calls `rob_mark_done`, which updates the ROB.

* On the other hand, the update of the `ExceptionCauseRegister` is two cycles from CSR's `get_result`: one cycle for the FIFO, and another cycle for the register update.

* Therefore there is a possibility for `Retirement` to see the ROB exception bit set, but not have the right cause.

Am I wrong?

Ideally, I would like to have this latency-insensitive; in other words, `Retirement` should wait if the `rob_id` in `ExceptionCauseRegister` is not the `rob_id` of currently processed entry. One way to do this cleanly would be to add a `rob_id` argument to the `ExceptionCauseRegister`'s `get` method, and use `validate_arguments` to only accept the call when the `rob_id` matches  the `rob_id` in the register.

piotro888: Oh, you are right, thanks! I missed the sync delay in `ExceptionCauseRegister`.

`Retirement` solution seems nice. 
Now I think about it, we don't even need to do that! We just need to check a `valid` bit of `ExceptionCauseRegister`. We only care about the first interrupt. And if it is processed, InterruptCauseRegister is in the clear() state!
I will look into `validate_arguments`, I didn't used it before. Hope it is possible to do that (we need exception bit from rob_peek and exception cause register result to calculate it)

At the same time I'm preparing Retirement refactor (and will make PR for it in a minute), I think it should be merged first (but lets see what the reception will be)

tilk: > Now I think about it, we don't even need to do that! We just need to check a valid bit of ExceptionCauseRegister. We only care about the first trap. 

The first trap in wall clock time might come from an instruction which was not the first in instruction order. So the valid bit will be 1, but the rob_id will be wrong, and we get the wrong cause. Am I missing something?

piotro888: You are right, my bad

piotro888: There is still one problem - ExceptionCauseRegister allows reporting multiple exceptions for one instruction, and decides if the cause should be updated. In this case rob_id would match, but returned cause would be outdated.

Looking at current usage of ExceptionCauseRegister, this functionality may be not needed at all. All exceptions in FUs are reported in one place only (when setting exception bit), so priority is selected locally. If any exception happens before FU it is also reported in ExceptionFU - priority also needs to be manually determined  in pipeline.

I think we can safely add limitation to support reporting only one exception per rob_id.
ECR will be simplified and I don't know if the addition of  separation FIFO for Fmax will be necessary. Should it stay? It probably be too much noticeable, because JB mispedictions report exceptions one cycle earlier that your example.

coreblocks/stages/retirement.py
```diff
@@ -97,49 +97,53 @@ def flush_instr(rob_entry):
...
-                        # Normally retire all non-trap instructions
-                        m.d.av_comb += commit.eq(1)
-
                     # Methods cannot be called multiple times from the same Transaction >:(
                     with condition(m, priority=False) as cond:
                         with cond(commit):
                             retire_instr(rob_entry)
-                        with cond(~commit):  # Not using default, because we want to block if condition is not ready
+                        with cond(~commit):
                             flush_instr(rob_entry)
```
piotro888, line 147: This `condition` in retirement can be eliminated with merge of #551, by using `retire_instr` and `flush_instr` instead of setting `commit` flag. Non-blocking is guaranteed by condition that is needed on higher level for `rob.exception` (and makes more sense)

piotro888:
For some reason Fmax is now 40 MHz with critical path not related to changes here :/

coreblocks/stages/retirement.py
tilk: If I remember right, `condition` was put here not for aesthetics, but to avoid locking FRAT's `rename` method.

piotro888: It could be removed because it would be separated by existing `condition` of (~)`rob_entry.exception`.

No longer valid, because that exception `condition` caused problems with 2* LUT usage.
I fixed it by using separate Transaction instead of `validate_args`. `get` method is not blocking so `excpetion` condition can be removed and `commit` condition stays the same.

piotro888:
Fixed, Fmax is 54 MHz, LUT usage back to normal.
It was caused by `rob_entry.exception` condition. Maybe it needs investigation, I don't see a reason why it doubled the LUT usage. Still happened at [this commit](https://github.com/kuznia-rdzeni/coreblocks/pull/545/commits/96400a6db3b247cbab89abdc293b5ae89a9475d3) after removal of validate_args.
For more change details refer to review comments

lekcyjna123:
It is interesting. I checkouted 96400a6db3b247cbab89abdc293b5ae89a9475d3 generated verilog by `./scripts/gen_verilog.py -c basic` and tried to synthesise it with Quartus, but I got syntax error. After some manual tweaks code get accepted. On the newest commit there is no more syntax error reported by Quartus.

tilk (approved):

lekcyjna123 (approved):

PR #548, opened 2023-12-20
Retirement refactor

piotro888:
_~CONTROVERSY ALERT ❗~_

I started working on speculation and saw a pattern in Retirement. I refactored it as FSM that calls python methods!
This way we can eliminate:
* All of the ugly sync/comb duo control signals - python functions that call Methods are executed under If from any state. When implementing misprediction interrupt I would need to add another pair of signalling like that. (also turns out there was previously probably a bug, because rrat call only used sync side_fx)
* Forwarders - now potentially (almost all) blocking methods are called from separate Transactions.
* Conditions in Ifs simplified

I think that logic is also way easier to understand with clear separation of retire/flush and state of what the retirement currently is doing.

~~Unfortunately the limitation that methods can be called only once from transaction, breaks fun here. It would look even nicer without it [see commit](https://github.com/kuznia-rdzeni/coreblocks/commit/c95fdcb0b665d5ffa086265d84d0ec1396c6dee1). And `MethodBrancher` cannot be used in this case.~~

For me it looks great. What do you think about it?
Also it would be nice to review/merege it quickly, as my further work on speculation depends on this (it would be ugly and bug-prone otherwise)

### DIFF
--- a/coreblocks/core.py
+++ b/coreblocks/core.py
@@ -148,6 +148,7 @@ class Core(Elaboratable):
             rob_peek=rob.peek,
             rob_retire=rob.retire,
             r_rat_commit=rrat.commit,
+            r_rat_peek=rrat.peek,
             free_rf_put=free_rf_fifo.write,
             rf_free=rf.free,
             precommit=self.func_blocks_unifier.get_extra_method(InstructionPrecommitKey()),
--- a/coreblocks/params/layouts.py
+++ b/coreblocks/params/layouts.py
@@ -216,18 +216,19 @@ class RATLayouts:
         self.old_rp_dst: LayoutListField = ("old_rp_dst", gen_params.phys_regs_bits)
         """Physical register previously associated with the given logical register in RRAT."""
 
-        self.rat_rename_in: LayoutList = [
+        self.frat_rename_in: LayoutList = [
             fields.rl_s1,
             fields.rl_s2,
             fields.rl_dst,
             fields.rp_dst,
         ]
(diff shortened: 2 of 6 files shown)


piotro888:
There seems to be loss in benchmark IPC for some reason. Needs investigation

piotro888:
I see, FRAT is still called under one condition form main Retirement. I will probably re-introduce one Forwarder, maybe there is some better solution?

tilk:
> I see, FRAT is still called under one condition form main Retirement. I will probably re-introduce one Forwarder, maybe there is some better solution?

Maybe try using `condition`? This would split that transaction into two.

piotro888:
> Maybe try using `condition`? This would split that transaction into two.

Great! It also fixed problem with call to `free_phys_reg`. Thanks!
I will check the IPC in a moment to be sure

piotro888:
oops, there is a problem with generating graph and condtion :(. [There is also graph syntax error on master again]

tilk:
> oops, there is a problem with generating graph and condtion :(. [There is also graph syntax error on master again]

The issue is in `get_caller_class_name`, I'll prepare a fix soon.

lekcyjna123 (approved):

tilk:
The Fmax is still pretty bad, but I imagine that #545 can still fix it?

piotro888:
Fmax issue is unrelated to this PR, it depends only on critical path from JBUnit to ExcepionCauseRegister.
IPC will be back to best value also after merge #545, becaue confilcts are removed here. It will not affect this Retirement code, because stall method is called from exclusive Transaction

tilk (approved):

PR #550, opened 2023-12-21
Fix get_caller_class_name to handle calls in elaborate

tilk:
This fixes call graph generation when using `condition` and in other cases.


piotro888 (approved):
LGTM

lekcyjna123 (approved):

PR #551, opened 2023-12-22
Exclusive branches

### LINKED ISSUE #549: Use mutual exclusivity of branches in Transactron
The branches of `m.If`, `m.FSM` etc. are mutually exclusive. This information might be used by the Transactron framework the following ways:

* Allow calling the same method in multiple different branches. Would make #539 unnecessary.
* Allow to remove conflicts between methods/transactions defined in different branches. This can improve the performance of transaction scheduling in such cases.

tilk:
This work in progress PR is intended to improve readability and performance of code where Amaranth branches interact with Transactron. In particular:

* Transactions/methods defined in different branches can be marked as conflict-free, as they can't run in the same cycle.
* Calling the same method from different branches are allowed.

TODO:
* [x] Make exclusive branch detection correct.
* [x] Find places where a workaround for single call per body was used
* [x] Implement tests.

Closes #549.

### DIFF
--- a/coreblocks/stages/retirement.py
+++ b/coreblocks/stages/retirement.py
@@ -102,8 +102,6 @@ class Retirement(Elaboratable):
 
                         cause_register = self.exception_cause_get(m)
 
-                        cause_entry = Signal(self.gen_params.isa.xlen)
-
                         with m.If(cause_register.cause == ExceptionCause._COREBLOCKS_ASYNC_INTERRUPT):
                             # Async interrupts are inserted only by JumpBranchUnit and conditionally by MRET and CSR.
                             # The PC field is set to address of instruction to resume from interrupt (e.g. for jumps
@@ -113,16 +111,18 @@ class Retirement(Elaboratable):
 
                             # TODO: set correct interrupt id from InterruptController
                             # Set MSB - the Interrupt bit
-                            m.d.av_comb += cause_entry.eq(1 << (self.gen_params.isa.xlen - 1))
+                            m_csr.mcause.write(m, data=1 << (self.gen_params.isa.xlen - 1))
                         with m.Else():
                             # RISC-V synchronous exceptions - don't retire instruction that caused exception,
                             # and later resume from it.
                             # Value of ExceptionCauseRegister pc field is the instruction address.
                             m.d.av_comb += commit.eq(0)
 
-                            m.d.av_comb += cause_entry.eq(cause_register.cause)
+                            cause_entry = Cat(
+                                cause_register.cause, C(0, self.gen_params.isa.xlen - len(cause_register.cause))
+                            )
+                            m_csr.mcause.write(m, data=cause_entry)
 
-                        m_csr.mcause.write(m, cause_entry)
                         m_csr.mepc.write(m, cause_register.pc)
                         self.trap_entry(m)
 
@@ -135,7 +135,6 @@ class Retirement(Elaboratable):
                         # Normally retire all non-trap instructions
                         m.d.av_comb += commit.eq(1)
 
-                    # Methods cannot be called multiple times from the same Transaction >:(
                     with condition(m, priority=False) as cond:
                         with cond(commit):
                             retire_instr(rob_entry)
--- a/coreblocks/stages/retirement.py
+++ b/coreblocks/stages/retirement.py
@@ -102,6 +102,8 @@ class Retirement(Elaboratable):
 
                         cause_register = self.exception_cause_get(m)
 
+                        cause_entry = Signal(self.gen_params.isa.xlen)
+
                         with m.If(cause_register.cause == ExceptionCause._COREBLOCKS_ASYNC_INTERRUPT):
                             # Async interrupts are inserted only by JumpBranchUnit and conditionally by MRET and CSR.
                             # The PC field is set to address of instruction to resume from interrupt (e.g. for jumps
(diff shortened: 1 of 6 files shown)


coreblocks/stages/retirement.py
```diff
@@ -113,16 +111,18 @@ def flush_instr(rob_entry):
...
                             # RISC-V synchronous exceptions - don't retire instruction that caused exception,
                             # and later resume from it.
                             # Value of ExceptionCauseRegister pc field is the instruction address.
                             m.d.av_comb += commit.eq(0)
 
-                            m.d.av_comb += cause_entry.eq(cause_register.cause)
+                            cause_entry = Cat(
+                                cause_register.cause, C(0, self.gen_params.isa.xlen - len(cause_register.cause))
+                            )
+                            m_csr.mcause.write(m, data=cause_entry)
```
piotro888, line 124: I would not like this particular logic to be changed, because previous way will look nicer with misprediction support (when some causes don't trigger _architecture trap_ - this could be handled under one If with `mepc` and `trap_entry`)

> Find places where a workaround for single call per body was 

Other places that I remember this change would be useful are `CSRUnit`, and `JumpBranchUnit` (both related to `report` method)

transactron/core.py
```diff
@@ -610,6 +629,76 @@ def __setitem__(self, name: str, value):
...
+        et = EnterType
+
+        def flush():
+            while len(self._ctrl_path) > self.depth:
+                self._ctrl_path.pop()
+
+        if enter_type in [et.ADD, et.ENTRY]:
+            self._ctrl_path[-1] = replace(self._ctrl_path[-1], alt=self._ctrl_path[-1].alt + 1)
+        if enter_type == et.PUSH:
+            if len(self._ctrl_path) > self.depth:
```
lekcyjna123, line 682: Is this condition correct?

I don't see a scenario in which we get True. We start with balanced `depth` and `len(_ctrl_path)`. Each `Push` either increase both `_ctrl_path` and `depth` by the same amount or increase `depth` more than the `_ctrl_path`. Each `Add` increase `depth` and the `Entry` don't change neither lenght of `_ctrl_path` nor `depth`.

So the difference can be only `len(self._ctrl_path) <= self.depth`.

transactron/core.py
```diff
@@ -227,6 +228,18 @@ def _conflict_graph(method_map: MethodMap) -> Tuple[TransactionGraph, Transactio
             Linear ordering of transactions which is consistent with priority constraints.
         """
 
+        def transactions_exclusive(trans1: Transaction, trans2: Transaction):
+            tms1 = [trans1] + method_map.methods_by_transaction[trans1]
+            tms2 = [trans2] + method_map.methods_by_transaction[trans2]
+
```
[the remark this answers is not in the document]
lekcyjna123: Lets take such example:
```py
with transaction().body(): #parent
  method1()
  with m.If(1):
    method2()
  with m.Else():
    with transaction.body(): #child
      method1()
      with m.If(1):
        method2()
```
Are parent and child transaction exclusive or not? According to the below comment they should be exclusive because there exists the `method2` that is exclusive. But in parallel we have call to `method1` which is obviously wrong and shouldn't be allowed.

lekcyjna123 (commented):
Overall it is great change. I left some comments in code and please, add some documentation, docstrings, comments, markdown even in Quenya, but something is very needed.

transactron/core.py
tilk: The depth is computed eagerly, but the stack is popped lazily. So, before the stack flush, the stack can be larger than the depth.

transactron/core.py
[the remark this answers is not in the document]
tilk: They are exclusive and the child transaction will never run. This situation was not detected by the previous check, and is still not detected in this PR. I think this should be considered a separate issue.

transactron/core.py
tilk: Also, it is never the case that the depth increases, but the stack doesn't. ADD is never stand-alone, it always follows PUSH.

tilk: Refactored the code to eliminate ’depth’ and ’flush’. Hopefully it's clearer now.

test/transactions/test_branches.py
```diff
@@ -0,0 +1,97 @@
...
+        tm = TransactionManager()
+        dut = TransactionModule(circ, tm)
+
+        with self.run_simulation(dut):
+            pass
+
+        cgr, _, _ = tm._conflict_graph(MethodMap(tm.transactions))
+
+        for s in cgr.values():
+            self.assertFalse(s)
```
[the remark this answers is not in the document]
tilk: Using a data structure in a boolean context is a standard Pythonic way of testing emptiness.

transactron/core.py
```diff
@@ -227,6 +228,18 @@ def _conflict_graph(method_map: MethodMap) -> Tuple[TransactionGraph, Transactio
...
+            # if first transaction is exclusive with the second transaction, or this is true for
+            # any called methods, the transactions can't run at the same time
```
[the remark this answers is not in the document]
tilk: Changed to "will never".

lekcyjna123 (approved):
Thanks. Now it is much cleaner.

transactron/core.py
```diff
@@ -610,6 +629,118 @@ def __setitem__(self, name: str, value):
...
+        Path in the control tree, starting from the root.
+    """
+
+    module: int
+    path: list[PathEdge]
+
+    def exclusive_with(self, other: "CtrlPath"):
+        """Decides if this path is mutually exclusive with some other path.
+
+        Paths are mutually exclusive iff they refer to the same module and
```
piotro888, line 677:
```suggestion
        Paths are mutually exclusive if they refer to the same module and
```

piotro888 (approved):
Looks good!

transactron/core.py
tilk: Was intended, "iff" is a shorthand for [if and only if](https://en.wikipedia.org/wiki/If_and_only_if). But I'll change it - this is not a mathematically accurate text anyway.

Issue #405, opened 2023-06-30
Use pytest to parallellize tests


lekcyjna123:
We currently run tests using standard library "unittests" but it offers only running tests on one core. Please use `pytest` to parallelize testing.

lekcyjna123:
Solved by #554 and #611

Issue #200, opened 2023-01-18
Use SimpleTestCircuit where applicable


tilk:
In #187 there was a silent introduction of a class which simplifies writing typical tests, `SimpleTestCircuit`. It automatically creates `TestbenchIO`s for exposed methods, which allows to avoid creating a new `Elaboratable` for each test just for `TestbenchIO` wrapping. This task involves replacing such custom `Elaboratable`s with `SimpleTestCircuit`.

tilk:
As for today, test elaboratables can be found in tests of:

* Scheduler,
* Interrupt controller,
* JB unit,
* FPU,
* CSR unit,
* LSU,
* Wishbone and AXI,
* ICache,
* Retirement.

PR #553, opened 2023-12-31
Add pytest - step 1

lekcyjna123:
Hi,
here is a PR which is a first step towards using pytest as described in #405. In this PR:
- ~a basic configuration for pytest is added~
- ~a new dependecy to requirements is added~
- updated some stubs
- removed pytest warnings:
  - removed deprecated `Repl`
  - ported deprecated `RoundRobin` from amaranth to transactron, according to the recomendation from amaranth
  - updated some old tests which break test naming convention, to use newer way of writing tests by using `SimpleTestCircuit`
  - ~added exception for `TestbenchIO` naming which breaks test naming convention~
  - ~added exception for deprecation warnings related with `Record`~

There is a one warning left, but it should be fixed under separate PR, because it hides some error in `test_retirement.py` and it has to be investigated deeper.

### DIFF
--- a/test/frontend/test_decode.py
+++ b/test/frontend/test_decode.py
@@ -1,47 +1,37 @@
-from amaranth import Elaboratable, Module
-
 from transactron.lib import AdapterTrans, FIFO
 
-from ..common import TestCaseWithSimulator, TestbenchIO
+from ..common import TestCaseWithSimulator, TestbenchIO, SimpleTestCircuit, ModuleConnector
 
 from coreblocks.frontend.decode import Decode
 from coreblocks.params import GenParams, FetchLayouts, DecodeLayouts, OpType, Funct3, Funct7
 from coreblocks.params.configurations import test_core_config
 
 
-class TestElaboratable(Elaboratable):
-    def __init__(self, gen_params: GenParams):
-        self.gen_params = gen_params
-
-    def elaborate(self, platform):
-        m = Module()
+class TestFetch(TestCaseWithSimulator):
+    def setUp(self) -> None:
+        self.gen_params = GenParams(test_core_config.replace(start_pc=24))
 
         fifo_in = FIFO(self.gen_params.get(FetchLayouts).raw_instr, depth=2)
         fifo_out = FIFO(self.gen_params.get(DecodeLayouts).decoded_instr, depth=2)
 
-        self.io_in = TestbenchIO(AdapterTrans(fifo_in.write))
-        self.io_out = TestbenchIO(AdapterTrans(fifo_out.read))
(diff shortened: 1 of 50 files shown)


tilk:
This is mostly a refactor PR. I believe the first two points related to pytest can be delayed for another PR.

Question: why does the testing framework even care about `TestbenchIO`? Could you elaborate?

lekcyjna123:
`pytest` do an inspection over all modules which starts with `test_` or ends with `_test` and looks for the classes which starts from `Test` and try to interpret them as a test suites (see: https://docs.pytest.org/en/7.4.x/explanation/goodpractices.html#test-discovery).  We import `TestbenchIO` to nearly every test module and this class name starts with `Test`. So pytest tries to interpret `TestbenchIO` as a test suite which is not successful.

tilk:
Is it possible to just exclude the directories which shouldn't contain tests? The link you gave suggests it's possible.

test/frontend/test_decode.py
```diff
@@ -1,47 +1,37 @@
...
 from coreblocks.params.configurations import test_core_config
 
 
-class TestElaboratable(Elaboratable):
-    def __init__(self, gen_params: GenParams):
-        self.gen_params = gen_params
-
-    def elaborate(self, platform):
-        m = Module()
+class TestFetch(TestCaseWithSimulator):
```
piotro888, line 10:
```suggestion
class TestDecode(TestCaseWithSimulator):
```
:)

piotro888 (approved):
Nice test simplification, I think that #200 can be linked.

What is the problem with `test_retirement`?

lekcyjna123:
> Is it possible to just exclude the directories which shouldn't contain tests? The link you gave suggests it's possible.

It is possible, but not useful in our case, because `TestbenchIO` is imported in our test file namespace (by `from test.common import TestbenchIO`) and it is found in regular test files.

> test/common/testbenchio.py:11: 12 warnings
  .../coreblocks/test/common/testbenchio.py:11: PytestCollectionWarning: cannot collect test class 'TestbenchIO' because it has a \_\_init\_\_ constructor (from: test/structs_common/test_exception.py)
    class TestbenchIO(Elaboratable):

lekcyjna123:
> What is the problem with `test_retirement`?

We try to control signal value before start of simulation:
```py
    def test_rand(self):
        self.retc = RetirementTestCircuit(self.gen_params)

        yield from self.retc.mock_fetch_stall.enable()  # To be fixed
        with self.run_simulation(self.retc) as sim:
             ...
```

tilk:
> We try to control signal value before start of simulation:

This code doesn't make any sense. We missed it in #523.

tilk (approved):

PR #554, opened 2024-01-01
Add pytest - step 2

lekcyjna123:
Next step of adding pytest.
- a basic configuration for pytest is added
- a new dependecy to requirements is added


lekcyjna123:
TODO: Update scripts

lekcyjna123:
There is still one more race to fix :(

test/regression/test_regression.py
```diff
@@ -0,0 +1,127 @@
+from .memory import *
```
tilk, line 1: Why the file rename?

tilk (approved):

test/regression/test_regression.py
piotro888, line 1: I think that we should separate 'regression' as [external tests that may be run with verilator], apart from [the `riscv-tests` suite].

In this file `MMIO` and `rv32uc-rvc` are both very `riscv-test` specific, but it also contains `regression_body_with_cocotb`, `verilate_model` that seem like they are generic for all verilator regeression test concept.
It was also a problem with previous solution, so maybe it is a good time to separate and generalize it now? Or am I missing something and those would not be useful in more generic context?

Also I think that in `test/regression/conftest.py`, test collecting functions should be named more to reflect that they are configuring `riscv-tests` suite and not regression tests in general.

lekcyjna123: It is a good point, but I would prefer to do that in separate PR (the one in which multithreading will be added to riscof tests).

test/conftest.py
```diff
@@ -0,0 +1,92 @@
...
+        "--coreblocks-test-name",
+        action="store",
+        type=str,
+        help="Name or regexp in flatten format pointing to test to run.",
+    )
+    group.addoption(
+        "--coreblocks-test-count",
+        action="store",
+        type=int,
+        help="Number of tests to starts. If less than number of all selected tests, then starts only subset of them.",
```
piotro888, line 27:
```suggestion
        help="Number of tests to start. If less than number of all selected tests, then starts only subset of them.",
```

test/conftest.py
```diff
@@ -0,0 +1,92 @@
...
+        choices=["cocotb", "pysim"],
+        help="Simulation backend for regression tests",
+    )
+    group.addoption("--coreblocks-traces", action="store_true", help="Generate traces from regression tests")
+    group.addoption("--coreblocks-list", action="store_true", help="List all tests in flatten format.")
+    group.addoption(
+        "--coreblocks-test-name",
+        action="store",
+        type=str,
+        help="Name or regexp in flatten format pointing to test to run.",
```
piotro888, line 21:
```suggestion
    group.addoption(
        "--coreblocks-test-name",
        action="store",
        type=str,
        help="Name or regexp in flatten format matching the tests to run.",
```

test/regression/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+
+def pytest_generate_tests(metafunc: pytest.Metafunc):
+    if not metafunc.config.getoption("coreblocks_regression"):
+        # Add regression to skiped tests
+        metafunc.parametrize(["test_name", "backend", "traces", "verbose"], [])
+        return
+
+    all_tests = (
+        load_regression_tests()
+    )  # The list has to be always in the samo order (e.g. sorted) to allow for parallel testing
```
piotro888, line 37:
```suggestion
    )  # The list has to be always in the same order (e.g. sorted) to allow for parallel testing
```

piotro888 (approved):
LGTM, only some typos and description suggestions.

PR #557, opened 2024-01-06
Speculation Backend

piotro888:
Speculation implementation. Fetch is no longer stalled on jump/branches.

Branch Predictor is not in the scope of this PR. `Fetch` always fetches next instruction and `JBUnit` has hard-coded misprediction on jump taken.

* `JBUnit` triggers internal exception if misprediction is detected and submits hint for predictor.
* Fetch continue and branch predictor [from_pc, next_pc] are finally separated.
* ExceptionCauseRegister stalls the fetch immediately after receiving report. (optimization, previously done in Retirement)
* Retirement handles misprediction as normal exception, but without modifying CSRs and continuing from reported `pc` instead of handler
* Test cycle counts are temporarily adjusted, all will need to be set again after some branch predictor is implemented (TODO: set regression and benchmark timeouts)

Everything else should be hopefully straight-forward.

For implementation of jump prediction new pipeline field needs to be added. Stalling fetch on jumps should be probably restored, but jump predictor that will be after fetch will unstall it in small cycle delay (unless it is single-cycle predictor embedded into fetch)

### DIFF
--- a/coreblocks/frontend/fetch.py
+++ b/coreblocks/frontend/fetch.py
@@ -32,7 +32,7 @@ class Fetch(Elaboratable):
 
         self.verify_branch = Method(i=self.gen_params.get(FetchLayouts).branch_verify)
         self.stall_exception = Method()
-        self.stall_exception.add_conflict(self.verify_branch, Priority.LEFT)
+        self.stall_exception.add_conflict(self.verify_branch, Priority.LEFT)  #
 
         # PC of the last fetched instruction. For now only used in tests.
         self.pc = Signal(self.gen_params.isa.xlen)
@@ -130,7 +130,7 @@ class UnalignedFetch(Elaboratable):
 
         self.verify_branch = Method(i=self.gen_params.get(FetchLayouts).branch_verify)
         self.stall_exception = Method()
-        self.stall_exception.add_conflict(self.verify_branch, Priority.LEFT)
+        self.stall_exception.add_conflict(self.verify_branch, Priority.LEFT)  #
 
         # PC of the last fetched instruction. For now only used in tests.
         self.pc = Signal(self.gen_params.isa.xlen)
--- a/coreblocks/frontend/fetch.py
+++ b/coreblocks/frontend/fetch.py
@@ -73,9 +73,7 @@ class Fetch(Elaboratable):
 
             opcode = res.instr[2:7]
             # whether we have to wait for the retirement of this instruction before we make futher speculation
-            unsafe_instr = (
-                (opcode == Opcode.BRANCH) | (opcode == Opcode.JAL) | (opcode == Opcode.JALR) | (opcode == Opcode.SYSTEM)
-            )
+            unsafe_instr = opcode == Opcode.SYSTEM
 
             with m.If(spin == target.spin):
                 instr = Signal(self.gen_params.isa.ilen)
@@ -194,9 +192,7 @@ class UnalignedFetch(Elaboratable):
 
             opcode = instr[2:7]
             # whether we have to wait for the retirement of this instruction before we make futher speculation
-            unsafe_instr = (
-                (opcode == Opcode.BRANCH) | (opcode == Opcode.JAL) | (opcode == Opcode.JALR) | (opcode == Opcode.SYSTEM)
-            )
+            unsafe_instr = opcode == Opcode.SYSTEM
 
             # Check if we are ready to dispatch an instruction in the current cycle.
             # This can happen in three situations:
--- a/coreblocks/frontend/fetch.py
+++ b/coreblocks/frontend/fetch.py
@@ -30,9 +30,9 @@ class Fetch(Elaboratable):
         self.icache = icache
         self.cont = cont
 
-        self.verify_branch = Method(i=self.gen_params.get(FetchLayouts).branch_verify)
+        self.resume = Method(i=self.gen_params.get(FetchLayouts).resume)
         self.stall_exception = Method()
-        self.stall_exception.add_conflict(self.verify_branch, Priority.LEFT)
+        self.stall_exception.add_conflict(self.resume, Priority.LEFT)
 
         # PC of the last fetched instruction. For now only used in tests.
         self.pc = Signal(self.gen_params.isa.xlen)
@@ -92,9 +92,9 @@ class Fetch(Elaboratable):
 
                 self.cont(m, instr=instr, pc=target.addr, access_fault=fetch_error, rvc=0)
 
-        @def_method(m, self.verify_branch, ready=stalled)
-        def _(from_pc: Value, next_pc: Value, resume_from_exception: Value):
-            m.d.sync += speculative_pc.eq(next_pc)
+        @def_method(m, self.resume, ready=stalled)
+        def _(pc: Value, resume_from_exception: Value):
+            m.d.sync += speculative_pc.eq(pc)
             m.d.sync += stalled_unsafe.eq(0)
             with m.If(resume_from_exception):
                 m.d.sync += stalled_exception.eq(0)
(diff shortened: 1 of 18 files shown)


test/frontend/test_fetch.py
```diff
@@ -130,7 +130,7 @@ def fetch_out_check(self):
             instr = self.instr_queue.popleft()
             if instr["is_branch"]:
                 yield from self.random_wait(10)
-                yield from self.m.verify_branch.call(from_pc=instr["pc"], next_pc=instr["next_pc"])
+                yield from self.m.fetch_resume.call(from_pc=instr["pc"], next_pc=instr["next_pc"])
```
tilk, line 133: The `fetch_resume` method doesn't have those arguments, and the test fails.

test/stages/test_retirement.py
```diff
@@ -153,11 +147,12 @@ def precommit_process(rob_id, side_fx):
...
 
-        with self.run_simulation(retc) as sim:
-            sim.add_sync_process(retire_process)
-            sim.add_sync_process(peek_process)
-            sim.add_sync_process(free_reg_process)
-            sim.add_sync_process(rat_process)
-            sim.add_sync_process(rf_free_process)
-            sim.add_sync_process(precommit_process)
-            sim.add_sync_process(exception_cause_process)
+        # TODO: Retirement test is currently broken because of nonexclusive mock issue
```
tilk, line 150: This PR needs to be updated to master, as this issue is already fixed.

coreblocks/frontend/fetch.py
```diff
@@ -29,8 +30,9 @@ def __init__(self, gen_params: GenParams, icache: ICacheInterface, cont: Method)
         self.icache = icache
         self.cont = cont
 
-        self.verify_branch = Method(i=self.gen_params.get(FetchLayouts).branch_verify)
+        self.resume = Method(i=self.gen_params.get(FetchLayouts).resume)
         self.stall_exception = Method()
+        self.stall_exception.add_conflict(self.resume, Priority.LEFT)
```
tilk, line 35: Does this conflict actually happen? If so, in what situations?

test/fu/functional_common.py
```diff
@@ -155,7 +155,9 @@ def setUp(self):
 
             self.responses.append({"rob_id": rob_id, "rp_dst": rp_dst, "exception": int(cause is not None)} | results)
             if cause is not None:
-                self.exceptions.append({"rob_id": rob_id, "cause": cause, "pc": pc})
+                # JumpBranchUnit specific :( - misprediction has target pc (appended form verify_branch layout)
```
lekcyjna123, line 158: So maybe we should provide a function which will calculate `e_pc` with default implementation `e_pc = pc`. In JumpBranchUnit this function could be overwritten.

lekcyjna123 (commented):
LGTM

coreblocks/frontend/fetch.py
piotro888: This conflict can happen and is valid. `resume` can be called from `CSRUnit` to unstall from 'unsafe' `SYSTEM` instruction, and fetch can be stalled at the same time by `ExceptionCauseRegister` reporting.

It should not cause performance problems, because `stall_exception` is called only by `ExceptionCauseRegister` that uses entirely separate Transaction (separated via FIFO). Am I right?

test/fu/functional_common.py
piotro888: I extended this idea to use 'temporary' field in returned result like it is done for `exception` field. Thanks!

piotro888:
Regression tests are failing with 10/20x larger timeouts, so it looks like there could be some hidden bug in speculation (or exception?) implementation :(((

I will try to look at the traces tomorrow ((but any help is welcome))

coreblocks/frontend/fetch.py
tilk: > This conflict can happen and is valid. resume can be called from CSRUnit to unstall from 'unsafe' SYSTEM instruction, and fetch can be stalled at the same time by ExceptionCauseRegister reporting.

Can it? SYSTEM instructions stall the pipeline, so that at the moment the instruction executes, there is no other instruction in the core. Therefore, for this to happen, the CSR unit would have to simultaneously call `resume` and report an exception, which feels weird. What did I miss?

Still, if this doesn't affect performance, this can stay - but maybe a comment to explain why it's there would be nice.

piotro888:
I found that on the `xor` test Retirement switches to `TRAP_FLUSH` state, executes correctly for last instruction (core instruction counter decrements to 0), but it never changes state to `TRAP_RESUME`.
I don't see why. 
I tried temporarily changing condition in `decrement` from `(count == 1) & ~self.increment.run` to only `count == 1`, but it doesn't help

piotro888:
I found an issue with current approach to flushing the core that needs discussion:

1. Instruction that will cause exception enters pipeline
2. Instruction with `rp_dst` = 8 enters pipeline
3. Few other instructions enter pipeline
4. Instruction that depends on `rp_s1` = 8 enters core.
5. Result of instruction 1. and 2. is computed, retirement fetches 1. from ROB
6. Exception flag is set, ROB is flushing the core
7. `rp_dst` = 8 of instruction 2. is flushed and its `rp_dst` is discarded as invalid. That means pushing to freeRF fifo and marking `rp_dst` in `RF` as NOT VALID.
8. After this moment, instruction 4. eneters `RSInsertion`. `rp_s1` = 8 is marked as NOT VALID in RF, so it is inserted into RS to wait for its Announcement.
9. Core hangs up, RS will wait indefinitely for register to be announced, that was already freed!

Solutions that I see:
1) Simplest - Detect flusing in `RSInsertion` stage and always treat source regisers as ready with dummy value
2) Detect flushing in `RSInsertion`, and forward all instructions directly to `Backend` with dummy results. Skipping RS and FUs. Would need some special way to unreserve FUs.
3) Detect flushing in `RS`. Don't send any instruction to FUs, forward all stored/waiting (in any entry completion stage) instructions directly to `Backend` with dummy result.
4) ??? Something better ???? Do you have some ideas?

--
Additionally I fixed bug with `jal` not writing its result because of unneded `If` in `Backend`.
Issue from the previous comment happens only on cocotb backend, and I'm worried this may be Amaranth issue? Probably needs inspection of generated verilog for FSM.

lekcyjna123:
In my opinion the first solution is consistent with our idea of core cleaning, where everything flows nearly as normal, so I would recommend using it. Alternatively I am ready to accept 3. solution, but I am worry about critical paths in it.

tilk:
I'm thinking about the solution 3. This would have the additional upside that long instructions wouldn't run and instructions wouldn't have to wait for dependencies, reducing the miss penalty.

piotro888:
> Issue from the previous comment happens only on cocotb backend, and I'm worried this may be Amaranth issue? Probably needs inspection of generated verilog for FSM.

It turns out I misinterpreted CoreInstructionCounter value (not that obvious in verilator traces) and there was 1 instruction still waiting in core because of RS bug.
There is no unexpected behaviour, all riscv-tests pass after fix of previous two issues.
Only thing it shows is that we need to rewrite pysim/coctb regression memory models, so they behave in the same way, with the same timing.

I will push PR this morning.

piotro888:
Everything fixed now!

test/frontend/test_fetch.py
```diff
@@ -43,14 +43,16 @@ def setUp(self) -> None:
 
         fifo = FIFO(self.gen_params.get(FetchLayouts).raw_instr, depth=2)
         self.io_out = TestbenchIO(AdapterTrans(fifo.read))
+
         self.fetch = SimpleTestCircuit(Fetch(self.gen_params, self.icache, fifo.write))
-        self.verify_branch = TestbenchIO(AdapterTrans(self.fetch._dut.verify_branch))
+
+        # self.fetch_resume = TestbenchIO(AdapterTrans(self.fetch.resume))
+        # self.fetch_stall_exception = TestbenchIO(AdapterTrans(self.fetch.stall_exception))
```
tilk, line 50: These are commented out. Are these for extending the fetch test in the future? If so, a comment should document this.

tilk (approved):

test/scheduler/test_scheduler.py
```diff
@@ -371,3 +382,4 @@ def process():
             )
             sim.add_sync_process(self.make_queue_process(io=self.m.free_rf_inp, input_queues=[self.free_regs_queue]))
             sim.add_sync_process(instr_input_process)
+            sim.add_sync_process(core_state_mock)
```
lekcyjna123, line 385: Shouldn't be needed any more

lekcyjna123 (approved):

piotro888:
i will try fix conflicts and comments today

test/frontend/test_fetch.py
piotro888: These `TestbenchIOs` were duplicated by `SimpleTestCircuit`, it was missed in earlier PR. I forgot to remove them after testing

PR #558, opened 2024-01-07
Allow additional method arguments in Adapter

piotro888:
Fixes ancient "bug" with mocking non-exclusive methods. Who would thought that mocked Method created in adapter actually needs to be specified nonexclusive?

I think that forwarding arguments via `**kwargs` makes more sense here, other solution would be to explicitly add all other Method parameters to Adapter.

### DIFF
--- a/transactron/lib/adapters.py
+++ b/transactron/lib/adapters.py
@@ -100,9 +100,7 @@ class Adapter(AdapterBase):
             See transactron.core.Method.__init__ for parameters description.
         """
 
-        if "src_loc" not in kwargs:
-            kwargs["src_loc"] = 0
-        kwargs["src_loc"] = get_src_loc(kwargs["src_loc"])
+        kwargs["src_loc"] = get_src_loc(kwargs.setdefault("src_loc", 0))
 
         super().__init__(Method(**kwargs))
         self.data_in = Record.like(self.iface.data_out)
--- a/transactron/lib/adapters.py
+++ b/transactron/lib/adapters.py
@@ -3,6 +3,7 @@ from amaranth import *
 from ..utils import SrcLoc, get_src_loc
 from ..core import *
 from ..core import SignalBundle
+from ..utils._typing import type_self_kwargs_as
 
 __all__ = [
     "AdapterBase",
@@ -91,6 +92,7 @@ class Adapter(AdapterBase):
         Data passed as argument to the defined method.
     """
 
+    @type_self_kwargs_as(Method.__init__)
     def __init__(self, **kwargs):
         """
         Parameters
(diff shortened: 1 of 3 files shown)


piotro888:
IIRC, there was a discussion when `rob.peek`, that `rob.retire` could return nothing now peek could be used in the same transaction. It was not implemented because testing failed (and that is the reason for TODO: comment). Currently `retire` duplicates `peek` return data.

Should we remove return data in this/next PR or leave it as is?

tilk:
I guess this explains the problem ;)

> Should we remove return data in this or other PR or leave it as is?

Maybe in another PR. You could change it together with restoring the retirement tests.

tilk (approved):

transactron/lib/adapters.py
```diff
@@ -92,22 +91,20 @@ class Adapter(AdapterBase):
...
+        **kwargs
+            Keyword arguments for Method that will be created.
+            See transactron.core.Method.__init__ for parameters description.
         """
-        src_loc = get_src_loc(src_loc)
-        super().__init__(Method(name=name, i=i, o=o, src_loc=src_loc))
+
+        if "src_loc" not in kwargs:
+            kwargs["src_loc"] = 0
+        kwargs["src_loc"] = get_src_loc(kwargs["src_loc"])
```
makz00, line 105: You can use _setdefault_ on kwargs and remove condition. It will be shorter

piotro888:
@tilk I found a way to 'copy' kwargs type from other method definition:

```python
P = ParamSpec("P")
T = TypeVar("T")
def type_kwargs(
    kwargs_call: Callable[Concatenate[Any, P], Any]
) -> Callable[[Callable[Concatenate[Any, ...], T]], Callable[Concatenate[Any, P], T]]:
    def return_func(func:Callable[Concatenate[Any, ...], T]) -> Callable[Concatenate[Any, P], T]:
        return cast(Callable[Concatenate[Any, P], T], func)

    return return_func
```
```python
class Adapter(AdapterBase):
   @type_kwargs(Method.__init__)
    def __init__(self, **kwargs):
```
Not too obvious, but it works.

Source:
* https://stackoverflow.com/questions/71968447/python-typing-copy-kwargs-from-one-function-to-another
* https://sobolevn.me/2021/12/paramspec-guide
* https://stackoverflow.com/questions/73856901/how-can-i-use-paramspec-with-method-decorators

Its basically solution from first link but with added `Concatenate[Any, *]` to support `Adapter->Method` `self` argument mismatch.

Do you think it is worth including in `transactron.utils._typing` and using it here?

tilk:
If it works, you can go ahead.

makz00 (approved):

tilk (approved):

PR #562, opened 2024-01-10
Retirement test fix

### LINKED ISSUE #556: Fix retirement test
It was missed in review that test_rand in retirement tests has a yield outside of test processes.

piotro888:
Closes #556 
Adds missing mocks, fixes `nonexclusive`and resolves todo with retire/peek functionality duplication
(test was not running at all because of yield that lekcyjna123 found)


test/stages/test_retirement.py
```diff
@@ -162,3 +184,8 @@ def test_rand(self):
             sim.add_sync_process(self.rf_free_process)
             sim.add_sync_process(self.precommit_process)
             sim.add_sync_process(self.exception_cause_process)
+            sim.add_sync_process(self.exception_clear_process)
```
lekcyjna123, line 187: I have resolved comments from #540, so it can be now merged. In such case this should be removed.

lekcyjna123 (approved):

PR #563, opened 2024-01-10
More readable exception on invalid def_method_mock function

piotro888:
Writing class-level mock with `def_method_mock` and without `self` argument passes typing and produces very unreadable exception of `Invalid method signature` without any indication of method/mock that caused problem (not even is stack trace!)

As this is an easy to do mistake, I think this patch is worth including.


tilk (approved):
LGTM

lekcyjna123 (approved):

PR #565, opened 2024-01-11
Remove unneeded random_wait

### LINKED ISSUE #564: Standarize "random_wait" in tests
As pointed by https://github.com/kuznia-rdzeni/coreblocks/pull/540#discussion_r1441609329 we have in many places copied definition of "random_wait", this should be cleaned.

lekcyjna123:
To resolve #564, here is a PR which remove unneeded `random_wait` from many places. Additionally it refactores two next tests to use `SimpleTestCircuit`

### DIFF
--- a/test/common/infrastructure.py
+++ b/test/common/infrastructure.py
@@ -225,11 +225,9 @@ class TestCaseWithSimulator(unittest.TestCase):
 
     def random_wait(self, max_cycle_cnt: int, *, min_cycle_cnt: int = 0):
         """
-        Wait for a random amount of cycles in range [min_cycle_cnt, max_cycle_cnt)
+        Wait for a random amount of cycles in range [min_cycle_cnt, max_cycle_cnt]
         """
-        if max_cycle_cnt == 0:
-            return
-        yield from self.tick(random.randrange(min_cycle_cnt, max_cycle_cnt))
+        yield from self.tick(random.randrange(min_cycle_cnt, max_cycle_cnt + 1))
 
     def random_wait_geom(self, prob: float = 0.5):
         """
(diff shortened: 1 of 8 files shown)


test/transactions/test_transaction_lib.py
```diff
@@ -164,7 +158,8 @@ def writer():
                 for _ in range(2):
                     yield Settle()
                 data_dict[a] = d
-                yield from random_wait(writer_rand)
+                yield
```
xThaid, line 161: We can also add an optional parameter to `random_wait` to specify the minimum number of cycles.

makz00:
Also there is another _random_wait()_ method in _test_csr.py_ which looks redundant

lekcyjna123:
> Also there is another _random_wait()_ method in _test_csr.py_ which looks redundant

Added `wait_random_geom` to solve that.

piotro888 (approved):

test/common/infrastructure.py
```diff
@@ -215,16 +215,25 @@ def run_simulation(self, module: HasElaborate, max_cycles: float = 10e4, add_tra
...
-    def random_wait(self, max_cycle_cnt):
+    def random_wait(self, max_cycle_cnt: int, *, min_cycle_cnt: int = 0):
         """
-        Wait for a random amount of cycles in range [1, max_cycle_cnt)
+        Wait for a random amount of cycles in range [min_cycle_cnt, max_cycle_cnt)
         """
-        yield from self.tick(random.randrange(max_cycle_cnt))
+        if max_cycle_cnt == 0:
+            return
+        yield from self.tick(random.randrange(min_cycle_cnt, max_cycle_cnt))
```
piotro888, line 232: A bit of bikeshedding, but (if already modifying existing compatibility) I think that [min, max] ranges would be more intuitive to use.

Additional benefit is that would eliminate this extra `if` and rand_wait(0) would actually differ from rand_wait(1)
What do you think?

lekcyjna123: Done

xThaid (approved):

PR #568, opened 2024-01-21
Fix core freezing during flushing

piotro888:
Exported from #557 for easier review. For details about changes see comments in #557 

I implemented workaround 1) for now, because it is quickest to do and it will be still needed with solution 3) (because not all FBs use RS - CSRUnit, LSU). 3) can be added in future PR.

### DIFF
--- a/test/scheduler/test_scheduler.py
+++ b/test/scheduler/test_scheduler.py
@@ -369,6 +369,7 @@ class TestScheduler(TestCaseWithSimulator):
 
         @def_method_mock(lambda: self.m.core_state)
         def core_state_mock():
+            # TODO: flushing test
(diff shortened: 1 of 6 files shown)


test/scheduler/test_scheduler.py
```diff
@@ -360,6 +367,10 @@ def process():
 
             return process
 
+        @def_method_mock(lambda: self.m.core_state)
+        def core_state_mock():
+            return {"flushing": 0}
```
tilk, line 372: Leave some TODO here.

tilk:
```suggestion
            # TODO: flushing test
            return {"flushing": 0}
```

PR #572, opened 2024-01-27
Hardware metrics (the preview PR)

xThaid:
edit: this PR contains all the changes to provide more context. I am going to split it into a few smaller ones that will be easier to review.

I thought it will be nice to have more observability in our core. It is good to know what's happening, what are causing delays, what the bottlenecks are. It also provides positive value for debugging - it can be easily noticed that something went wrong, when for example number of illegal instructions is bigger than expected.

The metrics can be optionally printed on the screen after the regression test is run (right now only for pysim backend):
image

Signals are also exposed in the gtkwave:
image

Benchmarks:
image

I added metrics in a few places to show how to use this feature. When this PR is ready to review, I will create another PR which adds metrics in the core.

Yet left to do:
- [x] Add support for cocotb simulator
- [x] Add documentation (after we agree on the interfaces etc)
- [x] Make sure that the core with metrics disabled synthesizes almost unchanged.  

Criticism is welcome

### DIFF
--- a/coreblocks/structs_common/hw_metrics.py
+++ b/coreblocks/structs_common/hw_metrics.py
@@ -1,0 +1,545 @@
+from dataclasses import dataclass, field
+from dataclasses_json import dataclass_json
+from typing import Optional
+from abc import ABC
+
+from amaranth import *
+from amaranth.utils import bits_for
+
+from transactron.utils import ValueLike
+from transactron import Method, def_method, TModule
+from transactron.utils import SignalBundle
+from transactron.lib import FIFO
+from coreblocks.params.genparams import GenParams
+from transactron.utils.dependencies import DependencyManager, ListKey
+
+
+@dataclass_json
+@dataclass(frozen=True)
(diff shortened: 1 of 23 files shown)


coreblocks/structs_common/hw_metrics.py
```diff
@@ -0,0 +1,320 @@
...
+
+            m.d.sync += self.cnt.value.eq(self.cnt.value + 1)
+
+        @def_method(m, self.incr_when)
+        def _(cond):
+            if not self.gen_params.hardware_metrics_enabled:
+                return
+
+            with m.If(cond):
+                m.d.sync += self.cnt.value.eq(self.cnt.value + 1)
```
tilk, line 89: How about defining only one method, and defining these as regular functions? They could even contain the `if` outside the actual method call, removing the call completely from a core without metrics.

lekcyjna123:
@xThaid can we close this PR as #580 and #578 have been merged?

Issue #560, opened 2024-01-08
Port ExtendedEncoder from #395


lekcyjna123:
In #395 there is implemented ExtenderEncoder. It is needed in new LSU, so port it to master branch.

Issue #559, opened 2024-01-08
Port CAM from #395


lekcyjna123:
In #395 there is implemented CAM. Port it to master.

PR #573, opened 2024-01-29
Port ContentAddressableMemory from #395

lekcyjna123:
Yet another PR to port something from #395. This time I ported CAM, which was requested by @makz00 for the LSU implementation. To port CAM few other things were needed:
- MultiPriorityEncoder
- generate_based_on_layout
- SimpleLayout

Resolves: #560
Resolves: #559

------
EDIT 03.05.2024:

I have uploaded a new set of commits with the refactored functionalities, changes:
- Added CAM
- Added MultiPriorityEncoder
- Added support for hypothesis
- Bump flake8 and black to remove false positives from 3.12
- Fix few formatting issues because of flake8 update


transactron/utils/amaranth_ext/elaboratables.py
```diff
@@ -239,3 +240,63 @@ def elaborate(self, platform):
...
+        self.input = Signal(self.input_width)
+        self.outputs = [Signal(range(self.input_width), name="output") for _ in range(self.outputs_count)]
+        self.valids = [Signal(name="valid") for _ in range(self.outputs_count)]
+
+    def elaborate(self, platform):
+        m = Module()
+
+        current_outputs = [Signal(range(self.input_width)) for _ in range(self.outputs_count)]
+        current_valids = [Signal() for _ in range(self.outputs_count)]
+        for j in reversed(range(self.input_width)):
```
tilk, line 282: This leads to a very inefficient priority encoder. Recursive construction is preferred.

transactron/lib/storage.py
```diff
@@ -133,3 +133,76 @@ def _(arg):
             m.d.comb += assign(write_args, arg, fields=AssignType.ALL)
 
         return m
+
+
+class ContentAddressableMemory(Elaboratable):
+    """Content addresable memory
+
+    This module implements a transactorn interface for the content addressable memory.
```
tilk, line 141:
```suggestion
    This module implements a Transactron interface for a content-addressable memory.
```

tilk: Does it make sense to call it an interface, when there is no internal module which has a traditional interface?

Shouldn't you write a few words to explain what a content addressable memory is? Also, this is a very specialized kind of content-addressable memory - no non-destructive reads, no writes.

transactron/lib/storage.py
```diff
@@ -133,3 +133,76 @@ def _(arg):
...
+        self.entries_number = entries_number
+
+        self.pop = Method(i=[("addr", self.address_layout)], o=[("data", self.data_layout), ("not_found", 1)])
+        self.push = Method(i=[("addr", self.address_layout), ("data", self.data_layout)])
+
+    def elaborate(self, platform) -> TModule:
+        m = TModule()
+
+        address_array = Array([Record(self.address_layout) for _ in range(self.entries_number)])
+        data_array = Array([Record(self.data_layout) for _ in range(self.entries_number)])
```
tilk, line 178: The data could be stored in a memory.

test/common/input_generation.py
```diff
@@ -0,0 +1,17 @@
+import random
+from typing import Optional
+from transactron.utils import SimpleLayout
+
+
+def generate_based_on_layout(layout: SimpleLayout, *, max_bits: Optional[int] = None):
```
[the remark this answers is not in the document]
xThaid: Another layout type is quite confusing. I agree that we should just throw an exception when the layout is incompatible.

transactron/lib/storage.py
```diff
@@ -133,3 +133,76 @@ def _(arg):
...
+        valids = Signal(self.entries_number, name="valids")
+
+        m.submodules.encoder_addr = encoder_addr = MultiPriorityEncoder(self.entries_number, 1)
+        m.submodules.encoder_valids = encoder_valids = MultiPriorityEncoder(self.entries_number, 1)
+        m.d.comb += encoder_valids.input.eq(~valids)
+
+        @def_method(m, self.push, ready=~valids.all())
+        def _(addr, data):
+            id = Signal(range(self.entries_number), name="id_push")
+            m.d.comb += id.eq(encoder_valids.outputs[0])
```
xThaid, line 188: `m.d.top_comb` can be used.

transactron/lib/storage.py
tilk: And also, no possibility of adding an eviction mechanism, for CAMs used as a cache.

tilk:
This could be nice to have in Transactron (I see some use cases for it), but the work has stopped. :(

lekcyjna123:
I was able to run first test using `hypothesis` sadly this library is very fragile to the order of drawing arguments, so we have to provide deterministic order between different runs of tests, what as we all know is very hard in pysim. I have to think what this problem should be fixed.

lekcyjna123:
I have prepared different solution which should be less prone to the order of processes because all data are generated before test execution (so only result has to be deterministic). Normal test execution is fast (no observable difference to the solution with randoms), but the case reduction can take a lot of time (45s). 

Below are two example messages with found failing examples. They are saved locally in `.hypothesis` directory, so each next run is fast.

obraz
obraz

transactron/lib/storage.py
```diff
@@ -133,3 +133,76 @@ def _(arg):
...
+
+    def elaborate(self, platform) -> TModule:
+        m = TModule()
+
+        address_array = Array([Record(self.address_layout) for _ in range(self.entries_number)])
+        data_array = Array([Record(self.data_layout) for _ in range(self.entries_number)])
+        valids = Signal(self.entries_number, name="valids")
+
+        m.submodules.encoder_addr = encoder_addr = MultiPriorityEncoder(self.entries_number, 1)
+        m.submodules.encoder_valids = encoder_valids = MultiPriorityEncoder(self.entries_number, 1)
```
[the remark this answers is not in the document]
lekcyjna123: Yes. But PriorityEncoders are going to be removed from `amaranth.lib` and `MultiPriorityEncoder` is a more generic solution which will be useful in future.

test/common/input_generation.py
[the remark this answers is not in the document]
lekcyjna123: Rewritten

transactron/lib/storage.py
lekcyjna123: After change of interface I decided to not use the `Memory` because it would need 3 ports.

transactron/utils/amaranth_ext/elaboratables.py
```diff
@@ -237,3 +238,77 @@ def elaborate(self, platform):
         m.d.sync += self.valid.eq(self.requests.any())
 
         return m
+
+
+class MultiPriorityEncoder(Elaboratable):
```
xThaid, line 243: I was thinking that since the priority encoder is just a combinational logic, it would be nice if we could use it like:
```python
idx, valid = prio_encoder(m, one_hot_signal)
```
instead of the boilerplate code
```python
m.submodules.prio_encoder = prio_encoder = PriorityEncoder(cnt)
m.d.av_comb += prio_encoder.i.eq(one_hot_singal)
idx = prio_encoder.o
valid = ~prio.encoder.n
```

xThaid: And another thing. It would be great if we could specify the bit order of the priority encoder. In theory by having one type of the priority encoder, we could get the other type by reversing the input and subtracting the outputs, but that requires a few adders.

lekcyjna123: I have added `create_priority_encoder` to add the automatic connections. Please see what you think about it.

Regarding to the different bit ordering I don't see a reason why we should add it, because it can be handled without subtraction. You pass the reversed array and you use indexes returned by encoder in this reversed array.

tilk: > I have added create_priority_encoder to add the automatic connections.

First, that's a mouthful. Second, a shorthand for creating `outputs_count=1` encoders would be nice, as the return type of the function complicates deconstruction.

transactron/testing/infrastructure.py
```diff
@@ -259,19 +253,39 @@ def f():
...
+            with self.configure_profiles():
+                self.configure_logging()
+                yield
+        self._transactron_hypothesis_iter_counter += 1
+
+    @pytest.fixture(autouse=True)
+    def fixture_initialize_testing_env(self, request):
+        self._transactron_hypothesis_iter_counter = 0
+        self._transactron_base_output_file_name = ".".join(request.node.nodeid.split("/"))
+        self._transactron_current_output_file_name = self._transactron_base_output_file_name
```
tilk, line 285: The logic behind `_transactron_current_output_file_name` deserves a comment.

transactron/utils/amaranth_ext/elaboratables.py
```diff
@@ -237,3 +238,131 @@ def elaborate(self, platform):
...
+        self.outputs = [Signal(range(self.input_width), name=f"output_{i}") for i in range(self.outputs_count)]
+        self.valids = [Signal(name=f"valid_{i}") for i in range(self.outputs_count)]
+
+    @staticmethod
+    def create_priority_encoder(
+        m: Module, input_width: int, input: ValueLike, outputs_count: int = 1, name: Optional[str] = None
+    ) -> list[tuple[Signal, Signal]]:
+        """Syntax sugar for creating MultiPriorityEncoder
+
+        This staticmethod allows to use MultiPriorityEncoder in a more functional
```
tilk, line 281:
```suggestion
        This static method allows to use MultiPriorityEncoder in a more functional
```

transactron/utils/amaranth_ext/elaboratables.py
lekcyjna123: I have updated the name to the `create` additionally I have added `create_simple` as a special case for `outputs_count=1`.

transactron/testing/infrastructure.py
lekcyjna123: Added.

tilk (approved):

transactron/lib/storage.py
```diff
@@ -138,6 +138,103 @@ def _(arg):
         return m
 
 
+class ContentAddressableMemory(Elaboratable):
+    """Content addresable memory
+
+    This module implements a content-addressable memory (in short CAM) with Transactron interface.
+    CAM is a type of memory where instead of predefined indexes there are used values feed in runtime
```
xThaid, line 145: typo: 'feed' -> 'fed'

transactron/lib/storage.py
```diff
@@ -138,6 +138,103 @@ def _(arg):
...
+    as keys (smimlar as in python dictionary). To insert new entry a pair `(key, value)` has to be
```
xThaid, line 146: typo: 'smimlar'

xThaid (approved):

PR #576, opened 2024-01-31
Update amaranth version to the v0.4.x head

lekcyjna123:
In #357 we need to use the `v0.4.x` head to have fix for traces which was delivered [here](https://github.com/amaranth-lang/amaranth/commit/1506f08b81fab25087bb449c96d276d294617576), because we are not ready yet to use the master branch. Till proper updates for master branch will be ready lets use the backport branch where this fix was also delivered in [that commit](https://github.com/amaranth-lang/amaranth/commit/94c504afc7d81738ecdc9523a2615ef43ecbf51a).

PR #577, opened 2024-01-31
A step towards amaranth `v0.5.0`

lekcyjna123:
In Amaranth `v0.5.0` there are some backward incompatible changes, to which we aren't align:
- RFC17
- RFC18
- RFC20
- RFC39
- MemoryInstance breaks Monkey patching
- `amaranth` modules names

This PR fixes issues which we have after these RFC's and updates used amaranth to the newest commit on `main`. This PR deprecates #576 if merged before.


lekcyjna123:
Updated monkey patching in `transactron/tracing.py` to take into account the https://github.com/amaranth-lang/amaranth/pull/1039.

lekcyjna123:
One more fix was needed. Recently there was a change in internal amaranth module namings. There were added underscore before them, to mark them private... 

(https://github.com/amaranth-lang/amaranth/commit/cf83193bf988fa9076cbffeb5da0972bdc40a233)

tilk:
> One more fix was needed. Recently there was a change in internal amaranth module namings. There were added underscore before them, to mark them private...
> 
> ([amaranth-lang/amaranth@cf83193](https://github.com/amaranth-lang/amaranth/commit/cf83193bf988fa9076cbffeb5da0972bdc40a233))

Didn't you go too keyboard-happy with these changes? You changed the import to an underscored module even when a non-underscored one exports all the imported names.

stubs/amaranth/vendor/_gowin.pyi
```diff
@@ -0,0 +1,98 @@
+"""
+This type stub file was generated by pyright.
+"""
```
tilk, line 3: Do we need auto-generated vendor stubs?

stubs/amaranth/vendor/__init__.pyi
```diff
@@ -0,0 +1,20 @@
...
+from ._lattice_machxo_2_3l import LatticeMachXO2Or3LPlatform as LatticeMachXO2Platform
+from ._lattice_machxo_2_3l import LatticeMachXO2Or3LPlatform as LatticeMachXO3LPlatform
+from ._quicklogic import QuicklogicPlatform
+from ._xilinx import XilinxPlatform
+
+__all__ = ["GowinPlatform", "IntelPlatform", "LatticeECP5Platform", "LatticeICE40Platform", "LatticeMachXO2Platform", "LatticeMachXO3LPlatform", "QuicklogicPlatform", "XilinxPlatform"]
+def __dir__(): # -> list[str]:
+    ...
+
+def __getattr__(name): # -> type[GowinPlatform] | type[IntelPlatform] | type[LatticeECP5Platform] | type[LatticeICE40Platform] | type[LatticeMachXO2Or3LPlatform] | type[QuicklogicPlatform] | type[XilinxPlatform]:
```
tilk, line 18: This doesn't need to be in the stub, it's actually counterproductive to do so.

coreblocks/params/instr.py
```diff
@@ -1,6 +1,6 @@
 from abc import abstractmethod, ABC
 
-from amaranth.hdl.ast import ValueCastable
+from amaranth.hdl._ast import ValueCastable
```
tilk, line 3:
```suggestion
from amaranth.hdl import ValueCastable
```

transactron/tracing.py
```diff
@@ -4,15 +4,19 @@
 
 import warnings
 
-from amaranth.hdl.ir import Elaboratable, Fragment, Instance
-from amaranth.hdl.xfrm import FragmentTransformer
-from amaranth.hdl import dsl, ir, mem, xfrm
+from amaranth.hdl._ir import Elaboratable, Fragment, Instance
```
tilk, line 7:
```suggestion
from amaranth.hdl import Elaboratable, Fragment, Instance
```

stubs/amaranth/hdl/__init__.pyi
```diff
@@ -2,12 +2,12 @@
...
+from ._ast import Array, C, Cat, ClockSignal, Const, Mux, Repl, ResetSignal, Shape, Signal, Value, signed, unsigned
+from ._dsl import Module
+from ._cd import ClockDomain
+from ._ir import Elaboratable, Fragment, Instance
+from ._mem import Memory
 from .rec import Record
-from .xfrm import DomainRenamer, EnableInserter, ResetInserter
+from ._xfrm import DomainRenamer, EnableInserter, ResetInserter
 
 __all__ = ["Shape", "unsigned", "signed", "Value", "Const", "C", "Mux", "Cat", "Repl", "Array", "Signal", "ClockSignal", "ResetSignal", "Module", "ClockDomain", "Elaboratable", "Fragment", "Instance", "Memory", "Record", "DomainRenamer", "ResetInserter", "EnableInserter"]
```
tilk, line 13: This list should be updated.

whitequark:
Just to be clear, every name (module, class, variable) with a leading underscore (under `hdl` and elsewhere) is private and the contents will be changed without notice or any consideration towards downstream code (other than in very exceptional circumstances). There are multiple current and future large-scale refactorings planned that include merging, splitting, and renaming those modules.

The public API is exported from `amaranth.hdl`. Soon there will also be an exhaustive reference for every name in the public API detailing its complete contract and compatibility guarantees, but that is not yet ready.

transactron/utils/_typing.py
```diff
@@ -18,8 +18,8 @@
 from enum import Enum
 from amaranth import *
 from amaranth.lib.data import StructLayout, View
-from amaranth.hdl.ast import ShapeCastable, Statement, ValueCastable
-from amaranth.hdl.dsl import _ModuleBuilderSubmodules, _ModuleBuilderDomainSet, _ModuleBuilderDomain, FSM
+from amaranth.hdl._ast import ShapeCastable, Statement, ValueCastable
```
whitequark, line 21:
```suggestion
from amaranth.hdl import ShapeCastable, ValueCastable
from amaranth.hdl._ast import Statement
```

whitequark:
I gave a careful look to all of your remaining imports of private functionality.

```py
from amaranth.hdl._ast import Statement, Assign
```

These will probably eventually be exported as-is. There is some doubt over the exact name and some edge cases of the functionality, but no large scale breakage is expected before 1.0.

```py
from amaranth.hdl._ast import ArrayProxy, Slice
```

These are not exported because the `Value` methods all return just `Value` (this will eventually be visible in the type annotations), and AST introspection is not supported as the exact classes are not stable. You're only using them minimally in tests, so should be fine for you.

```py
# transactron/tracing.py
from amaranth.hdl._xfrm import FragmentTransformer
new_obj._tracing_original = obj  # type: ignore
```

I am in awe of the ingenuity being shown in this file. Fragment transformers are definitely going away as soon as we can kill them (probably by the end of the 0.6 development cycle); they are a bad abstraction and also extremely inefficient in implementation, sometimes causing exponential blowups due to unsharing of ASTs that are DAGs rather than trees (which is perfectly fine in a netlist).

The specific functionality you're adding, if I understand it correctly, is something that has been desirable for a while and it may eventually become a part of core Amaranth. I would like to have a conversation about your needs so that these are taken into account when implementing the final functionality.

```py
# transactron/core.py
from amaranth.hdl._dsl import FSM, _ModuleBuilderDomain, ...
```

It is very difficult for me to understand exactly what is being implemented here in detail. The internals of the Amaranth `Module` DSL will be gradually improved, but probably not in immediate future and probably in ways that will let you continue the current approach.

The way you're using the `Module` API is definitely not something that will ever be supported. However, you may find it feasible, or perhaps even preferable, to implement the transformations on top of NIR (https://github.com/amaranth-lang/amaranth/pull/1102) and/or scopes (upcoming feature; currently in design phase, projected availability by the 0.6 development cycle). The former allows structured manipulation on netlist level, the latter on AST level.

I have also considered reengineering the `Module` internals to make it an entry point for DSL composition, enabling custom control flow constructs to exist. However, I am not sure this is the way to go, chiefly because languages with extensible syntax rapidly splinter into incompatible and not mutually legible dialects, and also because the need for custom control flow is quite niche.

In any case, once scopes stabilize, you should be able to vendor in a copy of the `Module` DSL and tweak it to your heart's content, which I suspect might be the preferred solution in the end. It will then be like any other elaboratable, except with the AST constructed in an unusual way.

stubs/amaranth/vendor/_gowin.pyi
lekcyjna123: I left stubs for the Lattice. Rest has been removed.

whitequark:
> Using `top_comb` we are able to have better code locality without sacrificing the performance.

This is definitely something that scopes will enable.

> The `av_comb` is more tricky. It is used to do some combinational assignment regardless if the method have been called or not.

It's hard for me to say whether this is the best way to achieve the goal or not, at least without investing a significant amount of time into understanding your design.

In general, it sounds like once scopes are available, the best way forward for you is to basically copy the `Module`-with-`Scope` sources into your tree and adjust however you like. It's unclear exactly when `Scope` will become a stable public API but in any case it will be something that is broken less often, and easier to use, than patching `Module` like this.



> It collects data, which are next used by `transactron/graph.py` to create a graph.

The way you've uploaded this image is weird. I can see a thumbnail but if I try to download it or view at a higher resolution it breaks. Can you reupload?

lekcyjna123:
Previously I uploaded image using the Github, so if it is problematic, then here is a link to the OneDrive where I uploaded [original image](https://uniwroc-my.sharepoint.com/:i:/g/personal/309016_uwr_edu_pl/EX_AZAq4rq1EoVu0c4QhsRoBjhznagbyzkpXlSKulYP-bA?e=zkeKVD)

whitequark:
@lekcyjna123 Can you explain what the arrows, boxes, and ovals on the flow graph refer to?

whitequark:
Is this sufficient for your tracing needs? https://github.com/amaranth-lang/amaranth/pull/1129

lekcyjna123:
> @lekcyjna123 Can you explain what the arrows, boxes, and ovals on the flow graph refer to?

- Boxes represents elaboratables. On the top of box there is an instance name and class name of the Elaboratable.
- ovals are transactions and methods (in future we should make a difference in shape)
- arrows/lines connects transactions with methods which are called by the transaction, arrow head is added based on heuristic and should describe the data flow direction

lekcyjna123:
> Is that something you would adopt? It seems like AvoidedIf could use something similar, though I do not understand the semantics well enough to propose a concrete solution.

This is an interesting idea. But If I correctly see, we already use such pattern (a little bit overcomplicated, but still). Whats more currently we use the `_ModuleBuilderDomain` only for typing/verification so this isn't something which we require. And type in this case doesn't provide us any additional information, because  `_ModuleBuilderDomain` is only wrapper with specific getters and setters.

I have prepared a proposition #589, which adds in transactron an abstraction layer, which uses public interfaces to implement the same functionality without explicite `_ModuleBuilderDomain`. Now @tilk should take a look on the #589 and share his opinion about my proposition.

tilk:
I have approved #589.

stubs/amaranth/hdl/_mem.pyi
```diff
@@ -90,4 +110,10 @@ class DummyPort:
         ...
     
 
-
+class MemoryInstance(Fragment):
+    memory : Memory
+    read_ports :list[ReadPort]
+    write_ports : list[WritePort]
+    attrs : dict
+    def __init__(self, memory : Memory, read_ports : list[ReadPort], write_ports : list[WritePort]) -> None:
```
tilk, line 118:
```suggestion
    memory: Memory
    read_ports: list[ReadPort]
    write_ports: list[WritePort]
    attrs: dict
    def __init__(self, memory: Memory, read_ports: list[ReadPort], write_ports: list[WritePort]) -> None:
```

transactron/utils/gen.py
```diff
@@ -3,10 +3,13 @@
...
 from amaranth.back import verilog
-from amaranth.hdl import ir
-from amaranth.hdl.ast import SignalDict
+from amaranth.hdl import Fragment
 
 from transactron.lib.metrics import HardwareMetricsManager
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from amaranth.hdl._ast import SignalDict
```
tilk, line 12: Is there a reason for that?

lekcyjna123: In my opinion yes. It is to emphasize that we need this import only for typing and not for the work, as it is an internal amaranth class.

tilk: If so, why the lack of consistency - this is not the only place where a private import is used only for typing.

lekcyjna123: I have updated also other files from this review.

tilk (approved):

whitequark (commented):
LGTM

tilk:
In the interest of moving things forward, I'm merging this without a second review.

PR #578, opened 2024-01-31
Refactor & improvements of regression scripts

xThaid:
A few changes to the scripts running regression tests and benchmarks distilled from #572:

- Fixed small typos in doc strings in Transactron
- Removed `VERILOG_SOURCES` env flag from makefiles and now I set it in the python script.
- Added a serializable dataclass for passing info between benchmark scripts for type safety and extendability.
- The core simulators now return an object instead of a single flag, so more information can be passed back.
- Added stubs for metrics.
- The benchmark scripts now prints a table with results (I plan to add there metrics as well):
image

### DIFF
--- a/scripts/run_benchmarks.py
+++ b/scripts/run_benchmarks.py
@@ -8,6 +8,7 @@ import re
 import sys
 import os
 import subprocess
+import tabulate
 from typing import Literal
 from pathlib import Path
 
@@ -15,6 +16,7 @@ topdir = Path(__file__).parent.parent
 sys.path.insert(0, str(topdir))
 
 import test.regression.benchmark  # noqa: E402
+from test.regression.benchmark import BenchmarkResult  # noqa: E402
 from test.regression.pysim import PySimulation  # noqa: E402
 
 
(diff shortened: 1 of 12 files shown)


tilk (approved):

scripts/run_benchmarks.py
```diff
@@ -100,6 +105,38 @@ def run_benchmarks(benchmarks: list[str], backend: Literal["pysim", "cocotb"], t
...
+
+    columns = [header]
+    for benchmark_name, result in results.items():
+        ipc = result.instr / result.cycles
+
+        column = [benchmark_name, result.cycles, result.instr, ipc]
+
+        for metric_name in sorted(result.metric_values.keys()):
+            regs = result.metric_values[metric_name]
+            for reg_name in regs:
```
lekcyjna123, line 129: Are you sure that order will be exactly the same as in line 118? `result.metric_values[metric_name]` is a dict, and dictionaries according to my knowledge doesn't guarantee order of elements returned by iterator.

lekcyjna123 (approved):

scripts/run_benchmarks.py
lekcyjna123: Ok. So lets merge it.

PR #580, opened 2024-02-01
Hardware Metrics

xThaid:
The main part of PR #572, which adds the core classes for hardware metrics along with some unit tests.


coreblocks/structs_common/hw_metrics.py
```diff
@@ -0,0 +1,545 @@
...
+        """
+        Returns all metrics registered in the core.
+        """
+        if self._metrics is None:
+            self._metrics = self._collect_metrics()
+        return self._metrics
+
+    def get_register_value(self, metric_name: str, reg_name: str) -> Signal:
+        """
+        Returns the signal holding the register value of the given metric.
```
lekcyjna123, line 498: This doc string doesn't explain anything. How to get `metric_name` and `reg_name`? Only checking the docs strings I would suppose that they are partf of `fully_qualified_name`. But based on the implementation I suppose that this is wrong and `reg_name` is implementation defined (different for Counter, Histogram, Latency etc.)

coreblocks/structs_common/hw_metrics.py
```diff
@@ -0,0 +1,545 @@
...
+            metrics[metric.fully_qualified_name] = metric
+
+        return metrics
+
+    def get_metrics(self) -> dict[str, HwMetric]:
+        """
+        Returns all metrics registered in the core.
+        """
+        if self._metrics is None:
+            self._metrics = self._collect_metrics()
```
lekcyjna123, line 493: Are you sure that this function will be called after all metrics have been registered? I think that current implementation can be still too eager and someone can try to use metric value in an `elaborate` function while some metrics wouldn't be created yet.

coreblocks/structs_common/hw_metrics.py
lekcyjna123, line 1: I would prefer to remove all `gen_params` usage not related to `DependencyManager`. This would make the code more portable and allow to move it in future to `transactron` (after having the way to use the `DependencyManager` without `GenParams` -- the same problem as in #538)

test/structs_common/test_hw_metrics.py
```diff
@@ -0,0 +1,305 @@
...
+from coreblocks.structs_common.hw_metrics import HwCounter, HwExpHistogram, HardwareMetricsManager
+from coreblocks.params import GenParams
+from coreblocks.params.configurations import test_core_config
+from transactron import *
+from transactron.utils import DependencyManager
+
+from ..common import *
+
+
+class CounterInMethodCircuit(Elaboratable):
```
lekcyjna123, line 15: Why you can not use `SimpleTestCircuit`?

test/structs_common/test_hw_metrics.py
```diff
@@ -0,0 +1,305 @@
...
+            for _ in range(200):
+                call_now = random.randint(0, 1) == 0
+
+                if call_now:
+                    yield from m.method.call()
+                else:
+                    yield
+
+                val = yield m._dut.counter.count.value
+                self.assertEqual(val, called_cnt)
```
lekcyjna123, line 85: More logical in my opinion it would be to have that code before `if`, because you compare counter to the previous one.

coreblocks/structs_common/hw_metrics.py
```diff
@@ -0,0 +1,545 @@
...
+
+        m = TModule()
+
+        @def_method(m, self._incr)
+        def _():
+            m.d.sync += self.count.value.eq(self.count.value + 1)
+
+        return m
+
+    def incr(self, m: TModule):
```
lekcyjna123, line 199: Why you decided for such interface instead of allowing to the direct calls to the `_incr`?

test/structs_common/test_hw_metrics.py
```diff
@@ -0,0 +1,305 @@
...
+            sum = 0
+            count = 0
+
+            bucket_cnt = 8
+            buckets = [0] * bucket_cnt
+
+            for _ in range(500):
+                if random.randrange(3) == 0:
+                    value = random.randint(0, max_value)
+                    value = 6
```
lekcyjna123, line 177: Why you overwrite the chosen value?

lekcyjna123 (commented):
Overall looks good, but I am not sure if it isn't overly complicated.

coreblocks/structs_common/hw_metrics.py
```diff
@@ -0,0 +1,545 @@
...
+    """Hardware Exponential Histogram
+
+    Represents the distribution of sampled data through a histogram. A histogram
+    samples observations (usually things like request durations or queue sizes) and counts
+    them in configurable buckets. The buckets are of exponential size. For example,
+    a histogram with 5 buckets would have the following value ranges:
+    [0, 1); [1, 2); [2, 4); [4, 8); [8, 16).
+
+    Additionally, the histogram tracks the number of observations, the sum
+    of observed values, and the minimum and maximum values.
```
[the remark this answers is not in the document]
xThaid: we could, but it is more convenient to just look at one number. And it doesn't cost us much to have it

coreblocks/structs_common/hw_metrics.py
```diff
@@ -0,0 +1,545 @@
...
+            The value that will be added to the histogram
+        """
+
+        if not self.gen_params.hardware_metrics_enabled:
+            return
+
+        self._add(m, sample)
+
+
+class LatencyMeasurer(Elaboratable):
```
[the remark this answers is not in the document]
xThaid: It is just a wrapper over the histogram metric, not a metric per se.

test/structs_common/test_hw_metrics.py
xThaid: Whoops, should have removed that

coreblocks/structs_common/hw_metrics.py
tilk: The first statement of this function is the answer. When metrics are disabled, the calls are not made.

coreblocks/structs_common/hw_metrics.py
```diff
@@ -0,0 +1,545 @@
...
+
+
+class HwExpHistogram(Elaboratable, HwMetric):
+    """Hardware Exponential Histogram
+
+    Represents the distribution of sampled data through a histogram. A histogram
+    samples observations (usually things like request durations or queue sizes) and counts
+    them in configurable buckets. The buckets are of exponential size. For example,
+    a histogram with 5 buckets would have the following value ranges:
+    [0, 1); [1, 2); [2, 4); [4, 8); [8, 16).
```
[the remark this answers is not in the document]
xThaid: Done

coreblocks/structs_common/hw_metrics.py
xThaid: I need to rethink this. I am using `gen_params` to read the configuration of the core to see if the metrics can be enabled.

test/structs_common/test_hw_metrics.py
xThaid: It won't work. It takes one cycle to update the value of the underlying counter, so if the method is called in cycle `n`, the counter will be increased in cycle `n+1`. Thus in fact I need to compare it with the previous value of the register.

And I don't want to add another yield after the method call, because I want to test how the counter works when it is possibly called every cycle (without stalls).

I added a clarifying comment to the code.

coreblocks/structs_common/hw_metrics.py
xThaid: Ok, gave some clarification in the comment.

coreblocks/structs_common/hw_metrics.py
xThaid: If metrics are created in the constructions, it won't be a problem. If metrics are created in `elaborate`, it will be. But I think it applies to anything that uses `DependencyManager`

test/structs_common/test_hw_metrics.py
xThaid: Because metric methods are just wrappers and `SimpleTestCircuit` won't handle it.

coreblocks/structs_common/hw_metrics.py
tilk: Maybe the `DependencyManager` refactor from #581 could be used for that? This way, metrics would depend only on the `DependencyManager` and could be moved to the Transactron library.

tilk: An idea I had would be to add a `[Bool, Bool]` key which normally returns `False`, but if a `True` value is added, returns `True`. This would make metrics disabled by default, but simple to enable.

xThaid: Done

tilk (approved):

lekcyjna123 (approved):

lekcyjna123:
I have created #585 to fix pipeline problem.

PR #583, opened 2024-02-08
Fix missing requirements to synthesize core

piotro888:
We use `typing-extensions` in Transactron, and verilog core cannot be synthesized without it.


piotro888:
> Then why does the "Synthesize full core" workflow not fail?

It installed `requirements-dev.txt` and it was probably dependency of some other package. I changed core systhesis to use `requirements.txt`, to verify that core can be build with non-dev version.

> I see is for Self, which can be imported from typing in Python 3.11, the currently used version.

Changed. Pyright python version was incorrectly set to `3.10` and this was reported as error before

piotro888:
Why the checks are not running?

tilk (approved):

lekcyjna123 (approved):

PR #587, opened 2024-02-12
Add metrics to a few parts of Coreblocks

xThaid:
This PR adds a few counters to a few parts of the core.

Blocked by: #586


transactron/testing/infrastructure.py
```diff
@@ -218,7 +218,7 @@ def wrapper():
             return wrapper
 
         for k in dir(self):
-            if k.startswith("test"):
+            if k.startswith("test") or k == "setUp":
```
lekcyjna123, line 221: How about teardown?

xThaid: I could add it, but we are not using this function anyway.

tilk (approved):

PR #590, opened 2024-02-26
Assertions in cocotb

tilk:
This PR adds support for assertions in cocotb simulation.

TODO:
- [x] Test it.


transactron/utils/assertion.py
```diff
@@ -32,35 +34,26 @@ def assertion(value: Value, *, src_loc_at: int = 0):
         identify the failing assertion.
     """
     src_loc = get_src_loc(src_loc_at)
+    sig = Signal()
+    m.d.comb += sig.eq(value)
```
xThaid, line 38: Shall we latch the assertion signal here? My concern is that a combinatorial signal with 0 fanout will be either removed by Amaranth (when generating Verilog) or by Verilator. I haven't verified it though, so I am not sure if this is a problem.

test/regression/cocotb.py
```diff
@@ -157,6 +157,16 @@ def get_cocotb_handle(self, path_components: list[str]) -> ModifiableObject:
 
         return obj
 
+    async def assert_handler(self, clock):
+        clock_edge_event = FallingEdge(clock)
+
+        while True:
+            for assert_info in self.gen_info.asserts:
+                assert_val = self.get_cocotb_handle(assert_info.location)
+                assert assert_val.value
```
xThaid, line 166: How about including the source location of the assertion in the assert message?

tilk: Fixed.

transactron/utils/assertion.py
tilk: I have verified this, and it turns out that neither Amaranth nor Verilator removed the wire and assertions work as expected.

xThaid (approved):

lekcyjna123 (approved):

Issue #592, opened 2024-02-29
Fix B extension


piotro888:
Regression tests are failing for the following B extension instructions:

- [x] `clz`
- [x] `cpop`
- [x] `ctz` 
- [x] `orcb`
- [x] `rev8`
- [ ] `rol`
- [ ] `ror`
- [ ]  `rori`
- [x]  `sext.b`
- [x]  `sext.h`
- [x]  `zext.h`

https://productionresultssa1.blob.core.windows.net/actions-results/b6033def-fe7d-49ca-94fe-90b456585539/workflow-job-run-3ffad427-3fce-534e-df7b-6607fc34e12a/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-02-29T10%3A08%3A14Z&sig=FDIia%2FpzwXQOAy8YwzaIPK7dnOPzMZzJuTnezCOdE5c%3D&sp=r&spr=https&sr=b&st=2024-02-29T09%3A58%3A09Z&sv=2021-12-02

xThaid:
The link does not work for me.

piotro888:
correct link https://github.com/piotro888/coreblocks/actions/runs/8093851411/job/22117470063 , "view raw logs" may be needed if compare log is truncated

piotro888:
I found a problem with instruction decoding, working on fix

PR #593, opened 2024-02-29
Fix Zbb instructions en/decoding

piotro888:
UNARY_BIT_MANIPULATION_* had only Funct12 without Funct7 specified in instruction definition -> decoder was outputting 0 ad funct7 -> ALU was using fuct7 to decode instructions and failing, because it was always 0.

Instead of adding Funct7 to instruction in decoder definitions, I optimized it by moving one instruction to other OpType, so all of the UNARY_BIT_MANIPULATION_* can be decoded only by OpType + Funct3

Issue 2) There was an OpType collsion between FUs on rotation. Both ALU and SHIFT unit used BIT_MANIPULATION optype.

Fixes ~(~all~/part of)~ all #592 
~Test job is currently in progress~

### DIFF
--- a/coreblocks/frontend/instr_description.py
+++ b/coreblocks/frontend/instr_description.py
@@ -170,14 +170,15 @@ instructions_by_optype = {
         Encoding(Opcode.OP, Funct3.XNOR, Funct7.XNOR),
     ],
     OpType.UNARY_BIT_MANIPULATION_1: [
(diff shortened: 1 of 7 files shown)


piotro888:
Should fix everything now

piotro888:
Tested by #594

tilk (approved):

xThaid (approved):

PR #595, opened 2024-02-29
Debug logs

xThaid:
Inspired by recent PR with assertions, I made a draft of a small utility for logging various events. The usage would be simple - call
`log.debug(m, trigger, "message with formats, e.g. {:08x}", pc)` somewhere in the circuit and then during the simulation, a log message would be printed whenever `trigger` is high with the signals provided.

I think this could also supersede assertions - a specific level of logging (could be `error`) could basically raise an assertion whenever a trigger is high.

TODO:
- [ ] support for cocotb
- [ ] CLI flags for setting the level
- [ ] unify with assertions?


tilk:
Looks pretty cool - this makes another software development comfort available in hardware..

tilk:
An idea - maybe the log entries should also contain some sort of subsystem identifiers? This way, a given subsystem might make log entries at different log levels, and it would be possible to filter both on log levels and subsystems.

scripts/run_benchmarks.py
```diff
@@ -74,17 +74,18 @@ def run_benchmarks_with_cocotb(benchmarks: list[str], traces: bool) -> bool:
     return res.returncode == 0
 
 
-def run_benchmarks_with_pysim(benchmarks: list[str], traces: bool, verbose: bool) -> bool:
+def run_benchmarks_with_pysim(
+    benchmarks: list[str],
+    traces: bool,
+) -> bool:
```
lekcyjna123, line 80: Why this is multiline now, when there is less arguments?

lekcyjna123 (approved):

tilk (approved):

tilk:
Waiting on branch conflicts.

scripts/run_benchmarks.py
xThaid: Fixed.

transactron/testing/logging.py
```diff
@@ -0,0 +1,108 @@
...
+_sim_cycle = 0
+
+
+class _LogFormatter(logging.Formatter):
+    """
+    Log formatter to provide colors and to inject simulator times into
+    the log messages. Adapted from https://stackoverflow.com/a/56944256/3638629
+    """
+
+    magenta = "\033[0;35m"
```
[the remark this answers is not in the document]
xThaid: I don't want to add another dependency for such a simple functionality.

PR #596, opened 2024-03-01
Profiles for regression tests

tilk:
This PR adds support for generating transaction profiles from regression tests, both in pysim and cocotb. The profiler was refactored so that its code can be used both in pysim and cocotb.

By the way, it's unfortunate that benchmarks and riscof tests have some significant code duplication, it would be nice to clean up this mess one day.


transactron/profiler.py
```diff
@@ -29,6 +40,50 @@ class ProfileInfo:
     is_transaction: bool
 
 
+@dataclass
+class ProfileData:
```
xThaid, line 44: At first sight it was not very clear for me what the role of this class is. A single comment line or a name like `ProfileDescription` or `EntireProfileDescription` would make it clear that it only holds non-runtime information.

xThaid (approved):
I like it

transactron/profiler.py
```diff
@@ -29,6 +40,50 @@ class ProfileInfo:
...
+
+        method_map = MethodMap(transaction_manager.transactions)
+        cgr, _, _ = TransactionManager._conflict_graph(method_map)
+        get_id = IdGenerator()
+
+        def local_src_loc(src_loc: SrcLoc):
+            return (os.path.relpath(src_loc[0]), src_loc[1])
+
+        for transaction in method_map.transactions:
+            transactions_and_methods[get_id(transaction)] = ProfileInfo(
```
lekcyjna123, line 65: Maybe a comment should be added to point that we encode transactions and methods to ints so that `CycleProfile` can be more efficient?

lekcyjna123 (approved):

transactron/profiler.py
```diff
@@ -29,6 +40,50 @@ class ProfileInfo:
...
+            transactions_by_method[get_id(method)] = [
+                get_id(t_or_m) for t_or_m in method_map.transactions_by_method[method]
+            ]
+
+        for transaction, transactions in cgr.items():
+            transaction_conflicts[get_id(transaction)] = [get_id(transaction2) for transaction2 in transactions]
+
+        return (
+            ProfileData(transactions_and_methods, method_parents, transactions_by_method, transaction_conflicts),
+            get_id,
```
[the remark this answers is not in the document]
tilk: Because it's a stateful object which doesn't need to be serialized, because it is not used during simulation.

transactron/profiler.py
tilk: We also encode transactions and methods to ints so that the required information about transactions can be JSON-serialized. JSON serialization doesn't preserve object identities.

tilk:
Stumbled upon an issue with pytest (#603). I can't really check if my changes still work, so I'm not merging yet. @lekcyjna123, could you take a look?

PR #597, opened 2024-03-05
Remove TestCoreSimple and TestCoreRandomized

xThaid:
They were useful at the beginning when we didn't have other means to test the whole core. Now they are not that useful.


tilk:
There is a fetcher change piggybacked here - but it also consists of removing stuff.

tilk (approved):

lekcyjna123 (approved):

PR #608, opened 2024-03-06
Add CI runs for #604

### LINKED ISSUE #605: Extend CI with more tests of run_test script
#604 wasn't caught in CI because not all options are tested in CI, so let add more testing to doesn't repeat the failure in future.

lekcyjna123:
#604 wasn't caught before merge so here is an extension to CI, which should catch such error in future.

Resolves #605

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -249,6 +249,12 @@ jobs:
       - name: Check traces and profiles
         run: ./scripts/run_tests.py -t -p -c 1 TestCore
 
+      - name: Check regression with pysim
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -220,6 +220,10 @@ jobs:
           . venv/bin/activate
           scripts/run_tests.py -a regression
 
+      - name: Check regression with pysim
+        run: ./scripts/run_tests.py -c 1 -a -b pysim regression
```
tilk, line 224: Activate missing.

tilk (approved):

piotro888 (approved):

PR #609, opened 2024-03-07
Migration to amaranth.lib.wiring

tilk:
This PR eliminates the last place where we use Amaranth's `Record` - namely, in bus interfaces.

* [x] Migrate Wishbone classes
* [x] Migrate AXI classes
* [x] Fix the core
* [x] Make the tests pass
* [x] Fix core synthesis
* [x] Fix docstrings


coreblocks/peripherals/axi_lite.py
```diff
@@ -24,65 +26,116 @@ def __init__(self, *, data_width: int = 64, addr_width: int = 64):
...
-        (if false it's generatd for the slave side)
-
-    Attributes
-    ----------
-    axil_layout: Record
-        Record of a AXI-Lite bus.
+        Patameters used to generate AXI-Lite signature
     """
 
     def __init__(self, axil_params: AXILiteParameters, *, master: bool = True):
```
lekcyjna123, line 38: Is `master` still needed?

tilk: It's not.

coreblocks/peripherals/wishbone.py
```diff
@@ -307,17 +302,17 @@ def _(arg) -> None:
...
     Parameters
     ----------
-    master_wb: Record (like WishboneLayout)
-        Record of master inteface.
-    slaves: List[Record]
-        List of connected slaves' Wishbone Records (like WishboneLayout).
+    master_wb: WishboneInterface
+        Master inteface.
+    slaves: list of WishboneInterface
+        List of connected slaves' Wishbone interfaces.
```
piotro888, line 315: Interfaces should be moved to `Attributes` section. Same in `WishboneArbiter`

coreblocks/core.py
```diff
@@ -38,7 +39,7 @@
 
 
 class Core(Elaboratable):
-    def __init__(self, *, gen_params: GenParams, wb_instr_bus: WishboneBus, wb_data_bus: WishboneBus):
+    def __init__(self, *, gen_params: GenParams, wb_instr_bus: WishboneInterface, wb_data_bus: WishboneInterface):
```
piotro888, line 42: Maybe `Core` should also be a `Compontent`?
But it is probably not worth changing now, `Core` should be generalized soon to support both Wishbone and AXI, that would overwrite that.

piotro888 (approved):
LGTM

coreblocks/peripherals/wishbone.py
tilk: Indeed, I have forgot about that.

coreblocks/core.py
tilk: That was my reasoning for not making the Core a Component.

coreblocks/peripherals/wishbone.py
```diff
@@ -354,10 +363,9 @@ def elaborate(self, platform):
 
         for i in range(len(self.slaves)):
             # connect all M->S signals except stb
-            m.d.comb += self.master_wb.connect(
-                self.slaves[i],
-                include=["dat_w", "rst", "cyc", "lock", "adr", "we", "sel"],
-            )
+            # workaround for the lack of selective connecting in wiring
+            for n in ["dat_w", "cyc", "lock", "adr", "we", "sel", "stb"]:
+                m.d.comb += getattr(self.slaves[i], n).eq(getattr(self.master_wb, n))
```
whitequark, line 368: The lack of `exclude` is *very* intentional, but you can always override a connection made by `connect()` afterwards with a normal `m.d.comb +=`, which gives you basically the same functionality here.

coreblocks/peripherals/wishbone.py
```diff
@@ -367,33 +375,49 @@ def elaborate(self, platform):
         m.d.comb += self.master_wb.rty.eq(reduce(operator.or_, [self.slaves[i].rty for i in range(len(self.slaves))]))
         for i in OneHotSwitchDynamic(m, self.txn_sel):
             # mux S->M data
-            m.d.comb += self.master_wb.connect(self.slaves[i], include=["dat_r", "stall"])
+            # workaround for the lack of selective connecting in wiring
+            for n in ["dat_r", "stall"]:
+                m.d.comb += getattr(self.master_wb, n).eq(getattr(self.slaves[i], n))
```
whitequark, line 380: This could be two lines instead of three :)

```
m.d.comb += self.master_wb.dat_r.eq(self.slaves[i].dat_r)
m.d.comb += self.master_wb.stall.eq(self.slaves[i].stall)
```

PR #616, opened 2024-03-11
Enable division and bit manipulation extensions in benchmarks

### LINKED ISSUE #613: Use division and bit extensions in IPC benchmarks
Currently the benchmarks are compiled with the C and Zmmul extensions only.

xThaid:
Blocked until the new image is in the package registry.

Fixes #613.

I ran the benchmarks here: https://github.com/xThaid/coreblocks/actions/runs/8239973106/job/22534443637

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -63,7 +63,7 @@ jobs:
   build-perf-benchmarks:
     name: Build performance benchmarks
     runs-on: ubuntu-latest
-    container: ghcr.io/xthaid/riscv-toolchain:2024.03.12
+    container: ghcr.io/kuznia-rdzeni/riscv-toolchain:2024.03.12
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -63,7 +63,7 @@ jobs:
   build-perf-benchmarks:
     name: Build performance benchmarks
     runs-on: ubuntu-latest
-    container: ghcr.io/kuznia-rdzeni/riscv-toolchain:2024.03.11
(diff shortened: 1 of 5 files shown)


tilk:
The source of the riscv-arch-test test failure is actually in the building workflow:
```
 /__w/coreblocks/coreblocks/test/external/riscof/riscv-arch-test/riscv-test-suite/rv32i_m/I/src/jalr-01.S:72: Error: illegal operands `la x0,5b'
```
This looks unexpected.

xThaid:
I updated the toolchain, because I was getting errors when tried to compile the toolchain with Zb* extensions.

.github/workflows/benchmark.yml
```diff
@@ -63,7 +63,7 @@ jobs:
   build-perf-benchmarks:
     name: Build performance benchmarks
     runs-on: ubuntu-latest
-    container: ghcr.io/kuznia-rdzeni/riscv-toolchain:2023.11.19_v
+    container: ghcr.io/xthaid/riscv-toolchain:2024.03.12
```
xThaid, line 66: I will revert it to `ghcr.io/coreblocks` when there is an updated image in our org registry.

tilk: Done.

tilk (approved):

lekcyjna123 (approved):

tilk:
It's pretty nice that the benchmark job now runs in 14 minutes instead of 16-17.

piotro888:
I reported the `jalr` test problem to `riscv-arch-tests` https://github.com/riscv-non-isa/riscv-arch-test/issues/444

cause in binutils: https://github.com/bminor/binutils-gdb/commit/27b33966b18ed8bf1701a60999448224b1d28273

PR #617, opened 2024-03-12
Split transactron core

lekcyjna123:
One more try to split `transactron/core.py`. I have started from scratch, because from the last one passed half a year. There was no changes in the code except imports and changing objects in types to type strings.


transactron/core/transaction.py
```diff
@@ -0,0 +1,120 @@
...
+
+if TYPE_CHECKING:
+    from .tmodule import TModule
+    from .manager import TransactionManager
+
+__all__ = ["Transaction", "TransactionManagerKey"]
+
+
+@dataclass(frozen=True)
+class TransactionManagerKey(SimpleKey["TransactionManager"]):
```
lekcyjna123, line 17: I am not sure if string in generic argument to `SimpleKey` is a feature, but it works.

tilk (approved):
That's not a bad split. I sure hope this PR foretells the end of the refactor season ;)

tilk:
This is related to #446.

piotro888 (commented):
The split structure also makes sense for me, everything is more readable now :)

EDIT:  Shouldn't `TransactionManagerKey` be placed  in `core/manager.py` instead of `core/transaction.py`?

For some bikeshedding:
I don't think that `eager_deterministic_cc_scheduler`, `trivial_roundrobin_cc_scheduler` and `TransactionManagerKey` shold be imported by default `from transactron import *` (if we care about default imports structure at all).

---

Not in the scope of this PR but before we separate transactron, I have two concerns about  lib and utils:
* _typing.py -> typing.py  - Underscore is misleading that it contains internal transactron types, while it has many useful types for Amaranth or use of transactron concepts externally in designs (also `# Internal Coreblocks types` :) )
* utils/transactron_helpers.py - is it supposed to contain functions used only internally in transactron? Now some of that functions look like that, but others (like functions for layouts) are used in Coreblocks and are useful for external use. I would split that to `utils/_transactron_utils.py` (with emphasized private underscore) and second file.

lekcyjna123:
> EDIT: Shouldn't TransactionManagerKey be placed in core/manager.py instead of core/transaction.py? 

That was my initial idea, but this cause cyclic dependency. `TransactionManager` needs `Transaction` and `Transaction` needs `TransactionManagerKey`.

Other possibility is to move `TransactionManagerKey` to separate file, but this would be overkill in my opinion.

lekcyjna123:
> I don't think that eager_deterministic_cc_scheduler, trivial_roundrobin_cc_scheduler and TransactionManagerKey 

Theoretically transactron was written with possibility to use user own instances of TransactionManager with chosen scheduler. That is the reason why I decided to export the symbols.

lekcyjna123:
One more proposition: https://github.com/kuznia-rdzeni/coreblocks/pull/617#pullrequestreview-1932978170

lekcyjna123:
> I would be for separate file then (maybe `transactron/core/keys.py` ?)

Moved.

> My point is that is should be imported via from transactron.core.schedulers import * and not pollute the from transactron import *. If implementing this is problematic or breaks documentation generation, current solution is fine then.

Done.

piotro888 (approved):
LGTM, thanks!

PR #621, opened 2024-03-18
Add selecting network circuit

xThaid:
I didn't call it sorting network, because in fact it doesn't sort -- we don't care about items that are not valid. We only select valid inputs and group them together.

The generated circuit looks nice (n=6):
https://digitaljs.tilk.eu/#2cde77f56b8cfd56a81defc25c4adc6257fc3f60f78f301a38ddc4a7d7330310


tilk (approved):

transactron/lib/connectors.py
```diff
@@ -275,3 +276,75 @@ def elaborate(self, platform):
...
+                b, cnt_b = current_level.pop(0)
+
+                total_cnt = Signal(max(len(cnt_a), len(cnt_b)) + 1)
+                m.d.comb += total_cnt.eq(cnt_a + cnt_b)
+
+                total_len = len(a) + len(b)
+                merged = Array(Signal(self.layout) for _ in range(total_len))
+
+                for i in range(len(a)):
+                    m.d.comb += merged[i].eq(Mux(cnt_a <= i, b[i - cnt_a], a[i]))
```
lekcyjna123, line 330: This generates a lot of subtractors. Why you doesn't use shifts?

transactron/lib/connectors.py
```diff
@@ -275,3 +276,75 @@ def elaborate(self, platform):
...
+
+    def elaborate(self, platform):
+        m = TModule()
+
+        current_level = []
+        for i in range(self.n):
+            current_level.append((Array([self.inputs[i]]), self.valids[i]))
+
+        # Create the network using the bottom-up approach.
+        while True:
```
lekcyjna123, line 314: Couldn't you write it as `while len(current_level) >= 2:` ?

lekcyjna123 (approved):

transactron/lib/connectors.py
xThaid: I don't think it is a problem. These are not proper substractions. In fact, the synthesis tool can permutate inputs of the mux and simply address it using `cnt_a`.

Shifting by variable number of bits in particular won't be better than a mux.

test/transactron/test_connectors.py
```diff
@@ -0,0 +1,46 @@
...
+from parameterized import parameterized_class
+
+from amaranth.sim import Settle
+
+from transactron.lib import StableSelectingNetwork
+from transactron.testing import TestCaseWithSimulator
+
+
+@parameterized_class(
+    ("n"),
```
whitequark, line 11: Should this be `("n",)`?

xThaid: Thanks for pointing this out - it should. However, apparently this decorator correctly handles this case:
```python
    if isinstance(attrs, string_types):
        attrs = [attrs]
```

PR #624, opened 2024-03-21
Support for superscalarity in the instruction cache

xThaid:
As the first step to the superscalarity, the instruction cache needs to return bigger blocks of memory (called Fetch Blocks).
I also renamed 'cache block' -> 'cache line' to avoid confusion with fetch blocks.


tilk (approved):

coreblocks/cache/refiller.py
```diff
@@ -23,51 +24,79 @@ def __init__(self, layouts: ICacheLayouts, params: ICacheParameters, bus_master:
...
+                    last=last_word,
+                )
+
+            word_counter_plus_one = Signal.like(word_counter)
+            m.d.av_comb += word_counter_plus_one.eq(word_counter + 1)
+
+            with m.If(last_word):
+                m.d.sync += refill_active.eq(0)
+            with m.Else():
+                request(Cat(word_counter_plus_one, cache_line_address))
```
lekcyjna123, line 84: Is this combinational path between transactions needed?

coreblocks/params/configurations.py
```diff
@@ -62,7 +62,7 @@ class CoreConfiguration:
         Associativity of the instruction cache.
     icache_sets_bits: int
         Log of the number of sets of the instruction cache.
-    icache_block_size_bits: int
+    icache_line_size_bits: int
         Log of the cache line size (in bytes).
```
lekcyjna123, line 66: The argument naming is inconsistent. In name you have `bits` and in docstring `bytes`.

coreblocks/params/configurations.py
```diff
@@ -87,7 +87,9 @@ class CoreConfiguration:
     icache_enable: bool = True
     icache_ways: int = 2
     icache_sets_bits: int = 7
-    icache_block_size_bits: int = 5
+    icache_line_size_bits: int = 5
+
+    fetch_block_bytes_log: int = 2
```
lekcyjna123, line 92: Maybe it will be simpler to pass number of bits in fetch block?

coreblocks/params/configurations.py
```diff
@@ -62,7 +62,7 @@ class CoreConfiguration:
         Associativity of the instruction cache.
     icache_sets_bits: int
         Log of the number of sets of the instruction cache.
-    icache_block_size_bits: int
+    icache_line_size_bits: int
         Log of the cache line size (in bytes).
     allow_partial_extensions: bool
         Allow partial support of extensions.
```
lekcyjna123, line 68: Add docstring for `fetch_block_bytes_log`.

coreblocks/cache/refiller.py
```diff
@@ -23,51 +24,79 @@ def __init__(self, layouts: ICacheLayouts, params: ICacheParameters, bus_master:
...
+                addr=request_address,
                 sel=C(1).replicate(self.bus_master.params.data_width // self.bus_master.params.granularity),
             )
 
+        refill_active = Signal()
+        cache_line_address = Signal(self.params.word_width - self.params.offset_bits)
+
+        word_counter = Signal(range(self.params.words_in_line))
+        block_buffer = Signal(self.params.word_width * (self.params.words_in_fetch_block - 1))
+        with Transaction().body(m):
```
lekcyjna123, line 48: It would be nice to have a comment that this transaction reads a couple of responses from bus and join them into one fetch block.

coreblocks/params/icache_params.py
```diff
@@ -11,35 +11,49 @@ class ICacheParameters:
         Associativity of the cache.
     num_of_sets_bits : int
         Log of the number of cache sets.
-    block_size_bits : int
-        Log of the size of a single cache block in bytes.
+    line_size_bits : int
```
lekcyjna123, line 14: If this is logarithm then maybe add `log` to the name?

coreblocks/cache/refiller.py
xThaid: It is not needed, but speeds things up. Now, both the request and the response transaction can be run at the same cycle.

coreblocks/params/configurations.py
xThaid: Ah, it got me as well. Apparently `bits` here really means "number of bits to write the number `icache_line_size`", in other words, it is just a log of the size.

It seems that we have the convention of using the word `bits` for denoting a logarithm, for example `phys_regs_bits`. I find it confusing a bit.

I renamed `icache_line_size_bits` -> `icache_line_bytes_log`.

coreblocks/params/configurations.py
xThaid: Having `fetch_block_bytes_log` naturally imposes two conditions:
- it must be divisible by 8 (because we want full bytes)
- it must be a power of 2

Otherwise, we would have to add code that verifies these.

coreblocks/params/icache_params.py
xThaid: See the comment above.

coreblocks/cache/refiller.py
lekcyjna123: But why we can not back pressure from `self.bus_master.request_read`? We can move the `word_counter` increasing outside of this trasaction. It can be increased after each `self.bus_master.request_read`. When `self.bus_master.request_read` is not ready then the transaction won't start and `word_counter` won't increase. We can have a separate `word_counter_recv` in transaction with `get_read_response`.

xThaid: I changed the refiller to make the request and the response transaction independent. This came at a cost of more complex circuit, as we need to now handle pipeline flushes, and requests are not forwarded to the bus controller, so refilling takes one more cycle.

lekcyjna123 (approved):

PR #631, opened 2024-03-27
Refactor RISC-V instruction models

xThaid:
The change in #627 was not fully working and while fixing it, I realized that the solution we had previously was not perfect. Based on the dependency we have removed, `riscv-model`, I made a refactor to avoid repeating code.

### DIFF
--- a/coreblocks/params/instr.py
+++ b/coreblocks/params/instr.py
@@ -1,14 +1,24 @@
-from abc import abstractmethod, ABC
+"""
+
+Based on riscv-python-model by Stefan Wallentowitz
+https://github.com/wallento/riscv-python-model
+"""
+
+from dataclasses import dataclass
+from abc import ABC
+from enum import Enum
+from typing import Optional
 
 from amaranth.hdl import ValueCastable
 from amaranth import *
 
-from transactron.utils import ValueLike, int_to_signed
+from transactron.utils import ValueLike
 from coreblocks.params.isa_params import *
 from coreblocks.frontend.decoder.isa import *
 
 
 __all__ = [
+    "RISCVInstr",
     "RTypeInstr",
     "ITypeInstr",
     "STypeInstr",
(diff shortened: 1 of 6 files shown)


coreblocks/params/instr.py
```diff
@@ -20,154 +30,171 @@
...
+    def __init__(self, opcode: ValueLike, **kwargs):
+        super().__init__(opcode=Cat(C(0b11, 2), opcode), **kwargs)
+
+
+class UTypeInstr(RISCVInstr):
+    field_rd = Field(name="rd", base=7, size=5)
+    field_imm = Field(name="imm", base=12, size=20, offset=12, signed=True)
+
+    def __init__(self, opcode: ValueLike, **kwargs):
+        super().__init__(opcode=Cat(C(0b11, 2), opcode), **kwargs)
```
tilk, line 182: This override is repeated multiple times. Can this be refactored to avoid code duplication?

coreblocks/params/instr.py
```diff
@@ -20,154 +30,171 @@
...
+
+    @classmethod
+    def get_fields(cls) -> list[Field]:
+        return [getattr(cls, member) for member in dir(cls) if member.startswith("field_")]
+
+    def encode(self) -> int:
+        const = Const.cast(self.as_value())
+        return const.value  # type: ignore
+
+    def __setattr__(self, key, value):
```
tilk, line 71: The generic-ness is cool and all, but it defeats typechecking and code completion. Could this be done with properties or descriptor objects instead?

coreblocks/params/instr.py
xThaid: Done

coreblocks/params/instr.py
xThaid: Descriptors seem to perfectly fit the use case here.

tilk (approved):

coreblocks/params/instr.py
```diff
@@ -20,154 +30,194 @@
 ]
 
 
+@dataclass(kw_only=True)
+class Field:
+    base: int | list[int]
+    size: int | list[int]
+
+    signed: bool = False
+    offset: int = 0
```
lekcyjna123, line 39: Please add doc string with information what is `offset`. I was confused while reading the code, because I though that this is an offset of field in instruction, but it is an offset of value saved in field.

coreblocks/params/instr.py
```diff
@@ -20,154 +30,194 @@
...
+    def __set_name__(self, owner, name):
+        self._name = name
+
+    def __get__(self, obj, objtype=None) -> Value:
+        if self.static_value is not None:
+            return self.static_value
+
+        return obj.__dict__.get(self._name, C(0, Shape(self.width(), self.signed)))
+
+    def __set__(self, obj, value) -> None:
```
lekcyjna123, line 62: Please add a note in doc string that in case when `offset!=0` the `value` should contains also bits which will be skipped because of offset (e.g. if offset=1 and size=2, and we would like to store `0b11` in field then `value` should be `0b110`.

coreblocks/params/instr.py
```diff
@@ -20,154 +30,194 @@
...
+        return sum(self.sizes())
+
+    def __set_name__(self, owner, name):
+        self._name = name
+
+    def __get__(self, obj, objtype=None) -> Value:
+        if self.static_value is not None:
+            return self.static_value
+
+        return obj.__dict__.get(self._name, C(0, Shape(self.width(), self.signed)))
```
lekcyjna123, line 60: You return here data with different shapes. If a default is returned it has width `self.width()` and if non-default value is returned it has `self.width() + self.offset` width (see `expected_shape` in `__set__`).

coreblocks/params/instr.py
```diff
@@ -20,154 +30,194 @@
...
+        for field in _get_fields(type(self)):
+            value = field.__get__(self, type(self))
+
+            base = field.bases()
+            size = field.sizes()
 
-    def shape(self):
+            offset = field.offset
+            for i in range(len(base)):
+                parts.append((base[i], value[offset : offset + size[i]]))
```
lekcyjna123, line 123: Parsing of `offset` shouldn't be the part of `Field` implementation?

coreblocks/params/instr.py
xThaid: Good catch. Fixed.

coreblocks/params/instr.py
xThaid: Could be. I moved the code.

lekcyjna123 (approved):

PR #632, opened 2024-03-27
More metrics for RF, RS and ROB

tilk:
This PR adds the following metrics:

* Histogram of the number of used rows in RF, RS and ROB in each cycle.
* Histogram of latency for instructions in RS.
* Histogram of times registers are valid in RF before being freed.

Conclusions from reading the metrics:
* ROB is currently very underused. Average ROB slots used is around 3-4, maximum is around 11-16. This will probably change with the introduction of jump prediction.
* RS latency is low, typically 1-2 cycles.

The PR also adds two modules: 

* `AsyncMemoryBank`  which basically wraps an Amaranth memory with asynchronous reads,
* `IndexedLatencyMeasurer` which allows to measure latency for things that are not processed in FIFO fashion, but have unique indexes (like RF and RS entries).

TODO:
* [x] Documentation.
* [x] Tests.


coreblocks/core_structs/rf.py
```diff
@@ -56,10 +74,22 @@ def _(reg_id: Value, reg_val: Value):
...
         @def_method(m, self.free)
         def _(reg_id: Value):
             with m.If(reg_id != 0):
                 m.d.sync += self.entries[reg_id].valid.eq(0)
+                self.perf_rf_valid_time.stop(m, slot=reg_id)
+
+        if self.perf_num_valid.metrics_enabled():
+            num_valid = Signal(self.gen_params.phys_regs_bits + 1)
+            m.d.comb += num_valid.eq(
+                reduce(
```
xThaid, line 88: `popcount(Cat([valid signals]))` can be used here.

coreblocks/func_blocks/fu/common/rs.py
```diff
@@ -105,4 +128,12 @@ def _(rs_entry_id: Value) -> RecordDict:
             def _() -> RecordDict:
                 return {"ready_list": ready_list}
 
+        if self.perf_num_full.metrics_enabled():
+            num_full = Signal(self.rs_entries_bits + 1)
+            m.d.comb += num_full.eq(
```
xThaid, line 133: `popcount` here as well

xThaid (approved):

test/transactron/test_metrics.py
```diff
@@ -305,6 +305,105 @@ def consumer():
...
+
+        latencies: list[int] = []
+
+        events = list(0 for _ in range(self.slots_number))
+        free_slots = list(k for k in range(self.slots_number))
+        used_slots: list[int] = []
+
+        time = 0
+
+        def ticker():
```
lekcyjna123, line 337: You can use `yield Now()` instead of creating separate process.

test/transactron/test_metrics.py
```diff
@@ -305,6 +305,105 @@ def consumer():
...
+
+        finish = False
+
+        def producer():
+            nonlocal finish
+
+            for _ in range(200):
+                if not free_slots:
+                    yield
+                    continue
```
lekcyjna123, line 354: This cause that we can have less than 200 iterations. Was this intentional?

test/transactron/test_metrics.py
```diff
@@ -305,6 +305,105 @@ def consumer():
...
+            self.assertEqual(max(latencies), (yield m._dut.histogram.max.value))
+            self.assertEqual(sum(latencies), (yield m._dut.histogram.sum.value))
+            self.assertEqual(len(latencies), (yield m._dut.histogram.count.value))
+
+            for i in range(m._dut.histogram.bucket_count):
+                bucket_start = 0 if i == 0 else 2 ** (i - 1)
+                bucket_end = 1e10 if i == m._dut.histogram.bucket_count - 1 else 2**i
+
+                count = sum(1 for x in latencies if bucket_start <= x < bucket_end)
+                self.assertEqual(count, (yield m._dut.histogram.buckets[i].value))
```
lekcyjna123, line 399: Could we do that common with FIFOLatencyMeasurer test?

coreblocks/func_blocks/fu/common/rs_func_block.py
```diff
@@ -41,10 +43,13 @@ def __init__(self, gen_params: GenParams, func_units: Iterable[tuple[FuncUnit, s
             Functional units to be used by this module.
         rs_entries: int
             Number of entries in RS.
+        rs_number: int
+            The number of this RS block. Used for debugging.
```
lekcyjna123, line 47: If used for debugging, maybe there should be a default value? So that if anyone doesn't care about debug feature, then it doesn't have to pass that value?

test/transactron/test_metrics.py
tilk: The test is written similar to to other metrics tests. I'm not going to do the change here without changing all the others. This is best left for a refactoring PR.

test/transactron/test_metrics.py
tilk: Not really. I guess changing that `if` to `while` should be enough.

test/transactron/test_metrics.py
tilk: Possibly, maybe that's a refactor worth doing.

coreblocks/func_blocks/fu/common/rs_func_block.py
tilk: I thought about it, and I don't think this is a good thing. If `rs_number` is not set in a given `CoreConfiguration` then the metric will be hard to read, so the person doing the metrics will need to change the configuration.

Probably it would be better to auto-generate these numbers.

tilk:
It seems like the test might still have a race. Those `Settle`s are really a nightmare.

lekcyjna123 (approved):

PR #635, opened 2024-03-29
DummyLSU as a FU

### LINKED ISSUE #591: More RS slots in DummyLSU load/store unit
The task is to implement multiple RS slots in DummyLSU, which will be processed in a FIFO fashion. This can improve the LSU performance because two consecutive load/store instructions won't stall the fetcher, increasing reordering opportunities. Correctness will not be affected, as the load/store instructions will still be executed in program order.

This is a simpler version of #437.

tilk:
Oops, I did it again. `DummyLSU` is now smaller, faster, and a FU. As this LSU is still dumb, `FifoRS` is used to maintain the instruction ordering. Still, having an actual RS for LSU allows for more latency hiding for memory requests.

Depends on #634. Fixes #591.

TODO:

* [x] Move `DummyLSU` to `fu` directory.
* [x] Fix tests.


test/lsu/test_dummylsu.py
```diff
@@ -180,7 +164,8 @@ def generate_instr(self, max_reg_val, max_imm_val):
                     }
                 )
 
-            self.exception_result.append(
+            print(f"addr:0x{addr:08x} misaligned:{misaligned} bus_err:{bus_err}")
```
lekcyjna123, line 167: Is this print intentional?

tilk: This is probably a leftover from debugging, will remove.

lekcyjna123 (approved):

xThaid (approved):

PR #638, opened 2024-04-01
Fix yet another bug in the instruction cache

xThaid:
The instruction cache code didn't correctly handle a case when the `start_refill` method of the cache refiller was not active. The tests didn't test it at all, as `start_refill` was always ready there.


tilk (approved):
I fail to understand both the bug and the fix, there are possibly some implicit connections in the code I just can't see. The cache code could benefit from some more documentation. Accepting to move things forward.

lekcyjna123:
If I correctly understand the bug was that in case when `start_refill` was not ready, then `MemRead` transaction wasn't started and this caused that `issue_req` could be called second time, what caused that data from first call were overwritten.

xThaid:
> If I correctly understand the bug was that in case when `start_refill` was not ready, then `MemRead` transaction wasn't started and this caused that `issue_req` could be called second time, what caused that data from first call were overwritten.

Exactly.

lekcyjna123 (approved):

Issue #629, opened 2024-03-26
Add support for array layouts in the method layout


xThaid:
We could leverage Amaranth's new aggregate data structure, `ArrayLayout`, to add support for method layouts that takes an array (of a fixed size) of arguments.

The syntax could be like:
```
layout = [
                ("instr", self.gen_params.isa.ilen, fetch_width),
                ("foobar", 1),
]
```
That is, if a layout field is a tuple of 3 elements, we treat it as an array where the last element of the tuple is the size of an array.

tilk:
Currently, an `ArrayLayout` should be possible to be used by just writing it in place of the bit width. I haven't tested this though.

xThaid:
> Currently, an `ArrayLayout` should be possible to be used by just writing it in place of the bit width. I haven't tested this though.

The layout itself works, but `assign` and perhaps something else need to be updated accordingly.

tilk:
Indeed, only `StructLayout` is currently supported in `assign` - support for other layout types needs to be added there.

xThaid:
Solved by #640

PR #640, opened 2024-04-02
Support ArrayLayouts in assign

tilk:
Related to #629.


xThaid:
Can we also add support for plain lists of signals, i.e.:
```python
layout = ...
s1 = data.View(data.ArrayLayout(layout, 4))
s2 = [Signal(layout) for _ in range(4)]

m.d.comb += assign(s1, s2)
```

xThaid (approved):

lekcyjna123 (approved):

PR #642, opened 2024-04-02
Drop `unittest` from `TestCaseWithSimulator`

lekcyjna123:
In #553 #554 #611 we have added support for the `pytest`. This caused that we are supporting two testing libraries in parallel, which adds code and conceptual complexity to coreblocks. Particularly I have hit in that problem when adding the `hypothesis` in #573, where I had to integrate it with both `unittest` and `pytest` at the same time.

This PR want to drop `unittest` support and left only `pytest`. The most important changes are in `TestCaseWithSimulator` class where much of code from `run_simulation` was moved to separate fixtures. Whats more it was possible to drop the dirty trick with wrapping functions in `DependencyContext`.

There is a little drawback. If we want to use `DependencyContext` in `setUp` function we have to add one more line:
```py
    @pytest.fixture(autouse=True)
    def setup(self, configure_dependency_context):
```
You can observe that in `test/fu/functional_common.py`. It is because we have to say that `configure_dependency_context` has to run before setup function.

ToDo:
- [x] Support for test execution
- [x] Support for profiles
- [x] Support for traces
- [x] Support for logging
- [x] Migrate some example tests
- [x] Migrate all tests to the new syntax

Please say what is your opinion about these changes? Are they ok for you and can I migrate all the tests?


piotro888:
Some unit tests still use `unitest.TestCase`, is that intentional? unittest is also used in scripts.

lekcyjna123:
Good catch. I will prepare a separate review for that to doesn't delay this one.

piotro888 (approved):

PR #644, opened 2024-04-03
Move tests after #620

lekcyjna123:
In #620 we have change coreblocks directory structure. This PR is a step towards aligning tests directory structure with coreblocks. Here I only move tests between directories to reduce number of merge conflicts with other PRs. In future some files with tests will be split to be more fain-grained.


tilk (approved):

xThaid (approved):

tilk:
Not merging due to conflicts.

PR #645, opened 2024-04-07
Fix logging handler connection

lekcyjna123:
In #642 I have found that capturing logs with `pytest` doesn't work, it was because of bug in our code, where we substitute all log handlers with our handler instead of appending it. Here is a small fix for that.


tilk (approved):

xThaid (approved):

xThaid:
Output is now doubled. For example run:
```
pytest --log-level=DEBUG --coreblocks-regression --coreblocks-backend pysim -s -v -k "rv32uc-rvc"
```

PR #647, opened 2024-04-09
Fix LSU FENCE issue

tilk:
The freshly refactored LSU could mess up ordering between FENCE and LOAD/STORE instructions, making LOAD instructions return invalid values if directly after FENCE.

### DIFF
--- a/coreblocks/func_blocks/fu/lsu/dummyLsu.py
+++ b/coreblocks/func_blocks/fu/lsu/dummyLsu.py
@@ -228,17 +228,18 @@ class LSUDummy(FuncUnit, Elaboratable):
         m.submodules.requester = requester = LSURequester(self.gen_params, self.bus)
 
         m.submodules.requests = requests = Forwarder(self.fu_layouts.issue)
-        m.submodules.results = results = self.forwarder = FIFO(self.lsu_layouts.accept, 2)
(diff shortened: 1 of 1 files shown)


xThaid (approved):
Seems that it fixed the error in #630

coreblocks/func_blocks/fu/lsu/dummyLsu.py
```diff
@@ -228,17 +228,18 @@ def elaborate(self, platform):
         m.submodules.requester = requester = LSURequester(self.gen_params, self.bus)
 
         m.submodules.requests = requests = Forwarder(self.fu_layouts.issue)
-        m.submodules.results = results = self.forwarder = FIFO(self.lsu_layouts.accept, 2)
+        m.submodules.results_noop = results_noop = self.forwarder = FIFO(self.lsu_layouts.accept, 2)
```
piotro888, line 231:
```suggestion
        m.submodules.results_noop = results_noop = FIFO(self.lsu_layouts.accept, 2)
```
`self.forwarder` is unused
`self.forwarder = FIFO` :)

piotro888 (approved):

coreblocks/func_blocks/fu/lsu/dummyLsu.py
tilk: And that's a severe case of perceptual blindness. ;)

PR #648, opened 2024-04-09
Turn off printing pytest logs

lekcyjna123:
After #645 logs are printed twice and I haven't found that in testing. This was caused by adding the pytest handler and the fact that we have had the pytest log printing turned on by default. This was a no-op without a handler so before #645 it wasn't a problem. Here I turn printing of pytest logs off.


pytest.ini
```diff
@@ -6,4 +6,4 @@ norecursedirs = '*.egg', '.*', 'build', 'dist', 'venv', '__traces__', '__pycache
 filterwarnings =
     ignore:cannot collect test class 'TestbenchIO':pytest.PytestCollectionWarning
     ignore:No files were found in testpaths:pytest.PytestConfigWarning:
-log_cli=true
+log_cli=false
```
xThaid, line 9: nit: this line can be removed altogether.

xThaid (approved):

PR #652, opened 2024-04-11
Flexible CSRs

piotro888:
Additions to the CSR register interface

PR #654, opened 2024-04-11
Internal Interrupt Controller

### LINKED ISSUE #599: Implement compliant internal interrupt controller
CSR internal mechanism compliant with RISC-V privileged spec (`mip`, `mie`, etc.).

Further work would include implementing (some of my finding on this mess):
* ACLINT (CLINT is now deprecated). ACLINT provides external memory mapped layer for `mtimecmp` - timer interrupts and IIRC for IPI (inter processor interrupts). ACLINT is backwards compatible with CLINT that (probably) is widely supported.
* SiFive CLIC looks like an alternative to ACLINT, that also partially replces internal CSR controller handler (but both modes must be supported). It is probably too complex and unnecessary for now. It is standarized by RISC-V now.
* Next layer of interrupt management (only external hardware interrupts) - PLIC also needs to be implemented to support Linux. Originally specified by SiFive, now standarized byt RISC-V too.

Some sources:
* https://www.starfivetech.com/uploads/sifive-interrupt-cookbook-v1p2.pdf
* https://github.com/riscv/riscv-platform-specs/blob/main/riscv-platform-spec.adoc#interrupts
* linux-on-vexriscv

piotro888:
Closes #599 

Implements priv spec core internal interrupt controller operating on CSRs. Details in docstrings.

Depends on #652 , needs rebase to reflect changes and clean-up from it after it is approved.


coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -1,44 +1,202 @@
...
+                "internal_report_level": In(ISA_RESERVED_INTERRUPTS),
+                "custom_report_level": In(gen_params.interrupt_custom_count),
+                "custom_report_edge": In(gen_params.interrupt_custom_count),
+            }
+        )
+
+
+class InternalInterruptController(Component):
+    """Core Internal Interrupt Controller
+    Compatiblie with RISC-V privileged specification.
```
xThaid, line 42: typo

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -1,44 +1,202 @@
...
-from coreblocks.interface.keys import AsyncInterruptInsertSignalKey, MretKey
+from coreblocks.interface.keys import AsyncInterruptInsertSignalKey, GenericCSRRegistersKey, MretKey
 
 from transactron.core import Method, TModule, def_method
+from transactron.core.transaction import Transaction
+from transactron.utils.dependencies import DependencyManager
+
+
+class InterruptCauseNumber(IntEnum):
+    SSI = 1
```
xThaid, line 18: Can we have these shortcuts expanded in comments? And a short comment if these are defined by the specification

coreblocks/priv/csr/aliased.py
```diff
@@ -0,0 +1,66 @@
...
+        # TODO: verify bounds
+
+    def elaborate(self, platform):
+        m = TModule()
+        self.elaborated = True
+
+        @def_method(m, self._fu_write)
+        def _(data: Value):
+            for start, csr in self.fields:
+                csr._fu_write(m, data.bit_select(start, csr.width))
```
xThaid, line 55: you can simply use array slicing here, e.g. `data[start : start + csr.width]`

coreblocks/priv/csr/aliased.py
```diff
@@ -0,0 +1,66 @@
...
+
+        @def_method(m, self._fu_read)
+        def _() -> Value:
+            read_data = []  # amaranth doesn't support assigning to bit_select
```
xThaid, line 59: I am not sure about `bit_select`, but it seems that you can assign to an array slice, for example in `alu.py:204`

piotro888:
The cause of the problem is `Forwarder` inside `FetchResume` `Unifier` key. It can illegally store resume call to next fetcher stall.

I'm still not sure how it happened that call with `resume_from_exception=0` was stored in forwarder instead of `resume_from_exception=1` that should be always be later (called from `Retirement`), and what is the best way to fix it.

xThaid:
I am not sure if I see where exactly the problem is. I looked at the execution of `TestCoreInterrupt_0_interrupt_asm` and it seems that the core works fine, at least from the perspective of the instruction stream it is given. I didn't dig into what it exactly executes and the test itself

Can you provide cycle at which there is a bug? And what is the expected behavior there?

piotro888:
@xThaid  Ok, I found it.
The problem:
1. Core is stalled due to CSR instruction
2. Misprediction happens
3. CSRUnit calls FetchResume with `resume_from_exception=0` that normally should be ignored, but Fetch is still doing something and `resume` method is not yet ready, so this call waits inside `Unifier` in `Forwarder`.
4. Core completes flushing, and `Retirement` tries to call `resume` with `resume_from_excpetion=1`, but the method blocks because fetcher is still not ready
5. Fetcher becomes ready, due to random priority `resume` call from `Retirement` wins with `Forwarder`.
6. Fetch is unstalled, and `resume` method is not ready again
7. Another unsafe instruction arrives
8. `Forwader` in `Unifier` still holds the old call value!!! When `resume` becomes ready, fetch is resumed to the old PC.

Setting priority between `Retirement.CORE_RESUME` and `Unifier`, to `Unifier.Forwarder` should fix the issue, so all calls with `resume_form_excpetion=0` arrive before call from the Retirement (with `resume_form_excpetion=1`).  I need to find a way how to declare it cleanly

piotro888:
Time ~707200ns

piotro888:
I think that best solution would be to split `resume` into `resume_from_exception` and `resume_from_unsafe` with conflict priority explicitly defined in `Fetch`

xThaid:
OK, thanks, now I can see it.

I wonder though, why do we even allow for speculatively resuming the fetch unit? If we are precommiting a CSR instruction _without_ side effects, (in my opinion) it must not call fetch resume - this is clearly a side effect.

This fixes the failing tests:
```diff
diff --git a/coreblocks/func_blocks/csr/csr.py b/coreblocks/func_blocks/csr/csr.py
index 3e203caf..5191ea37 100644
--- a/coreblocks/func_blocks/csr/csr.py
+++ b/coreblocks/func_blocks/csr/csr.py
@@ -103,6 +103,7 @@ class CSRUnit(FuncBlock, Elaboratable):
         done = Signal()
         accepted = Signal()
         exception = Signal()
+        can_resume = Signal()
         precommitting = Signal()
 
         current_result = Signal(self.gen_params.isa.xlen)
@@ -205,7 +206,7 @@ class CSRUnit(FuncBlock, Elaboratable):
 
         @def_method(m, self.get_result, done)
         def _():
-            m.d.comb += accepted.eq(1)
+            m.d.comb += accepted.eq(can_resume)
             m.d.sync += reserved.eq(0)
             m.d.sync += instr.valid.eq(0)
             m.d.sync += done.eq(0)
@@ -249,6 +250,7 @@ class CSRUnit(FuncBlock, Elaboratable):
         @def_method(m, self.precommit)
         def _(rob_id: Value, side_fx: Value):
             with m.If(instr.rob_id == rob_id):
+                m.d.sync += can_resume.eq(side_fx)
                 m.d.comb += precommitting.eq(1)
                 m.d.comb += exe_side_fx.eq(side_fx)
 

```
Current setup seems to be broken anyway. csr unit sets the ready signal for method `fetch_resume` whenever `get_result` is called. However, if the unifier forwarder is not ready (another func unit inserted something there), `fetch_resume` won't run and its ready signal will be gone.

piotro888:
In both cases besides side_fx, it also needs to check if exception is not going to be reported on that instruction, otherwise the same conflict could happen if resume is blocked.

> Current setup seems to be broken anyway. csr unit sets the ready signal for method 

It would work, because there is at most one unsafe instruction at time in the core (that call resume via Forwarder)

But it would still break when we would introduce interrupt delay enforcing via insertion of extra instruction at end of the pipeline. Unsafe instruction on precommit stage with exception clear, could still conflict with that interrupt. So I think that solution with separate `resume` methods would fit better in the terms of cleaner logic later.

coreblocks/priv/csr/aliased.py
piotro888: thanks!

piotro888:
Should be finally ready to merge

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -1,44 +1,197 @@
...
+
+        # mret/entry conflict cannot happen in real conditons - mret is called under precommit
+        # it is split here to avoid complicated call graphs and conflicts that are not handled well by Transactron
+        with Transaction().body(m):
+            with m.If(self.entry.run):
+                self.mstatus_mie.write(m, {"data": 0})
+                self.mstatus_mpie.write(m, self.mstatus_mie.read(m).data)
+            with m.Elif(self.mret.run):
+                self.mstatus_mie.write(m, self.mstatus_mpie.read(m).data)
+                self.mstatus_mpie.write(m, {"data": 1})
```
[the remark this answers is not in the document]
piotro888:
```
18.6.4. Trap Return
[...]
MRET then in mstatus/mstatush sets MPV=0, MPP=0, MIE=MPIE, and MPIE=1.
```

lekcyjna123 (approved):

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -1,44 +1,197 @@
...
 
         @def_method(m, self.entry)
         def _():
-            m.d.sync += interrupt_pending.eq(0)
-            m.d.sync += self.interrupts_enabled.eq(0)
+            log.info(m, True, "Interrupt handler entry")
+            pass
+
+        # mret/entry conflict cannot happen in real conditons - mret is called under precommit
+        # it is split here to avoid complicated call graphs and conflicts that are not handled well by Transactron
```
tilk, line 159: Did you do this because of a real issue, or is it a premature optimization? This solution assumes that the transaction below never blocks - it doesn't, but what if some later change might change that?

Quick idea: perhaps what is needed is a way to assert that two methods/transactions will never conflict. This could be used by Transactron to prune the conflict graph. Of course, such an assertion would also raise an error when it is not satisfied.

coreblocks/priv/csr/aliased.py
```diff
@@ -0,0 +1,62 @@
...
+from transactron.core.method import Method
+from transactron.core.sugar import def_method
+from transactron.core.tmodule import TModule
+from transactron.utils.dependencies import DependencyContext
+
+
+class AliasedCSR(CSRRegister):  # TODO: CSR interface protocol
+    """
+    Temporary simple support for CSR aliasing for InternalInterruptController. Will be replaced with more complete
+    implemantation soon.
```
xThaid, line 18: typo

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -1,44 +1,197 @@
...
-        @def_method(m, self.report_interrupt)
-        def _():
-            m.d.sync += interrupt_pending.eq(1)
+        m.submodules += [self.mstatus_mie, self.mstatus_mpie, self.mstatus_mpp, self.mie, self.mip]
+
+        interrupt_enable = Signal()
+        mie = Signal(self.gen_params.isa.xlen)
+        mip = Signal(self.gen_params.isa.xlen)
+        with Transaction().body(m):
+            m.d.comb += [
```
xThaid, line 117: `m.d.av_comb`?

xThaid (approved):
Solid work! Just a small question: why are `custom_report_level` and `custom_report_edge` separated as two signals? As far as I understand it, their bits are mutually exclusive anyway.

coreblocks/priv/traps/interrupt_controller.py
piotro888: From meeting: left as is for now, to be changed to transactron explicit conflict remove + assert never conflicts when implemented

PR #660, opened 2024-04-15
Run benchmarks on every PR

### LINKED ISSUE #657: Run benchmarks on every PR
It is not necessary, but we could run both performance and synthesis benchmarks for every open PR (only if CI workflow succeeded) and make a bot put all the important information in the PR as a comment (IPCs on different benchmarks, fmax, etc)

Starting point: https://github.com/actions/github-script#comment-on-an-issue

xThaid:
The motivation is that performance changes should be visible by default (as opposed to manually triggering the benchmarks).

- When benchmarks are run on the master, the results will be saved to the cache.
- On PRs with `benchmark` label, benchmarks are run and compared to the base branch as the baseline.
- Metrics are now exposed to `Github job summary`, so one doesn't have to search through logs.

I am not adding `benchmark` label, as the job will fail. Most likely we need to merge this first, so permissions are propagated correctly. Here is the preview: https://github.com/xThaid/coreblocks/pull/2#issuecomment-2054193764

Fixes #657.


xThaid:
I removed `benchmark` label as it will trigger the job that issues a comment in this PR, which will fail due to lack of permissions. I believe that putting this modified workflow on the master branch should solve permission issues.

tilk (approved):
Nice. Didn't know that `tabulate` can generate tables for Github. Hope this works!

piotro888 (approved):

PR #663, opened 2024-04-17
Add custom combiners

tilk:
This PR allows nonexclusive methods with arguments by introducing the concept of a combiner function. When there are multiple calls, their arguments are combined into a single one which is then passed to the method body.


xThaid (approved):

piotro888 (approved):
lgtm, will be very useful soon!

PR #667, opened 2024-04-18
Fix fetcher throughput

tilk:
The second stage of the fetcher couldn't proceed with flushing when the output method was not ready, which limited the fetch throughput. This PR fixes that.

This change makes the loop in the `fibonacci` test run at IPC=1 basically, instead of 0.66 (there were two unnecessary wait cycles which this PR removes).

Edit: Looks like this change somehow uncovered a correctness problem.

### DIFF
--- a/coreblocks/frontend/fetch/fetch.py
+++ b/coreblocks/frontend/fetch/fetch.py
@@ -4,6 +4,7 @@ from amaranth.utils import exact_log2
 from amaranth.lib.coding import PriorityEncoder
 from transactron.lib import BasicFifo, Semaphore, ConnectTrans, logging, Pipe
 from transactron.lib.metrics import *
+from transactron.lib.simultaneous import condition
 from transactron.utils import MethodLayout, popcount
 from transactron.utils.transactron_helpers import from_method_layout
 from transactron import *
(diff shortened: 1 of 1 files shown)


xThaid:
I found the bug. The problem was when the fetch unit was being stalled by an exception. In that case we want to set `flushing_counter` to the number of items in the pipeline, but `m.d.sync += flushing_counter.eq(flushing_counter - 1)` took precedence and set the counter to 0. This wasn't not a problem before the change, because `stall_exception` would block the second stage transaction and wouldn't let override the flushing counter.
The fix:
```diff
diff --git a/coreblocks/frontend/fetch/fetch.py b/coreblocks/frontend/fetch/fetch.py
index c3736066..bd7806e9 100644
--- a/coreblocks/frontend/fetch/fetch.py
+++ b/coreblocks/frontend/fetch/fetch.py
@@ -93,9 +93,6 @@ class FetchUnit(Elaboratable):
         def flush():
             m.d.comb += flush_now.eq(1)
 
-        with m.If(flush_now):
-            m.d.sync += flushing_counter.eq(req_counter.count_next)
-
         current_pc = Signal(self.gen_params.isa.xlen, reset=self.gen_params.start_pc)
 
         stalled_unsafe = Signal()
@@ -365,6 +362,9 @@ class FetchUnit(Elaboratable):
                 with branch(flushing_counter != 0):
                     m.d.sync += flushing_counter.eq(flushing_counter - 1)
 
+        with m.If(flush_now):
+            m.d.sync += flushing_counter.eq(req_counter.count_next)
+
         @def_method(m, self.resume, ready=(stalled & (flushing_counter == 0)))
         def _(pc: Value, resume_from_exception: Value):
             log.info(m, True, "Resuming new_pc=0x{:x} from exception={}", pc, resume_from_exception)
```

tilk:
Looks like the benchmark numbers didn't change much. But if you look into the metrics, you can see that ROB utilization increased (see e.g. rob.size/bucket-32). The performance gain was probably eaten up by the miss penalty.

xThaid (approved):

lekcyjna123 (approved):

PR #668, opened 2024-04-18
Change 2-FIFOs to Pipes

tilk:
This PR changes some 2-FIFOs (mostly in the scheduler) into `Pipe`s. Pipes have more combinational connections than FIFOs, but use less resources. The purpose is to validate the usefulness of pipes for gluing the core together.

~TODO: replace more 2-FIFOs with pipes and check benchmarks again.~

Reopened PR to reduce the impact of frontend superscalarity on the amount of lingering instructions in the scheduler.

### DIFF
--- a/coreblocks/scheduler/scheduler.py
+++ b/coreblocks/scheduler/scheduler.py
@@ -3,7 +3,7 @@ from collections.abc import Sequence
 from amaranth import *
 
 from transactron import Method, Methods, Required, Transaction, TModule, def_method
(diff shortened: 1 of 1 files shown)


tilk:
Device utilization metric does some weird non-deterministic jumps. The double-triggered workflow got completely different results.

tilk:
Performing some PR necromancy, but this FIFO -> Pipe change has potential to reduce performance drop due to instructions lingering in the scheduler.

qbojj (approved):

PR #670, opened 2024-04-18
Fix benchmark workflow

### LINKED ISSUE #669: Benchmark workflow issues
Two issues I noticed:

* On PR creation with the benchmark label, the bot comment is added twice.
* In the bot comment, the upwards arrows could be green instead of red.

xThaid:
1. I changed red triangles 🔺🔻 to black ones ▲▼, because a) there are no green arrows in unicode, b) for some metrics lower value means better, but for some don't and I don't want to implement this logic as well.
2. Fixes the bug when a PR was opened with `benchmark` label and the workflow was triggered twice.
3. Fixed problem with write permissions. It only happened on PRs with a branch from a fork. Github by default never gives access tokens with write permissions to workflows running in the context of changes coming from the fork (that's unsafe). The solution was to create a new workflow that runs in the context of the master branch (trusted context) and trigger that workflow after the benchmark workflow is completed.

I tested the changes on my own fork.

Fixes #669

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - master
   pull_request:
-    types: [ opened, synchronize, labeled ]
+    types: [ synchronize, labeled ]
     branches:
       - master
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


tilk:
> 1. I changed red triangles 🔺🔻 to black ones ▲▼, because a) there are no green arrows in unicode, b) for some metrics lower value means better, but for some don't and I don't want to implement this logic as well.

Fair enough, I think that this should have an issue, though - colored triangles might be possible by using LaTeX or embedded SVGs.

tilk (approved):
Hope for the best!

piotro888:
Maybe the arrow can be changed to white up and red down for now?

lekcyjna123 (approved):

PR #679, opened 2024-04-23
Mocking validate_arguments

tilk:
This PR allows to write method mocks with argument validation.

I don't like that mocks using argument validation must be declared on construction. I couldn't find a way to avoid this - if someone has some idea, please let me know.

Fixes #677.


test/transactron/test_testing.py
```diff
@@ -0,0 +1,59 @@
...
+
+        return process
+
+    def validate_arguments(self, data: int):
+        return data == self.accepted_val
+
+    def changer(self):
+        for _ in range(50):
+            yield Tick("sync_neg")
+        self.accepted_val = 1
```
lekcyjna123, line 46: why this always set `1`?

test/transactron/test_testing.py
```diff
@@ -0,0 +1,59 @@
...
+
+class TestValidateArguments(TestCaseWithSimulator):
+    def control_caller(self, caller: TestbenchIO, method: TestbenchIO):
+        def process():
+            for _ in range(100):
+                val = random.randrange(2)
+                pre_accepted_val = self.accepted_val
+                ret = yield from caller.call_try(data=val)
+                if ret is None:
+                    assert val != pre_accepted_val or (yield from method.done())
```
lekcyjna123, line 33: why `(yield from method.done())` is needed?

lekcyjna123 (approved):

test/transactron/test_testing.py
tilk: Because, in the beginning of the simulation, the value is 0. This ensures that both `accepted_val` values are tested.

test/transactron/test_testing.py
tilk: Because, if both callers are calling the method with the right value, one of the callers will not run not because of the mismatch, but because the calls conflicted and other caller took priority.

tilk: I've strengthened this assertion to better express this.

lekcyjna123 (approved):

xThaid (approved):

PR #680, opened 2024-04-24
Remove relation graph in Transactron

tilk:
The relation graph seemed useful at the time, but it really isn't. This PR removes relation graphs from the Transactron core.


xThaid (approved):

lekcyjna123:
Have you checked if the graph generation from coreblocks still work with this changes?

lekcyjna123 (approved):

PR #681, opened 2024-04-24
Fix fetch resume

piotro888:
Fixes bug found in #654
Splits `fetch.resume` to separate `resume_from_exception` and `resume_from_unsafe`

I tried adding formal-like verification constraints, to verify method ordering assumptions from the comments, but have mixed feelings on the final solution. What do you think?


coreblocks/core.py
```diff
@@ -195,4 +194,7 @@ def elaborate(self, platform):
             read = self.connections.get_dependency(BranchVerifyKey())
             read(m)  # Consume to not block JB Unit
 
+        if self.gen_params.extra_verification:
+            m.d.comb += self.fetch._verification_resume_from_collector_ready.eq(fb_fetch_resume.ready)
```
tilk, line 198: Can this be done in `FetchUnit` instead?

tilk: This is related to `UnfiierKey` handling in `DependencyManager`, left for future PR.

coreblocks/frontend/fetch/fetch.py
```diff
@@ -367,21 +367,80 @@ def flush():
...
+                    m.d.sync += expect_unstall_unsafe.eq(1)
+                    with m.If(self.resume_from_unsafe.run):
+                        m.d.sync += expect_unstall_unsafe.eq(0)
+                        m.d.sync += prev_stalled_unsafe.eq(0)  # it is fine to be stalled now
+                        m.next = "running"
+                    with m.If(self.stall_exception.run):
+                        m.next = "stalled_exception"
+                    log.error(
+                        m,
+                        self.resume_from_exception.run & ~self.stall_exception.run,
```
lekcyjna123, line 429: `~self.stall_exception.run` is always `False` here

coreblocks/frontend/fetch/fetch.py
```diff
@@ -367,21 +367,80 @@ def flush():
...
+                        "unexpected resume_from_exception",
+                    )
+                with m.State("stalled_exception"):
+                    with m.If(self.resume_from_unsafe.run):
+                        log.error(m, ~expect_unstall_unsafe, "unexpected resume_from_unsafe")
+                        m.d.sync += expect_unstall_unsafe.eq(0)
+                    with m.If(self.resume_from_exception.run):
+                        # unstall_form_unsafe may be skipped if excpetion was reported on unsafe instruction,
+                        # invlid cases are verified by readiness check in running state
```
lekcyjna123, line 438:
```suggestion
                        # invalid cases are verified by readiness check in running state
```

xThaid:
I agree that having verification code is a good idea. Perhaps we could make another module (something like `FetchVerification`) that contains that logic to avoid polluting the main module, but then we need to expose some signals, so it is not perfect. Another option would be to create a function within the main module that is called from `elaborate` and contains the verification logic.

coreblocks/core.py
piotro888: fixed with #688

coreblocks/frontend/fetch/fetch.py
piotro888: why? it is not under the If

tilk (approved):

lekcyjna123 (approved):

PR #693, opened 2024-05-03
Fix double logger adding

### LINKED ISSUE #675: pytest fixtures are executed multiple times
Fixtures like `configure_profiles` and `configure_logging` from `TestCaseWithSimulator` are called for every test function in a test class. As a result, for example multiple logging handlers are added and logs are printed many times.

lekcyjna123:
I moved adding of our custom logger to "session" fixture, so there shouldn't be any more problems with duplicated logs. I tested it with command:
```sh
pytest -k FetchUnit_0 -s --log-level=debug
```

Resolves #675 
Based on #573


lekcyjna123:
I can, but both edits the same part of code, so there will be conflict between them if this review will be based on master.

tilk:
This seems to be a straightforward change, and it also fixes a very annoying bug, so it could be merged very quickly. Meanwhile #573 has now around 500 lines of code and a number of features besides the main one (the CAM), so I expect that PR to be reviewed after some time.

lekcyjna123:
Ok, I will prepare a separate change, but this would be probably in Sunday.

lekcyjna123:
I have created a version which is based on master. Please see #694

PR #695, opened 2024-05-05
Add pipelining support to LSU requester

lekcyjna123:
Here is a small refactor of the LSURequester it now support the request pipelining thanks to using the fifo. Additionally unit tests has to be updated, because after that change DummyLSU started to support reordering of miss-aligned instructions before the correct once.

Based on #696


coreblocks/func_blocks/fu/lsu/lsu_requester.py
```diff
@@ -0,0 +1,174 @@
...
+from transactron.lib.logging import HardwareLogger
+from transactron.lib import BasicFifo
+
+from coreblocks.params import *
+from coreblocks.arch import Funct3, ExceptionCause
+from coreblocks.peripherals.bus_adapter import BusMasterInterface
+from coreblocks.interface.layouts import LSULayouts
+
+
+class LSURequester(Elaboratable):
```
tilk, line 14: Did you just combine a refactor with a feature? How am I supposed to review this?

tilk:
No change in benchmarks, as Wishbone Classic doesn't support pipelining.

lekcyjna123:
> No change in benchmarks, as Wishbone Classic doesn't support pipelining.

Yes, I have expected that, but I started the benchmark to make sure that there is no regression.

coreblocks/func_blocks/fu/lsu/lsu_requester.py
```diff
@@ -100,12 +101,11 @@ def check_align(self, m: TModule, funct3: Value, addr: Value):
     def elaborate(self, platform):
         m = TModule()
 
-        addr_reg = Signal(self.gen_params.isa.xlen)
-        funct3_reg = Signal(Funct3)
-        store_reg = Signal()
-        request_sent = Signal()
+        m.submodules.args_fifo = args_fifo = BasicFifo(
+            [("addr", self.gen_params.isa.xlen), ("funct3", Funct3), ("store", 1)], 6
+        )
```
tilk, line 106: Why is this FIFO this long? Is there any advantage for this? Can't this be replaced with just a `Pipe`?

lekcyjna123: I have already thought about possible future usage and I assumed that ca. 4-5 instructions can be in flight.

tilk: How?

tilk: My reasoning is this: In non-pipelined buses like Wishbone Classic, one request every two cycles is accepted - but having a queue here doesn't cause any gain. In pipelined buses, the bus can basically accept a request every cycle, so there is nothing to queue.

lekcyjna123: As agreed on meeting I made this a parameter.

tilk (approved):

xThaid (approved):

PR #699, opened 2024-05-05
Move CoreCounter behind the frontend

xThaid:
- moved `CoreCounter` behind the frontend, so in particular behind all of its FIFOs. As the result, the flushing process should take a few cycles less.
- `decode_fifo` is now a `Pipe`
- the size of the instruction buffer can be parametrized (I increased the default)

edit:
~~I also had to rename `debug_signals` in CoreConfiguration as `auto_debug_signals` incorrectly assumed that this field is of protocol `HasDebugSignals`~~

Metric `backend.retirement.trap_latency/sum`:

| | aha-mont64 | crc32 | minver | nettle-sha256 | nsichneu | slre  | statemate | ud    |
|-|------------|-------|--------|---------------|----------|-------|-----------|-------|
|Before| 70032      | 2138  | 51797  | 1924          | 27950    | 89013 | 10643     | 42058 |
|After| 52303      | 2136  | 33884  | 1473          | 20082    | 75631 | 8539      | 24714 |

Depends on #698

### DIFF
--- a/coreblocks/params/genparams.py
+++ b/coreblocks/params/genparams.py
@@ -77,5 +77,6 @@ class GenParams(DependentCache):
         self.fetch_width_log = exact_log2(self.fetch_width)
 
         self.instr_buffer_size = cfg.instr_buffer_size
+        self.extra_verification = cfg.extra_verification
 
         self._toolchain_isa_str = gen_isa_string(extensions, cfg.xlen, skip_internal=True)
--- a/coreblocks/params/genparams.py
+++ b/coreblocks/params/genparams.py
@@ -49,7 +49,7 @@ class GenParams(DependentCache):
             enable=cfg.icache_enable,
         )
 
-        self.debug_signals_enabled = cfg.debug_signals
+        self.debug_signals_enabled = cfg.debug_signals_enable
 
         # Verification temporally disabled
         # if not optypes_required_by_extensions(self.isa.extensions) <= optypes_supported(func_units_config):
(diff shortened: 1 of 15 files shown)


coreblocks/params/genparams.py
```diff
@@ -48,7 +49,7 @@ def __init__(self, cfg: CoreConfiguration):
             enable=cfg.icache_enable,
         )
 
-        self.debug_signals_enabled = cfg.debug_signals
+        self.debug_signals_enabled = cfg.debug_signals_enable
```
tilk, line 52: It's `enabled` on the left side, and `enable` on the right. I think it would be better if the naming was consistent.

coreblocks/params/genparams.py
```diff
@@ -20,6 +20,7 @@ class GenParams(DependentCache):
     def __init__(self, cfg: CoreConfiguration):
         super().__init__()
 
+        self.cfg = cfg
```
tilk, line 23: Configurations were supposed to be templates for `GenParams`, there is some redundancy between them. I think that changing this is beyond the scope of this PR.

tilk (approved):

tilk:
The performance loss on `crc32` is worrying. Did you investigate why that happened? Is this because of `clear` conflicts?

xThaid:
The performance loss is caused by the increased size of the fetch queue (the instruction buffer). I need to investigate further why that is.

xThaid:
It seems that the performance loss is not directly related to this change. I opened an issue #702 for that.

lekcyjna123 (approved):

PR #703, opened 2024-05-12
Ring priority encoder

lekcyjna123:
Few times we had already a problem, where we would like to select something from the circular buffer. I have hit this once more in RS LSU and accidentally I have found out how to solve that issue. So here is an encoder which can be used in such situations.

ToDo:
- [ ] make the tests common for `MultiPriorityEncoder` and `RingMultiPriorityEncoder`

Based on #573


tilk (approved):

transactron/utils/amaranth_ext/elaboratables.py
```diff
@@ -377,3 +378,156 @@ def elaborate(self, platform):
...
+        m.submodules.multi_enc = multi_enc = MultiPriorityEncoder(self.input_width, self.outputs_count)
+        m.d.comb += multi_enc.input.eq(multi_enc_input)
+        for k in range(self.outputs_count):
+            moved_out = Signal(range(2 * self.input_width))
+            m.d.comb += moved_out.eq(multi_enc.outputs[k] + self.first)
+            corrected_out = Mux(moved_out >= self.input_width, moved_out - self.input_width, moved_out)
+
+            m.d.comb += self.outputs[k].eq(corrected_out)
+            m.d.comb += self.valids[k].eq(multi_enc.valids[k])
+        self.valids = multi_enc.valids
```
piotro888, line 532: Aren't these two lines conflicting or overwriting the definition?

transactron/utils/amaranth_ext/elaboratables.py
piotro888, line 1: `MultiPriorityEncoder` and `RingMultiPriorityEncoder` could be moved to separate file

piotro888 (approved):

transactron/utils/amaranth_ext/elaboratables.py
lekcyjna123: Fixed

transactron/utils/amaranth_ext/elaboratables.py
lekcyjna123: Let left this for a another review

PR #708, opened 2024-05-21
Components at top level in Transactron

### LINKED ISSUE #706: TransactronModule and Amaranth Components
The method for using Transactron in a project is by wrapping the main module using a `TransactronModule`. However, if the wrapped module is a `Component`, this hides the I/O ports, which complicates synthesis. A way to work around this is needed. One idea is copying the I/O ports from the internal `Component` into the `TransactronModule`.

tilk:
This PR allows to create a top-level Transactron component, which can be easily used in synthesis.

Fixes #706.

### DIFF
--- a/transactron/core/manager.py
+++ b/transactron/core/manager.py
@@ -3,6 +3,7 @@ from typing import Callable, Iterable, Sequence, TypeAlias, Tuple
 from os import environ
 from graphlib import TopologicalSorter
 from amaranth import *
+from amaranth.lib.wiring import Component, connect, flipped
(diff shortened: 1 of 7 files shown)


transactron/core/manager.py
```diff
@@ -479,3 +480,22 @@ def elaborate(self, platform):
         )
 
         return m
+
+
+class TransactionComponent(TransactionModule, Component):
```
piotro888, line 485: Some short docstring would be helpful for that end-user important module

piotro888 (approved):
Very nice changes!!

PR #709, opened 2024-05-21
Shift storage

lekcyjna123:
Here is a PR with implemtation of shift storage to be used in LSU RS. I used hypothesis to do tests and now working with it was much more comfortable than previously.

TODO:
- [ ] refactor test_transactron_lib_storage to have common generic_process function

### DIFF
--- a/transactron/lib/storage.py
+++ b/transactron/lib/storage.py
@@ -309,23 +309,55 @@ class AsyncMemoryBank(Elaboratable):
         return m
 
 class ShiftStorage(Elaboratable):
+    """ Ordered storage, which shifts data
+
+    This module implements a storage, which maintain all elements in continuous, ordered array
+    according to the age of the elements (element inserted earlier will be earlier in the list).
+    The order is restored after each `delete` operation by shifting all elemets which are older
+    that the removed one.
+
+
+    Attributes
+    ----------
+    read: Method
+        The read method. Accepts an `addr` from which data should be read.
+        The read response method. Return `data_layout` View which was saved on `addr` at reading
+        time. The `valid` signal indicates if the `addr` was inside the range.
+    push_back : Method
+        Insert the new entry on the end of storage. Ready only if there is a free place.
+    delate: Method
+        Remove the element from the `addr`. Shifts the elements to restore the proper order.
+    update: Method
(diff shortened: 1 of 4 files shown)


transactron/lib/storage.py
```diff
@@ -307,3 +307,100 @@ def _(arg):
...
+        delete_address = Signal(range(self.depth))
+
+        @def_method(m, self.read)
+        def _(addr):
+            return {"data": self._data[addr], "valid": addr < self._last}
+
+        @def_method(m, self.delete, ready=self._last > 0)
+        def _(addr):
+            m.d.top_comb += delete_address.eq(addr)
+            with m.If(addr < self._last):
```
tilk, line 379: What's the purpose of this `If`? It makes the delete not happen if a wrong address is passed. What's the use case? I believe this behavior could create silent bugs. Maybe use an assertion instead?

transactron/lib/storage.py
```diff
@@ -307,3 +307,100 @@ def _(arg):
...
+    def elaborate(self, platform):
+        m = TModule()
+
+        last_incr = Signal()
+        last_decr = Signal()
+        delete_address = Signal(range(self.depth))
+
+        @def_method(m, self.read)
+        def _(addr):
+            return {"data": self._data[addr], "valid": addr < self._last}
```
tilk, line 374: Do you have an use case for reading invalid addresses?

transactron/lib/storage.py
```diff
@@ -307,3 +307,100 @@ def _(arg):
...
+
+            @def_method(m, self.update)
+            def _(addr, data):
+                update_req_valid = Signal()
+                m.d.top_comb += update_req_valid.eq(addr < self._last)
+                affected_by_delete = Signal()
+                m.d.top_comb += affected_by_delete.eq(addr > delete_address)
+                with m.If(update_req_valid & (last_decr.implies(delete_address != addr))):
+                    m.d.sync += self._data[addr - Mux(last_decr, affected_by_delete, 0)].eq(data)
+                return {"err": ~update_req_valid}
```
tilk, line 393: Same as above. If there is no use case, I prefer an assertion to a bit which can be easily ignored.

PR #711, opened 2024-05-22
Add support for target prediction in JumpBranch unit

xThaid:
This PR prepares the jump branch unit for JALR target prediction. To do so I did the following changes:
- move a big part of the JB unit logic from `issue` to `accept` method. The rationale is that JALR targets will be stored in memory, which will have one cycle delay. Thus, we can't tell if we mispredicted at the time of `issue`. I know that this most likely will affect our longest critical path, but the announcement needs to be cut somewhere in the middle anyway.
- I added stubs for the methods used to get the JALR target prediction
- The request method doesn't take any arguments right now and that's deliberate - we can't make target prediction now anyway.

### DIFF
--- a/coreblocks/func_blocks/fu/jumpbranch.py
+++ b/coreblocks/func_blocks/fu/jumpbranch.py
@@ -188,13 +188,13 @@ class JumpBranchFuncUnit(FuncUnit, Elaboratable):
             jump_result = Mux(instr.taken, instr.jmp_addr, instr.reg_res)
             is_auipc = instr.type == JumpBranchFn.Fn.AUIPC
 
-            predicted_target_correctly = (instr.type != JumpBranchFn.Fn.JALR) | (
+            predicted_addr_correctly = (instr.type != JumpBranchFn.Fn.JALR) | (
                 target_prediction.valid & (target_prediction.cfi_target == instr.jmp_addr)
             )
 
             misprediction = Signal()
             m.d.av_comb += misprediction.eq(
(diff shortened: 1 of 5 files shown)


xThaid:
Branches are now resolved one cycle later, so most likely we need to flush one more instruction during an exception. Thus performance dropped.

coreblocks/func_blocks/fu/jumpbranch.py
```diff
@@ -155,79 +161,123 @@ def elaborate(self, platform):
...
-            m.d.top_comb += decoder.exec_fn.eq(arg.exec_fn)
-            m.d.top_comb += jb.fn.eq(decoder.decode_fn)
+            jump_result = Mux(instr.taken, instr.jmp_addr, instr.reg_res)
+            is_auipc = instr.type == JumpBranchFn.Fn.AUIPC
 
-            m.d.top_comb += jb.in1.eq(arg.s1_val)
-            m.d.top_comb += jb.in2.eq(arg.s2_val)
-            m.d.top_comb += jb.in_pc.eq(arg.pc)
-            m.d.top_comb += jb.in_imm.eq(arg.imm)
+            predicted_target_correctly = (instr.type != JumpBranchFn.Fn.JALR) | (
```
lekcyjna123, line 191: Maybe `predicted_addr_correctly`?

lekcyjna123 (approved):

tilk (approved):

PR #712, opened 2024-06-14
Implement full `mtval`

piotro888:
Implements full `mtval` CSR.

Required some tricks. This CSR is optional to fill for all exception types, so if some case becomes too problematic in the future, it can be set to 0.

Boom and XiangShan doesn't support setting it to instruction bytes in case of Illegal Instruction, but it was added to coreblocks at currently little cost. It is helpful, because it could speed up emulation of missing instructions (possibly in OpenSBI?)
XiangShan supports setting mtval to upper half of instruction address in case of fault on misaligned instruction crossing fetch blocks, hard to tell if supported in Boom.
Other cases are handled by both processors.


coreblocks/frontend/decoder/rvc.py
```diff
@@ -291,6 +291,8 @@ def elaborate(self, platform):
 
         res = self.instr_mux(quadrant, quadrants)
 
-        m.d.comb += self.instr_out.eq(Mux(res[1], res[0], IllegalInstr()))
+        # In case of illegal instruction, ouput `instr_in` to be able to save it into `mtval` CSR.
```
tilk, line 294:
```suggestion
        # In case of illegal instruction, output `instr_in` to be able to save it into `mtval` CSR.
```

coreblocks/func_blocks/csr/csr.py
```diff
@@ -208,7 +209,21 @@ def _():
             interrupt = self.dependency_manager.get_dependency(AsyncInterruptInsertSignalKey())
 
             with m.If(exception):
-                report(m, rob_id=instr.rob_id, cause=ExceptionCause.ILLEGAL_INSTRUCTION, pc=instr.pc)
+                mtval = Signal(self.gen_params.isa.xlen)
+                # re-encode the CSR instruction to speed-up missing CSR emulation (optional, otherwise tval _must_ be 0)
```
tilk, line 213:
```suggestion
                # re-encode the CSR instruction to speed-up missing CSR emulation (optional, otherwise mtval _must_ be 0)
```

coreblocks/frontend/decoder/instr_decoder.py
```diff
@@ -317,6 +317,14 @@ def elaborate(self, platform):
                 self.rs1_v.eq(0),
             ]
 
+        # HACK: pass logical registers is unused high bits of CSR instruction for `mtval` reconstruction
```
tilk, line 320:
```suggestion
        # HACK: pass logical registers in unused high bits of CSR instruction for `mtval` reconstruction
```

tilk (approved):

coreblocks/frontend/decoder/decode_stage.py
```diff
@@ -95,7 +95,9 @@ def elaborate(self, platform):
                         "rl_s1": Mux(instr_decoder.rs1_v & (~exception_override), instr_decoder.rs1, 0),
                         "rl_s2": Mux(instr_decoder.rs2_v & (~exception_override), instr_decoder.rs2, 0),
                     },
-                    "imm": instr_decoder.imm,
+                    "imm": Mux(
+                        ~exception_override, instr_decoder.imm, Mux(raw.access_fault, raw.access_fault, raw.instr)
```
lekcyjna123, line 99: Shouldn't be `raw.pc` in case when `raw.access_fault` is true?

coreblocks/interface/layouts.py
```diff
@@ -444,7 +444,7 @@ class FetchLayouts:
     def __init__(self, gen_params: GenParams):
         fields = gen_params.get(CommonLayoutFields)
 
-        self.access_fault: LayoutListField = ("access_fault", 1)
+        self.access_fault: LayoutListField = ("access_fault", 2)
```
lekcyjna123, line 447: It would be nice to have the comment what mean each bit, because it is not obvious from the definition.

test/asm/mtval.asm
```diff
@@ -0,0 +1,97 @@
...
+    
+    addi x8, x8, 4
+    
+    mret
+
+fail:
+    j fail
+
+.org 0x0FFE 
+# it is legal - C is enabled in core, but can't be enabled in toolchain to keep 4-byte nops
```
lekcyjna123, line 56: So on the beginning of the file I would suggest to add comment that this test have to be run with C enabled.

lekcyjna123 (approved):

test/asm/mtval.asm
```diff
@@ -0,0 +1,97 @@
+    la x1, handler
```
lekcyjna123, line 1: Describing the high level idea of the test will be welcome

coreblocks/interface/layouts.py
tilk: Maybe even better, an enum?

tilk:
Waiting until comments are addressed.

coreblocks/frontend/decoder/decode_stage.py
piotro888: It is correct, Exception FU uses information from `access_fault` to determine correct PC. Switched to enum and documented this mux

PR #719, opened 2024-07-03
Handle transitive calls from a nonexclusive method

### LINKED ISSUE #717: Nonexclusivity of methods should be transitive
Let's assume we have a nonexclusive method `non_method` that calls another (exclusive) method. Now if we have two transactions `t1` and `t2` both calling `non_method`, transactron will create a conflict between `t1` and `t2`, which is unnecessary.

tilk:
Fixes #717.

### DIFF
--- a/transactron/core/manager.py
+++ b/transactron/core/manager.py
@@ -27,24 +27,28 @@ TransactionScheduler: TypeAlias = Callable[["MethodMap", TransactionGraph, Trans
 class MethodMap:
     def __init__(self, transactions: Iterable["Transaction"]):
         self.methods_by_transaction = dict[Transaction, list[Method]]()
+        self.exclusive_methods_by_transaction = dict[Transaction, list[Method]]()
         self.transactions_by_method = defaultdict[Method, list[Transaction]](list)
         self.readiness_by_method_and_transaction = dict[tuple[Transaction, Method], ValueLike]()
         self.method_parents = defaultdict[Method, list[TransactionBase]](list)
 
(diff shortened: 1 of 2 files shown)


xThaid (approved):
Thanks for the quick fix!

transactron/core/manager.py
```diff
@@ -149,7 +153,12 @@ def add_edge(begin: Transaction, end: Transaction, priority: Priority, conflict:
                 continue
             for transaction1 in method_map.transactions_for(method):
                 for transaction2 in method_map.transactions_for(method):
-                    if transaction1 is not transaction2 and not transactions_exclusive(transaction1, transaction2):
+                    if (
+                        transaction1 is not transaction2
+                        and not transactions_exclusive(transaction1, transaction2)
+                        and method in method_map.exclusive_methods_by_transaction[transaction1]
+                        and method in method_map.exclusive_methods_by_transaction[transaction2]
+                    ):
```
piotro888, line 161: In the case like
```
T1 ------------------------> NORMAL
T2 -> NONEXCLUSIVE ---------/
```
T1 should conflict with T2. But because `NORMAL` is not in the `exclusive_methods_by_transaction[T2]`, it doesn't. I modified the tests, and both transactions execute at the same time. Am I doing/understanding something incorrectly, or there is a problem?

tilk: There is a problem, thanks for noticing that.

lekcyjna123 (approved):

PR #721, opened 2024-07-08
Add a trap handler to the benchmark framework

### LINKED ISSUE #658: Add a trap handler to benchmarks
Right now if an exception is raised during the execution of benchmarks, the core will jump to 0x0 (the default value of `mtvec`) and will keep executing the code. This will result in a benchmark timeout - cryptic failure, hard to debug, long execution time. Instead, we could add a trap handler to `crt0.S` of benchmarks, so in the case of an exception the handler could write all useful information to the MMIO and terminate the simulation right away.

xThaid:
When an exception is thrown during the execution of benchmarks, we handle it in a custom trap handler and report it to the user. For example:
```
  1214.50ns INFO     cocotb.regression                  crc32 failed
                                                          File "/home/jurb/dev/repos/coreblocks/test/regression/cocotb.py", line 271, in _my_test
                                                            await function(dut, *args, **kwargs)
                                                          File "/home/jurb/dev/repos/coreblocks/test/regression/cocotb/benchmark_entrypoint.py", line 12, in _do_benchmark
                                                            await run_benchmark(CocotbSimulation(dut), benchmark_name)
                                                          File "/home/jurb/dev/repos/coreblocks/test/regression/benchmark.py", line 97, in run_benchmark
                                                            raise RuntimeError(
                                                        RuntimeError: An exception was thrown while executing the benchmark. mcause: ILLEGAL_INSTRUCTION, mepc: 0xb2

```

Previously `mtvec` was set to `0x0`, so the program would just start over again and the information about the exception would be lost.

edit: updated the example error

Fixes #658


piotro888 (approved):

test/regression/benchmark.py
```diff
@@ -85,6 +93,11 @@ async def run_benchmark(sim_backend: SimulationBackend, benchmark_name: str):
     if not result.success:
         raise RuntimeError("Simulation timed out")
 
+    if mmio.return_code() == -1:
+        raise RuntimeError(
+            f"An exception was thrown while executing the benchmark. mcause: {mmio.mcause()}, mepc: 0x{mmio.mepc():x}"
```
tilk, line 98: Maybe it's worth it to try to decode the cause here and show a human-readable string?

xThaid: added

tilk (approved):

PR #722, opened 2024-07-19
More uses of av_comb, slight refactor of RF

tilk:
Probably won't change much, but here we go.

### DIFF
--- a/coreblocks/core_structs/rf.py
+++ b/coreblocks/core_structs/rf.py
@@ -48,7 +48,8 @@ class RegisterFile(Elaboratable):
 
         @def_method(m, self.read1)
         def _(reg_id: Value):
-            forward = (being_written == reg_id) & (reg_id != 0)
+            forward = Signal()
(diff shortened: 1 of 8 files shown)


coreblocks/core_structs/rf.py
```diff
@@ -43,23 +46,17 @@ def elaborate(self, platform):
...
+            forward = (being_written == reg_id) & (reg_id != 0)
             return {
                 "reg_val": Mux(forward, written_value, self.entries[reg_id].reg_val),
                 "valid": Mux(forward, 1, self.entries[reg_id].valid),
             }
 
         @def_method(m, self.read2)
         def _(reg_id: Value):
-            forward = being_written == reg_id
+            forward = (being_written == reg_id) & (reg_id != 0)
```
lekcyjna123, line 59: Shouldn't be that a new signal to not copy the logic into two multiplexers?

lekcyjna123 (approved):

piotro888 (approved):

PR #726, opened 2024-08-29
IOInterface typing wrapper for lib.data signatures

piotro888:
Using `amaranth.lib.data` Signatures with typing comes with some boilerplate and repetition :(

I came up with a class that can be used both for typing and to get amaranth `Signature` for use in `Component` constructor.

It works by crating a class with `IOSignal` fields declared as `self.` in constructor, so it can be directly for typechecking. `IOSignal` type derives `Value` with additonal shape and direction information. 
`IOInterface` is never used globally, but is intended only for interface typing.  Those signals will be filled with actual objects at `super().__init__(...` `Component` constructor (that gets Signature from generated `IOInterface.signature`). Nested `IOInterfaces` are supported. 
 
It is *a bit* hacky, but what do you think?

### DIFF
--- a/coreblocks/peripherals/wishbone.py
+++ b/coreblocks/peripherals/wishbone.py
@@ -1,16 +1,15 @@
 from amaranth import *
-from amaranth.lib.wiring import PureInterface, Signature, In, Out, Component
+from amaranth.lib.wiring import In, Out, Component
 from functools import reduce
-from typing import Protocol, cast
 import operator
 
 from transactron import Method, def_method, TModule
 from transactron.core import Transaction
 from transactron.lib import AdapterTrans, BasicFifo
 from transactron.utils import OneHotSwitchDynamic, assign, RoundRobin
-from transactron.utils._typing import AbstractInterface, AbstractSignature
 from transactron.lib.connectors import Forwarder
 from transactron.utils.transactron_helpers import make_layout
+from transactron.utils.amaranth_ext.io_interface import *
 from transactron.lib import logging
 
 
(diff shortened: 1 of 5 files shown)


tilk:
The boilerplate reduction is nice, but this _is_ a little bit hacky. Maybe, instead of a custom `amaranth.Value` subclass, `typing.Annotated` could be used to store the same information?

coreblocks/peripherals/wishbone.py
```diff
@@ -365,8 +341,8 @@ class WishboneMuxer(Component):
     def __init__(self, wb_params: WishboneParameters, num_slaves: int, ssel_tga: Signal):
         super().__init__(
             {
-                "master_wb": Out(WishboneSignature(wb_params)),
-                "slaves": In(WishboneSignature(wb_params)).array(num_slaves),
+                "master_wb": Out(WishboneInterface(wb_params).signature),
+                "slaves": Out(WishboneInterface(wb_params).signature).array(num_slaves),
```
lekcyjna123, line 345: Why `In` was changed to `Out`?

transactron/utils/amaranth_ext/io_interface.py
```diff
@@ -0,0 +1,112 @@
...
+
+__all__ = [
+    "IOSignal",
+    "SignalIn",
+    "SignalOut",
+    "IOInterface",
+]
+
+
+class IOSignal(Value):
```
lekcyjna123, line 15: As discussed on the meeting it would be great to have the name changed.

lekcyjna123, line 15: Doc string will be welcome

lekcyjna123 (commented):
For me it looks good.

piotro888:
Merged in https://github.com/kuznia-rdzeni/transactron/pull/42

PR #727, opened 2024-08-31
Add Lookup3 and Sip hashes

lekcyjna123:
Here is a draft pull request of implementation of two non-cryptographic hashes:
- Lookup3 (http://www.burtleburtle.net/bob/hash/doobs.html)
- SipHash (https://eprint.iacr.org/2012/351.pdf)

I had an idea to test them using statistical test to check if they hash into uniform distribution, but sadly for small input data (from range [0,1023]) there is some bias and statistical tests fails... Whats more I am not sure if we want to have tests based on statistics, because this can cause instability.

### DIFF
--- a/test/transactron/utils/test_hw_hash.py
+++ b/test/transactron/utils/test_hw_hash.py
@@ -1,0 +1,95 @@
+import math
+from transactron.testing import *
+import random
+from transactron.utils.amaranth_ext.hw_hash import *
+import pytest
+
+
+def rot(x, k) -> int:
+    return ((x) << (k)) | ((x) >> (32 - (k)))
+
+
+def lookup3(value, seed):
+    if seed is None:
+        seed = 0xDEADBEEF
+    a = (value & ((1 << 32) - 1)) + seed
+    b = ((value >> 32) & ((1 << 32) - 1)) + seed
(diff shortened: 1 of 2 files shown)


tilk:
> Whats more I am not sure if we want to have tests based on statistics, because this can cause instability.

We're not supposed to implement some dreamed up hash functions, but some existing ones with established properties. The only thing to test is to compare a hardware implementation to a software one.

test/transactron/test_transactron_lib_storage.py
```diff
@@ -133,3 +133,79 @@ def test_random(self, in_push, in_write, in_read, in_remove):
...
+
+    def query_resp_process(self, input):
+        def f():
+            input_rev = list(reversed(input))
+            while not self.query_req_end:
+                count = yield from self.circ.query_resp.call()
+                in_val = input_rev.pop()
+                # Check if there is at least as many elements as we put inside (can be more because of aliasing)
+                assert count["count"] >= sum(map(lambda x: 1, filter(lambda x: x == in_val, input)))
+                assert count["count"] in self.table_snapshots[-1]
```
tilk, line 185: This is a very weak test. I believe a constant table which is filled with maximum values would satisfy it.

PR #728, opened 2024-09-10
FPU rounding module

Durchbruchswagen:
Implementation of the rounding module for FPU. 

This module excepts that input is in normalized form (unless the number is subnormal) and, in accordance with RISC-V specification, detects tininess after rounding. The flag that allows generation of a rounding module that assumes that input was already rounded is to be used by modules that are able to perform both their operation and rounding at the same time (adder/subtractor, for example). In that case, this module only checks for errors and special cases (one of the inputs was inf/nan).

### DIFF
--- a/coreblocks/func_blocks/fu/fpu/fpu_common.py
+++ b/coreblocks/func_blocks/fu/fpu/fpu_common.py
@@ -28,20 +28,3 @@ class FPUParams:
     ):
         self.sig_width = sig_width
         self.exp_width = exp_width
-
-
-class FPURoundingParams:
-    """FPU rounding module signature
-
-    Parameters
-    ----------
-    fpu_params: FPUParams
-        FPU parameters
-    is_rounded:bool
-        This flags indicates that the input number was already rounded.
-        This creates simpler version of rounding module that only performs error checking and returns correct number.
-    """
-
-    def __init__(self, fpu_params: FPUParams, *, is_rounded: bool = False):
-        self.fpu_params = fpu_params
-        self.is_rounded = is_rounded
--- a/coreblocks/func_blocks/fu/fpu/fpu_common.py
+++ b/coreblocks/func_blocks/fu/fpu/fpu_common.py
@@ -34,7 +34,7 @@ class FPURoundingParams:
     """FPU rounding module signature
 
     Parameters
-    -----------
+    ----------
     fpu_params: FPUParams
         FPU parameters
     is_rounded:bool
--- a/coreblocks/func_blocks/fu/fpu/fpu_common.py
+++ b/coreblocks/func_blocks/fu/fpu/fpu_common.py
@@ -9,12 +9,12 @@ class RoundingModes(enum.Enum):
     ROUND_NEAREST_AWAY = 4
 
 
-class Errors(enum.IntFlag, shape=5):
-    INVALID_OPERATION = 0
-    DIVISION_BY_ZERO = 1
-    OVERFLOW = 2
-    UNDERFLOW = 3
-    INEXACT = 4
+class Errors(enum.IntFlag):
+    INVALID_OPERATION = enum.auto()
+    DIVISION_BY_ZERO = enum.auto()
+    OVERFLOW = enum.auto()
+    UNDERFLOW = enum.auto()
+    INEXACT = enum.auto()
 
 
 class FPUParams:
--- a/coreblocks/func_blocks/fu/fpu/fpu_common.py
+++ b/coreblocks/func_blocks/fu/fpu/fpu_common.py
@@ -1,7 +1,7 @@
 from amaranth.lib import enum
 
 
-class RoundingModes(enum.Enum, shape=3):
+class RoundingModes(enum.Enum):
     ROUND_UP = 3
     ROUND_DOWN = 2
     ROUND_ZERO = 1
--- a/coreblocks/func_blocks/fu/fpu/fpu_common.py
+++ b/coreblocks/func_blocks/fu/fpu/fpu_common.py
@@ -9,13 +9,12 @@ class RoundingModes(enum.Enum):
     ROUND_NEAREST_AWAY = 4
 
 
-class Errors(enum.IntFlag):
+class Errors(enum.IntFlag, shape=5):
     INVALID_OPERATION = 0
     DIVISION_BY_ZERO = 1
     OVERFLOW = 2
     UNDERFLOW = 3
     INEXACT = 4
-
 
 
 class FPUParams:
--- a/coreblocks/func_blocks/fu/fpu/fpu_common.py
+++ b/coreblocks/func_blocks/fu/fpu/fpu_common.py
@@ -38,7 +38,7 @@ class FPURoundingParams:
     fpu_params: FPUParams
         FPU parameters
     is_rounded:bool
-        This flags indicates that the input number was already rounded. This creates simpler version of rounding module that only performs error checking and return corect number.
+        This flags indicates that the input number was already rounded. This creates simpler version of rounding module that only performs error checking and returns correct number.
     """
 
     def __init__(self, fpu_params: FPUParams, *, is_rounded: bool = False):
--- a/coreblocks/func_blocks/fu/fpu/fpu_common.py
+++ b/coreblocks/func_blocks/fu/fpu/fpu_common.py
@@ -1,0 +1,46 @@
+from amaranth.lib import enum
+
+
+class RoundingModes(enum.Enum, shape=3):
+    ROUND_UP = 3
+    ROUND_DOWN = 2
+    ROUND_ZERO = 1
+    ROUND_NEAREST_EVEN = 0
+    ROUND_NEAREST_AWAY = 4
+
+
+class FPUParams:
+    """FPU parameters
+
+    Parameters
+    ----------
+    sig_width: int
+        Width of significand
+    exp_width: int
+        Width of exponent
+    """
+
+    def __init__(
+        self,
+        *,
+        sig_width: int = 24,
+        exp_width: int = 8,
+    ):
+        self.sig_width = sig_width
+        self.exp_width = exp_width
+
+
+class FPURoundingParams:
+    """FPU rounding module signature
+
+    Parameters
+    -----------
+    fpu_params: FPUParams
+        FPU parameters
+    is_rounded:bool
+        This flags indicates that the input number was already rounded. This creates simpler version of rounding module that only performs error checking and return corect number.
+    """
+
+    def __init__(self, fpu_params: FPUParams, *, is_rounded: bool = False):
+        self.fpu_params = fpu_params
+        self.is_rounded = is_rounded
--- a/coreblocks/func_blocks/fu/fpu/fpu_common.py
+++ b/coreblocks/func_blocks/fu/fpu/fpu_common.py
@@ -9,13 +9,22 @@ class RoundingModes(enum.Enum):
     ROUND_NEAREST_AWAY = 4
 
 
+class Errors(enum.IntFlag):
+    INVALID_OPERATION = 0
+    DIVISION_BY_ZERO = 1
+    OVERFLOW = 2
+    UNDERFLOW = 3
+    INEXACT = 4
+
+
+
 class FPUParams:
     """FPU parameters
 
     Parameters
     ----------
     sig_width: int
-        Width of significand
+        Width of significand, including implicit bit
     exp_width: int
         Width of exponent
     """
--- a/coreblocks/func_blocks/fu/fpu/fpu_common.py
+++ b/coreblocks/func_blocks/fu/fpu/fpu_common.py
@@ -38,7 +38,8 @@ class FPURoundingParams:
     fpu_params: FPUParams
         FPU parameters
     is_rounded:bool
-        This flags indicates that the input number was already rounded. This creates simpler version of rounding module that only performs error checking and returns correct number.
+        This flags indicates that the input number was already rounded.
+        This creates simpler version of rounding module that only performs error checking and returns correct number.
     """
 
     def __init__(self, fpu_params: FPUParams, *, is_rounded: bool = False):
--- a/coreblocks/func_blocks/fu/fpu/fpu_error_module.py
+++ b/coreblocks/func_blocks/fu/fpu/fpu_error_module.py
@@ -3,6 +3,7 @@ from transactron import TModule, Method, def_method
 from coreblocks.func_blocks.fu.fpu.fpu_common import (
     RoundingModes,
     FPUParams,
+    Errors,
 )
 
 
@@ -30,7 +31,7 @@ class FPUErrorMethodLayout:
             ("sign", 1),
             ("sig", fpu_params.sig_width),
             ("exp", fpu_params.exp_width),
-            ("errors", 5),
+            ("errors",len(Errors)),
         ]
 
 
--- a/coreblocks/func_blocks/fu/fpu/fpu_error_module.py
+++ b/coreblocks/func_blocks/fu/fpu/fpu_error_module.py
@@ -1,0 +1,167 @@
+from amaranth import *
+from transactron import TModule, Method, def_method
+from coreblocks.func_blocks.fu.fpu.fpu_common import (
+    RoundingModes,
+    FPUParams,
+)
+
+
+class FPUErrorMethodLayout:
+    """FPU error checking module layouts for methods
+    Parameters
+    ----------
+    fpu_params: FPUParams
+        FPU parameters
+    """
+
+    def __init__(self, *, fpu_params: FPUParams):
+        self.error_in_layout = [
+            ("sign", 1),
+            ("sig", fpu_params.sig_width),
+            ("exp", fpu_params.exp_width),
+            ("rounding_mode", RoundingModes),
+            ("inexact", 1),
+            ("invalid_operation", 1),
+            ("division_by_zero", 1),
+            ("input_inf", 1),
+        ]
+        self.error_out_layout = [
+            ("sign", 1),
+            ("sig", fpu_params.sig_width),
+            ("exp", fpu_params.exp_width),
+            ("errors", 5),
+        ]
+
+
+class FPUErrorModule(Elaboratable):
+    """FPU error checking module
+
+    Parameters
+    ----------
+    fpu_params: FPUParams
+        FPU rounding module parameters
+
+    Attributes
+    ----------
+    error_checking_request: Method
+        Transactional method for initiating error checking of a floating point number.
+        Takes 'error_in_layout' as argument
+        Returns final number and errors as 'error_out_layout'
+    """
+
+    def __init__(self, *, fpu_params: FPUParams):
+
+        self.fpu_errors_params = fpu_params
+        self.method_layouts = FPUErrorMethodLayout(fpu_params=self.fpu_errors_params)
+        self.error_checking_request = Method(
+            i=self.method_layouts.error_in_layout,
+            o=self.method_layouts.error_out_layout,
+        )
+
+    def elaborate(self, platform):
+        m = TModule()
+
+        max_exp = C(
+            2 ** (self.fpu_errors_params.exp_width) - 1,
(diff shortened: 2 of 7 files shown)


coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+                m.d.comb += self.final_exp.eq(self.max_exp)
+                m.d.comb += self.final_sig.eq(arg.sig)
+                m.d.comb += self.final_sign.eq(arg.sign)
+
+            with m.Elif(self.is_inf):
+
+                m.d.comb += self.final_exp.eq(self.max_exp)
+                m.d.comb += self.final_sig.eq(arg.sig)
+                m.d.comb += self.final_sign.eq(arg.sign)
+
```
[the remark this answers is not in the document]
piotro888: Why is the `exp` part overwritten (like the number could be incorrect when flags are set), while `sig` value is preserved, while being critical for classification?

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+            with m.Elif(self.overflow):
+
+                with m.If(
```
piotro888, line 193: Amaranth switch/case could be used: `with m.Switch(arg.rounding_mode)`

test/func_blocks/fu/test_fpu_rounding.py
```diff
@@ -0,0 +1,393 @@
...
+                help_values.max_exp,
+                request_adapter,
+                is_input_not_rounded,
+            )
+            yield from one_rounding_mode_test(
+                RoundingModes.ROUND_NEAREST_AWAY,
+                tie_to_away_inc_array,
+                0,
+                help_values.max_exp,
+                0,
```
piotro888, line 350: Tests can be parametrized with `@parametrize_class`, `@parameterized.expand` or similar

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+
+    def __init__(self, *, fpu_params: FPUParams):
+        self.rounding_in_layout = [
+            ("sign", 1),
+            ("sig", fpu_params.sig_width),
+            ("exp", fpu_params.exp_width),
+            ("guard_bit", 1),
+            ("sticky_bit", 1),
+            ("rounding_mode", 3),
```
piotro888, line 54: enums can be used as signal shapes
```suggestion
            ("rounding_mode", RoundingModes),
```

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+
+                    with m.Else():
+
+                        m.d.comb += self.final_exp.eq(self.max_exp)
+                        m.d.comb += self.final_sig.eq(0)
+                        m.d.comb += self.final_sign.eq(arg.sign)
+
+            with m.Else():
+                with m.If((self.rounded_exp == 0) & (self.normalised_sig[-1] == 1)):
+                    m.d.comb += self.final_exp.eq(1)
```
piotro888, line 238: Can you explain this condition?

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+            m.d.comb += self.final_errors[2].eq(self.overflow)
+            m.d.comb += self.final_errors[3].eq(self.underflow)
+            m.d.comb += self.final_errors[4].eq(self.inexact)
+
+            self.rtval["exp"] = self.final_exp
+            self.rtval["sig"] = self.final_sig
+            self.rtval["sign"] = self.final_sign
+            self.rtval["errors"] = self.final_errors
+
+            return self.rtval
```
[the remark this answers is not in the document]
Durchbruchswagen: Done, removed `rtval` and changed it to inline dictionary in return

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+            ("sign", 1),
+            ("sig", fpu_params.sig_width),
+            ("exp", fpu_params.exp_width),
+            ("errors", 5),
+        ]
+
+
+class FPUrounding(Component):
+
+    fpu_rounding: FPURoundingSignature
```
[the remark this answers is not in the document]
Durchbruchswagen: You're right, the `Component` interface isn't used anywhere. It is now removed.

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+
+        self.final_exp = Signal(self.fpu_rounding_params.fpu_params.exp_width)
+        self.final_sig = Signal(self.fpu_rounding_params.fpu_params.sig_width)
+        self.final_sign = Signal()
+        self.final_errors = Signal(5)
+
+    def elaborate(self, platform):
+        m = TModule()
+
+        @def_method(m, self.rounding_request)
```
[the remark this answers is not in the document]
Durchbruchswagen: Done, replaced all `comb` domains with `av_comb`

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+        ]
+        self.rounding_out_layout = [
+            ("sign", 1),
+            ("sig", fpu_params.sig_width),
+            ("exp", fpu_params.exp_width),
+            ("errors", 5),
+        ]
+
+
+class FPUrounding(Component):
```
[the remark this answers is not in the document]
Durchbruchswagen: Done, added docstring

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+        self.inexact = Signal()
+        self.tininess = Signal()
+        self.is_inf = Signal()
+        self.is_nan = Signal()
+        self.input_not_special = Signal()
+        self.rounded_inexact = Signal()
+
+        self.final_exp = Signal(self.fpu_rounding_params.fpu_params.exp_width)
+        self.final_sig = Signal(self.fpu_rounding_params.fpu_params.sig_width)
+        self.final_sign = Signal()
```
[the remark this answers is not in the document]
Durchbruchswagen: Done

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+
+                with m.Else():
+                    m.d.comb += self.normalised_sig.eq(self.rounded_sig)
+                    m.d.comb += self.final_guard_bit.eq(arg.guard_bit)
+                    m.d.comb += self.final_sticky_bit.eq(arg.sticky_bit)
+                    m.d.comb += self.rounded_exp.eq(arg.exp)
+
+            m.d.comb += self.rounded_inexact.eq(self.final_guard_bit | self.final_sticky_bit)
+            m.d.comb += self.is_nan.eq(arg.errors[0] | arg.input_nan)
+            m.d.comb += self.is_inf.eq(arg.errors[1] | arg.input_inf)
```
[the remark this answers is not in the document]
Durchbruchswagen: So about the intended use of the module, when `is_rounded = False` the module does two things: it rounds the input (increments by one or not depending on `round_bit`, `sticky_bit`, and 'rounding_mode') and then checks for floating point exceptions: `invalid operation`, `division by zero`, `overflow`, 'undeflow', and `inexact`. This module is to be used by other modules that perform floating point operations (addition, subtraction, etc.). Some operations can be designed in a way that performs both computation and rouding at the same time (right now I only know that this is possible only for addition and subtraction), and those modules should use `FPUrounding` with `is_rounded = True` to skip the increment part.
 
About `errors`, at first I didn't know which errors I could detect directly during rounding, so I just put all of them as an input, but now I know that the only exceptions I can't detect during rounding are `invalid operation` and `division by zero` (this is because I need to have prior knowledge about operations and operands).

`inpunt_nan/inf` are not duplicates of errors. `inf` and `qNaN` are not only used to indicate exceptions; they are also floating point numbers/objects that we can perform operations on (for example, `inf + x` equals exact `inf` for finite `x`).

In the case of 'inf', I'm unable to distinguish between `inf` created by overflow and by the correct operation. This is what `input_inf` is used for; it tells the rounding module that the input `inf` is created due to `infinity arithmetic` and not overflow. 

`input_nan` is after consideration, possibly redundant. `Nan` can only be created when at least one operand is `Nan` or due to `invalid opeartion'. By checking the `invalid_operation` argument, I can distinguish between those two cases. The only use for having `input_nan` is to allow modules that perform computation to skip changing their result to `qNaN`; they only need to set `input_nan = 1` and the rounding module takes care of the rest, because in RISC-V we do not propagate `Nan` payloads; all `Nan` results have to be canonical `Nan`-s. After all, I decided it was better to just create both `NaN` and `inf` in modules that are performing floating point operations, so I removed 'input_nan'.

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+            m.d.comb += self.input_not_special.eq(~(self.is_nan) & ~(self.is_inf))
```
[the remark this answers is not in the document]
Durchbruchswagen: Done.

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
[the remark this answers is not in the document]
Durchbruchswagen: To be honest, I've no idea what my great plan was when I was writing that part of the code (I should either take all parts of result from `arg` or from constant values). As I explained in https://github.com/kuznia-rdzeni/coreblocks/pull/728#discussion_r1783456402 `Nan` and `Inf` are going to be created in modules that perform floating point operations. So I refactored this part into a single `If` that takes values from `arg` as a result.

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,257 @@
...
+    fpu_params: FPUParams
+        FPU parameters
+    """
+
+    def __init__(self, *, fpu_params: FPUParams):
+        self.rounding_in_layout = [
+            ("sign", 1),
+            ("sig", fpu_params.sig_width),
+            ("exp", fpu_params.exp_width),
+            ("guard_bit", 1),
```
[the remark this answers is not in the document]
Durchbruchswagen: Yes, my bad. After reading a few papers, it can get a little bit confusing, but most of the books and papers define this bit as a `round bit`.

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
Durchbruchswagen: Done.

test/func_blocks/fu/test_fpu_rounding.py
Durchbruchswagen: Done. I refactored and parametrized tests

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
Durchbruchswagen: Done

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
Durchbruchswagen: This is the corner case for rounding subnormal numbers. Basically, in IEEE 754 standard, the 0 and 1 exponents represent the same exponent `exp_min` (`1-bias` if I remember correctly). The difference is that `exp = 1` indicates that the number is normal (the implicit bit is equal to 1) and `exp = 0` indicates that the number is in denormalized form (the implicit bit is equal to 0). The numbers with `exp = 0` are called subnormal (except 0). There are 3 possible results of rounding for a subnormal number: zero, subnormal number or smallest normal number. This condition is for the last case. I basically check if `exp == 0` and if the msb bit (the one that is implicit) is equal to one. In that case, I have to change the exponent to 1 to indicate that the number is now normal.

test/func_blocks/fu/test_fpu_rounding.py
```diff
@@ -0,0 +1,410 @@
...
+            input_values_dict["sig"] = help_values.qnan
+            input_values_dict["invalid_operation"] = 1
+            input_values_dict["division_by_zero"] = 0
+
+            resp = yield from request_adapter.call(input_values_dict)
+
+            assert resp["sign"] == input_values_dict["sign"]
+            assert resp["exp"] == input_values_dict["exp"]
+            assert resp["sig"] == input_values_dict["sig"]
+            assert resp["errors"] == 1
```
tilk, line 140: There is a lot of needlessly duplicated code here, which would make maintenance harder. You should use a loop over a list of test cases.

Durchbruchswagen: Done, there are now two arrays, one with input data and one with expected results, that are looped.

tilk (approved):

coreblocks/func_blocks/fu/fpu/fpu_error_module.py
```diff
@@ -0,0 +1,168 @@
...
+    """FPU error checking module layouts for methods
+
+    Parameters
+    ----------
+    fpu_params: FPUParams
+        FPU parameters
+    """
+
+    def __init__(self, *, fpu_params: FPUParams):
+        self.error_in_layout = [
```
piotro888, line 19: Usage of input_inf as in input to previous stages is not obvious by name, comment about that here could be helpful for future.

coreblocks/func_blocks/fu/fpu/fpu_common.py
```diff
@@ -0,0 +1,30 @@
...
+class RoundingModes(enum.Enum):
+    ROUND_UP = 3
+    ROUND_DOWN = 2
+    ROUND_ZERO = 1
+    ROUND_NEAREST_EVEN = 0
+    ROUND_NEAREST_AWAY = 4
+
+
+class FPUParams:
+    """FPU parameters
```
piotro888, line 13: Same thing here with documentation: it would be helpful to add information that `sig_width` contains the implicit bit.

piotro888 (approved):
Sorry for late review

tilk:
Waiting with merge until reviews are addressed.

coreblocks/func_blocks/fu/fpu/fpu_error_module.py
```diff
@@ -0,0 +1,169 @@
...
+                with m.Else():
+                    m.d.av_comb += final_exp.eq(arg.exp)
+                m.d.av_comb += final_sig.eq(arg.sig)
+                m.d.av_comb += final_sign.eq(arg.sign)
+
+            m.d.av_comb += final_errors[Errors.INVALID_OPERATION].eq(arg.invalid_operation)
+            m.d.av_comb += final_errors[Errors.DIVISION_BY_ZERO].eq(arg.division_by_zero)
+            m.d.av_comb += final_errors[Errors.OVERFLOW].eq(overflow)
+            m.d.av_comb += final_errors[Errors.UNDERFLOW].eq(underflow)
+            m.d.av_comb += final_errors[Errors.INEXACT].eq(inexact)
```
piotro888, line 160: final_errors must be now constructed from ORing `Error` fields.
I think it can be done like that for now `final_errors = Mux(arg.invalid_operation, Error.INVALID_OPERATION, 0) | ....`.

Extra note:
It is a common use case, and it cannot be done too cleanly, so  we should probably later create some helper method in transactron library for that usage pattern or maybe suggest feature to amaranth enums?

coreblocks/func_blocks/fu/fpu/fpu_error_module.py
piotro888: Please resolve comment from previous review

coreblocks/func_blocks/fu/fpu/fpu_rounding_module.py
```diff
@@ -0,0 +1,117 @@
...
+            ("rounding_mode", RoundingModes),
+        ]
+        self.rounding_out_layout = [
+            ("sig", fpu_params.sig_width),
+            ("exp", fpu_params.exp_width),
+            ("inexact", 1),
+        ]
+
+
+class FPUrounding(Elaboratable):
```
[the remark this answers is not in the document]
Durchbruchswagen: Yes, my bad. Typo is now fixed.

coreblocks/func_blocks/fu/fpu/fpu_error_module.py
Durchbruchswagen: Sorry for the late response. I pushed changes at around 3 a.m. and went to sleep with a plan to resolve comments next day, but by then you had already reviewed them. I was also thinking if maybe a better solution would be to write about this flag in some class for input layout for modules that perform arithmetic operations or something like that, but probably a small comment about why an error module needs it is warranted, so I added it.

coreblocks/func_blocks/fu/fpu/fpu_common.py
Durchbruchswagen: Done

coreblocks/func_blocks/fu/fpu/fpu_common.py
```diff
@@ -0,0 +1,38 @@
...
+    ROUND_NEAREST_EVEN = 0
+    ROUND_NEAREST_AWAY = 4
+
+
+class Errors(enum.IntFlag, shape=5):
+    INVALID_OPERATION = 0
+    DIVISION_BY_ZERO = 1
+    OVERFLOW = 2
+    UNDERFLOW = 3
+    INEXACT = 4
```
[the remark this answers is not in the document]
Durchbruchswagen: My bad, I hadn't thought about assigning powers of 2 to enums. Changed `Errors` according to your suggestion

coreblocks/func_blocks/fu/fpu/fpu_error_module.py
```diff
@@ -0,0 +1,169 @@
...
+            ("inexact", 1),
+            ("invalid_operation", 1),
+            ("division_by_zero", 1),
+            ("input_inf", 1),
+        ]
+        self.error_out_layout = [
+            ("sign", 1),
+            ("sig", fpu_params.sig_width),
+            ("exp", fpu_params.exp_width),
+            ("errors", Shape.cast(Errors)),
```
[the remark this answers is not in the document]
Durchbruchswagen: Done

test/func_blocks/fu/test_fpu_error.py
```diff
@@ -0,0 +1,304 @@
...
+                    "errors": 16,
+                },
+                # underflow
+                {"sign": 0, "sig": help_values.sub_norm_sig, "exp": 0, "errors": 24},
+                # invalid operation
+                {"sign": 0, "sig": help_values.qnan, "exp": help_values.max_exp, "errors": 1},
+                # division by zero
+                {"sign": 0, "sig": 0, "exp": help_values.max_exp, "errors": 2},
+                # overflow but no round and sticky bits
+                {"sign": 0, "sig": 0, "exp": help_values.max_exp, "errors": 20},
```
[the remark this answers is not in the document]
Durchbruchswagen: Done

coreblocks/func_blocks/fu/fpu/fpu_error_module.py
Durchbruchswagen: Done

coreblocks/func_blocks/fu/fpu/fpu_error_module.py
piotro888: I thought it was finished, sorry!

PR #729, opened 2024-09-21
User Mode and remaining mstatus fields

piotro888:


coreblocks/func_blocks/csr/csr.py
```diff
@@ -128,8 +129,11 @@ def elaborate(self, platform):
             | ((instr.exec_fn.funct3 == Funct3.CSRRCI) & (instr.s1_val != 0))
         )
 
-        # Temporary, until privileged spec is implemented
-        priv_level = Signal(PrivilegeLevel, reset=PrivilegeLevel.MACHINE)
+        priv_level = Signal(PrivilegeLevel)
+        with Transaction().body(m):
```
tilk, line 133: This is an always running transaction, right? We'll need a way to assert that easily in Transactron. For now, maybe state this in a comment?

coreblocks/priv/csr/csr_instances.py
```diff
@@ -78,6 +82,18 @@ def __init__(self, gen_params: GenParams):
         mepc_ro_bits = 0b1 if Extension.C in gen_params.isa.extensions else 0b11  # pc alignment (SPEC)
         self.mepc = CSRRegister(CSRAddress.MEPC, gen_params, ro_bits=mepc_ro_bits)
 
+        self.priv_mode = CSRRegister(
+            None,
+            gen_params,
+            width=2,
```
tilk, line 88: Not something for this PR, but what about accepting arbitrary shapes in CSR definitions? This would allow, e.g., using strongly typed `Enum` instead of weakly typed `IntEnum` more.
For now, I would write:
```suggestion
            width=Shape.cast(PrivilegeLevel),
```

coreblocks/priv/csr/csr_instances.py
```diff
@@ -87,6 +103,81 @@ def elaborate(self, platform):
...
+
+            return (legal, v)
+
+        # MIE bit - global interrupt enable
+        self.mstatus_mie = CSRRegister(None, gen_params, width=1)
+        mstatus.add_field(MstatusFieldOffsets.MIE, self.mstatus_mie)
+        # MPIE bit - previous MIE
+        self.mstatus_mpie = CSRRegister(None, gen_params, width=1)
+        mstatus.add_field(MstatusFieldOffsets.MPIE, self.mstatus_mpie)
+        # MPP bit - previous priv mode
```
tilk, line 125:
```suggestion
        # MPP field - previous priv mode
```

coreblocks/priv/csr/csr_instances.py
```diff
@@ -87,6 +103,81 @@ def elaborate(self, platform):
...
+        self.mstatus_mpp = CSRRegister(
+            None, gen_params, width=2, fu_write_filtermap=filter_legal_priv_mode, reset=PrivilegeLevel.MACHINE
```
tilk, line 127:
```suggestion
            None, gen_params, width=Shape.cast(PrivilegeLevel), fu_write_filtermap=filter_legal_priv_mode, reset=PrivilegeLevel.MACHINE
```

coreblocks/priv/csr/csr_instances.py
```diff
@@ -87,6 +103,81 @@ def elaborate(self, platform):
...
+
+        # Extension Context Status bits
+        # future todo: implement actual state modification tracking of F and V registers and CSRs
+        # State = 3 is DIRTY. Implementation is allowed to always set dirty for VS and FS, regardless of CSR updates
+        mstatus.add_read_only_field(MstatusFieldOffsets.VS, 2, 3 if Extension.V in gen_params.isa.extensions else 0)
+        mstatus.add_read_only_field(MstatusFieldOffsets.FS, 2, 3 if Extension.F in gen_params.isa.extensions else 0)
+        mstatus.add_read_only_field(MstatusFieldOffsets.XS, 2, 0)
+        # SD field - set to one when one of the states is dirty
+        if gen_params.isa.xlen == 32:
+            mstatush.add_read_only_field(
```
tilk, line 169: Shouldn't the SD field always be the last bit of `mstatus`, regardless of xlen?

https://michaelliao.github.io/riscv-isa/privileged.html#_machine_status_mstatus_and_mstatush_registers

coreblocks/priv/csr/csr_instances.py
```diff
@@ -87,6 +103,81 @@ def elaborate(self, platform):
...
+
+        # Fixed MXLEN/SXLEN/UXLEN = isa.xlen
+        if gen_params.isa.xlen == 64:
+            # Registers only exist in RV64
+            mstatus.add_read_only_field(MstatusFieldOffsets.UXL, 2, XlenEncoding.W64 if gen_params.user_mode else 0)
+            mstatus.add_read_only_field(MstatusFieldOffsets.SXL, 2, 0)
+
+        # Little-endianess
+        mstatus.add_read_only_field(MstatusFieldOffsets.UBE, 1, 0)
+        if gen_params.isa.xlen == 32:
```
tilk, line 139: There is a lot of fields defined in `mstatush`, and therefore there will be a lot of code duplication. Is it possible to always have an internal 64-bit CSR for, and alias its halves to `mstatus` and `mstatush`? This way, the only special case for position would be SD.

coreblocks/arch/csr_address.py
```diff
@@ -430,4 +435,30 @@ class CSRAddress(IntEnum, shape=12):
...
 
     # Internal Coreblocks CSRs
-    COREBLOCKS_TEST_CSR = 0x7FF  # used only for testbench verification
+    # used only for testbench verification
+    COREBLOCKS_TEST_PRIV_MODE = 0x8FF
+    COREBLOCKS_TEST_CSR = 0x7FF
+
+
+@unique
+class MstatusFieldOffsets(IntEnum):
```
lekcyjna123, line 444: Why there is no SD field?

coreblocks/arch/csr_address.py
```diff
@@ -430,4 +435,30 @@ class CSRAddress(IntEnum, shape=12):
     DSCRATCH1 = 0x7B3  # Debug scratch register 1.
 
     # Internal Coreblocks CSRs
-    COREBLOCKS_TEST_CSR = 0x7FF  # used only for testbench verification
+    # used only for testbench verification
+    COREBLOCKS_TEST_PRIV_MODE = 0x8FF
```
lekcyjna123, line 439: It should be somehow documented how this internal CSRs should be used.

coreblocks/priv/csr/csr_instances.py
```diff
@@ -87,6 +103,81 @@ def elaborate(self, platform):
...
+    def mstatus_fields_implementation(self, gen_params: GenParams, mstatus: AliasedCSR, mstatush: AliasedCSR):
+        def filter_legal_priv_mode(m: TModule, v: Value):
+            legal = Signal(1)
+            with m.Switch(v):
+                with m.Case(PrivilegeLevel.MACHINE):
+                    m.d.comb += legal.eq(1)
+                with m.Case(PrivilegeLevel.USER):
+                    m.d.comb += legal.eq(gen_params.user_mode)
+                with m.Default():
+                    pass
```
lekcyjna123, line 115: Shouldn't be `0`? So that Supervisor mode will be illegal?

coreblocks/arch/csr_address.py
piotro888: because `SD` field has variable position (last bit of mstatus different on 32/64), but now I added it with  value -1, so position is calculated from `MstatusFieldOffsets.SD % mstatus.width`

coreblocks/priv/csr/csr_instances.py
piotro888: `0` is a default value, updated to set it explicitely

coreblocks/priv/csr/csr_instances.py
piotro888: This is not possible with current aliasing system, left for refactor in future PR

coreblocks/priv/csr/csr_instances.py
piotro888: Good catch!

coreblocks/priv/csr/csr_instances.py
piotro888: I'm not sure if that would be too beneficial, as CSR can be written with arbitrary values from functional unit and typing could be wrong in some cases.

coreblocks/func_blocks/csr/csr.py
piotro888: Call moved to other internal transaction.

Side note: It could be relatively easily asserted with naming transaction and asserting `.grant`. But is the assertion really needed for single-use transaction only to read nonexclusive and always-ready by definition CSR method?

lekcyjna123: > Call moved to other internal transaction.

But this will make critical path longer :(

lekcyjna123 (approved):

coreblocks/func_blocks/csr/csr.py
tilk: > But is the assertion really needed for single-use transaction only to read nonexclusive and always-ready by definition CSR method?

It is not, but in time code gets modified. It could happen that someone added another, potentially blocking, method call to the transaction, which could then create some kind of a heisenbug. Assertions help protect against future mistakes.

Because it's not the first time something like this happens, maybe some short way for asserting this is needed - e.g. using a keyword argument for `Transaction().body`.

tilk: Why?

tilk (approved):

coreblocks/func_blocks/csr/csr.py
lekcyjna123: Because now it is under Switch.

tilk: But the output of the `read` method does not depend combinationally on the `run` signal.

PR #733, opened 2024-09-25
Update Amaranth to after RFC 58

tilk:
Amaranth update. It includes, among others:

* RFC 58 - ValueCastable formatting.
* RFC 51 - from_bits and data.Const.
* RFC 50 - Print statement.
* RFC 45 - lib.memory.
* RFC 43 - reset= to init=.
* RFC 27 - testbench functions. For now, I've avoided the `Settle` issue by silencing a deprecation warning. I've fixed all the other deprecation warnings however.

I intend to do Amaranth version updates in parts, so that the pull requests will be manageable. This PR unfortunately turned out to be quite large because of an Amaranth bug - see history below.

Somehow, synthesis results are better, and the reasons for that are unknown. Is this a result of Yosys or Amaranth upgrade? Is this a bug? I don't know.

Disclaimer: a lot of the new type stubs are untested, and some of them will probably need to be corrected in the future.

### DIFF
--- a/test/regression/cocotb.py
+++ b/test/regression/cocotb.py
@@ -159,7 +159,6 @@ class CocotbSimulation(SimulationBackend):
                 elif component[0] == "\\":
                     obj = obj._id(component[1:], extended=False)
                 else:
-                    print(dir(obj))
                     raise
 
         return obj
--- a/test/regression/cocotb.py
+++ b/test/regression/cocotb.py
@@ -153,10 +153,13 @@ class CocotbSimulation(SimulationBackend):
                 # function instead of 'getattr' - this is required by cocotb.
                 obj = obj._id(component, extended=False)
             except AttributeError:
-                # Try with escaped name
+                # Try with escaped or unescaped name
                 if component[0] != "\\" and component[-1] != " ":
                     obj = obj._id("\\" + component + " ", extended=False)
+                elif component[0] == "\\":
+                    obj = obj._id(component[1:], extended=False)
                 else:
+                    print(dir(obj))
                     raise
 
         return obj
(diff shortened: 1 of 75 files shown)


test/regression/cocotb.py
```diff
@@ -153,10 +153,13 @@ def get_cocotb_handle(self, path_components: list[str]) -> ModifiableObject:
...
                 obj = obj._id(component, extended=False)
             except AttributeError:
-                # Try with escaped name
+                # Try with escaped or unescaped name
                 if component[0] != "\\" and component[-1] != " ":
                     obj = obj._id("\\" + component + " ", extended=False)
+                elif component[0] == "\\":
+                    obj = obj._id(component[1:], extended=False)
                 else:
+                    print(dir(obj))
```
piotro888, line 162: Is this print intentional?

piotro888 (approved):
interesting bug

piotro888:
There is a dramatic change in FMax and utilisation (50MHz to 60 MHz; 25k to 15k) on basic. interesting....

and all benchmarks time out as well :(

test/regression/cocotb.py
tilk: No, that's a lost debug-print. Thanks for catching that.

tilk:
> There is a dramatic change in FMax and utilisation (50MHz to 60 MHz; 25k to 15k) on basic. interesting....
> Is that factor of improvement possible on yosys/amaranth update? or could it be synthesis bug?
> 
> and all benchmarks time out as well :(

Dang, more investigation is needed.

tilk:
It looks like the CPU locks up on a CSR instruction.

tilk:
The issue was in Amaranth, it was silently fixed in this commit: https://github.com/amaranth-lang/amaranth/commit/2d59242bf731b258e2f155fece933ae66acbe03b
There goes my plan for updating Amaranth in small steps.

lekcyjna123 (approved):

PR #735, opened 2024-10-08
Fix WFI resume on mstatus.MIE disabled

piotro888:
~~Depends on #729 (see last commit for diff)~~

WFI should continue to next instruction if interrupt is pending and unmasked via `mie`, but global interrupt enable `mstatus.MIE` bit is disabled.
[Priv spec 3.3.3]

Previously it only triggers interrupt handler when `mstatus.MIE` is set, and waits otherwise.


lekcyjna123 (approved):

piotro888:
rebased

tilk (approved):

Issue #598, opened 2024-03-05
Skip instructions on core flush in RSes


piotro888:
Implement solution 3. from discussion on Speculation Backend: https://github.com/kuznia-rdzeni/coreblocks/pull/557#issuecomment-1894099078

Would make solution progress in #498  - but works only for waiting operations in FUs that use RS.
Separate optimization would be needed (and complexity trade-off estimated) to cancel operations currently in flight/
Also similar mechanism needs to be implemented for LSU that defines its own internal RS. Implementing it in CSRUnit is probably not worth it, because it only waits for precommit -> only 1 additional cycle delay from backend.

PR #740, opened 2024-10-29
Make RS feed FUs with garbage if flushing

Arusekk:
See #598; does not *skip* FUs but shows the concept.


coreblocks/func_blocks/fu/common/rs.py
```diff
@@ -67,10 +70,14 @@ def elaborate(self, platform) -> TModule:
     def _elaborate(self, m: TModule, selected_id: Value, select_possible: Value, take_vector: Value):
         m.submodules += [self.perf_rs_wait_time, self.perf_num_full]
 
-        for i, record in enumerate(self.data):
-            m.d.comb += self.data_ready[i].eq(
-                ~record.rs_data.rp_s1.bool() & ~record.rs_data.rp_s2.bool() & record.rec_full.bool()
-            )
+        with Transaction(name="readiness").body(m):
+            precommit = self.dependency_manager.get_dependency(InstructionPrecommitKey())
```
piotro888, line 74: from meeting: replace precommit with `CoreStateKey`

piotro888:
[additional comments to discussion from meeting]

I checked that synchronous `flushing` signal would work in `RSInsertion` - because FreeRF/RF valid bits are also updated in sync domain, effect would be visible next cycle (and old RF entry inserted into RS). Change in `RSInsertion` would also not cause any performance loss. (and is definitely safe in RS too)

Proposition with resetting RF valid in Register Allocation would be problematic with checkpointing, that pushes new instruction immediately.

The last place is LSU:
LSU operations have a very high cost, I don't see why we should de-optimize it if this part of LSU is not on critical path (unless it is).

PR #743, opened 2024-11-04
Implement `misa` CSR

### LINKED ISSUE #561: Implement machine-mode CSRs
The CSRs in question are:

* [ ] `misa`
* [x] `mvendorid`
* [x] `marchid`
* [x] `mimpid`
* [x] `mhartid`
* [x] `mstatus` and `mstatush`
* [x] `mtvec`
* [x] `mip` and `mie`
* [x] `mcycle` and `minstret`
* [x] `mscratch`
* [x] `mepc`
* [x] `mcause`
* [x] `mtval`
* [x] `mconfigptr`

piotro888:
Closes #561
Complete machine mode CSRs :tada:

### DIFF
--- a/coreblocks/priv/csr/csr_instances.py
+++ b/coreblocks/priv/csr/csr_instances.py
@@ -208,8 +208,7 @@ class MachineModeCSRRegisters(Elaboratable):
         # 7 - Hypervisor, 18 - Supervisor, 23 - Custom Extensions
 
         xml_field_mapping = {32: XlenEncoding.W32, 64: XlenEncoding.W64, 128: XlenEncoding.W128}
-        if gen_params.isa.xlen in xml_field_mapping:
-            misa_value |= xml_field_mapping[gen_params.isa.xlen] << (gen_params.isa.xlen - 2)
+        misa_value |= xml_field_mapping[gen_params.isa.xlen] << (gen_params.isa.xlen - 2)
 
         return misa_value
 
--- a/coreblocks/priv/csr/csr_instances.py
+++ b/coreblocks/priv/csr/csr_instances.py
@@ -84,6 +84,10 @@ class MachineModeCSRRegisters(Elaboratable):
 
         self.mtval = CSRRegister(CSRAddress.MTVAL, gen_params)
 
+        self.misa = CSRRegister(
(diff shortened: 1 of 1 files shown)


coreblocks/priv/csr/csr_instances.py
```diff
@@ -179,6 +183,36 @@ def filter_legal_priv_mode(m: TModule, v: Value):
...
+        for bit, extension in misa_extension_bits.items():
+            if extension in gen_params.isa.extensions:
+                misa_value |= 1 << bit
+
+        if gen_params.user_mode:
+            misa_value |= 1 << 20
+        # 7 - Hypervisor, 18 - Supervisor, 23 - Custom Extensions
+
+        xml_field_mapping = {32: XlenEncoding.W32, 64: XlenEncoding.W64, 128: XlenEncoding.W128}
+        if gen_params.isa.xlen in xml_field_mapping:
```
tilk, line 211: We already verify that `xlen` is one of 32, 64, 128 (in `ISA`), so this condition should be always true - and if it's not, I believe it's a bug.

tilk (approved):

lekcyjna123 (approved):

coreblocks/priv/csr/csr_instances.py
```diff
@@ -179,6 +183,36 @@ def filter_legal_priv_mode(m: TModule, v: Value):
...
+            12: Extension.M,
+            16: Extension.Q,
+            21: Extension.V,
+        }
+
+        for bit, extension in misa_extension_bits.items():
+            if extension in gen_params.isa.extensions:
+                misa_value |= 1 << bit
+
+        if gen_params.user_mode:
```
lekcyjna123, line 206: Why user_mode is handled separately? From ISA point of view it is "U" extension.

piotro888: It is treated as extension only in `misa`. It is not present in isa strings, not defined as extension in spec, doesn't define instructions and only changes CSR behaviour (and can't be determined from FU config, so needs config flag anyway)
I think it is better to not add it to isa.extensions

coreblocks/priv/csr/csr_instances.py
piotro888: removed

Issue #745, opened 2024-11-04
Implement vectored interrupt mode in `mtvec`


piotro888:
Defined in 3.1.7. Machine Trap-Vector Base-Address Register

Currently only direct mode is supported, where all traps set pc to `mtvec` value. 
Vectored mode uses different pc offsets for trap cause codes

This is an optional feature.

PR #755, opened 2024-11-19
mtvec vectored mode

kalinf:
Implement vectored interrupt mode in mtvec (issue #745)

### DIFF
--- a/coreblocks/priv/csr/csr_instances.py
+++ b/coreblocks/priv/csr/csr_instances.py
@@ -99,7 +99,6 @@ class MachineModeCSRRegisters(Elaboratable):
         self._mstatus_fields_implementation(gen_params, self.mstatus, self.mstatush)
         self._mtvec_fields_implementation(gen_params, self.mtvec)
 
-
     def elaborate(self, platform):
         m = Module()
 


test/asm/interrupt_vectored.asm
```diff
@@ -0,0 +1,141 @@
...
+
+fail:
+    csrwi 0x7ff, 2
+    j fail
+
+
+.org 0x200
+    nop
+    nop
+    nop
```
piotro888, line 123: It would be nice to check if any synchronous exception (like ex. ecall) jumps to 0x200 + 0x0 handler. 

It could be a separate simple handler that only sets some register for later verification and directly returns from handler; triggered with one ecall before enabling external interrupts.

test/asm/interrupt_vectored.asm
```diff
@@ -0,0 +1,141 @@
...
+    # fibonacci step
+    beq x3, x15, skip
+    add x3, x2, x1
+    mv x1, x2
+    mv x2, x3
+    # store state
+    mv x12, x1
+    mv x13, x2
+    mv x14, x3
+    j skip
```
piotro888, line 113: This a common section int0_handler, it can just jump inside there to avoid code duplication.

test/asm/interrupt_vectored.asm
```diff
@@ -0,0 +1,141 @@
...
+    nop
+    nop
+    nop
+    nop
+    nop
+    nop
+    nop
+    j int0_handler
+    j int1_handler
+    li x31, 0xae  # should never happen
```
piotro888, line 139:
```suggestion
    nop
    j fail
    j int0_handler
    j int1_handler
    li x31, 0xae  # should never happen
```
Jumping to any earlier handler address should be also guarded, now it would be not detected and complete just fine by executing nops and then jump to `int0_handler`

test/asm/interrupt_vectored.asm
kalinf: Two handlers use different registers, that stores global state checked in the test, only 4 middle lines are the same. I shortened it by removing the load state and store state sections

coreblocks/backend/retirement.py
```diff
@@ -213,8 +213,16 @@ def flush_instr(rob_entry):
                     self.perf_trap_latency.stop(m)
 
                     handler_pc = Signal(self.gen_params.isa.xlen)
-                    # mtvec without mode is [mxlen-1:2], mode is two last bits. Only direct mode is supported
-                    m.d.av_comb += handler_pc.eq(m_csr.mtvec.read(m).data & ~(0b11))
+                    mtvec_offset = Signal(self.gen_params.isa.xlen)
+                    mtvec = m_csr.mtvec._fu_read(m).data
```
piotro888, line 217: `_fu_read` should not be used anywhere apart `CSRUnit`.
I see that there is no `read` implemented, but in this case it would be better using `mtvec_base` and `mtvec_mode` anyway.

coreblocks/backend/retirement.py
```diff
@@ -213,8 +213,16 @@ def flush_instr(rob_entry):
...
+                    mcause = m_csr.mcause.read(m).data
+
+                    # mtvec without mode is [mxlen-1:2], mode is two last bits.
+                    # When mode=1 (Vectored), interrupts set pc to base + 4 * cause_number
+                    with m.If(mcause[-1] & (mtvec[0:1] == 1)):
```
piotro888, line 222: `TrapVectorMode.VECTORED` can be now used for mode comparison

coreblocks/priv/csr/csr_instances.py
```diff
@@ -105,6 +104,19 @@ def elaborate(self, platform):
 
         return m
 
+    def mtvec_fields_implementation(self, gen_params: GenParams, mtvec: AliasedCSR):
```
piotro888, line 107: There is a pending merge conflict changing `mstaus_fields...` to `_mstatus_fields...`.
Please resolve the conflict and also update this function to `_mtvec_fields...`

piotro888 (commented):
Looks good to merge after resolving comments, thanks!

piotro888 (approved):

tilk (approved):

PR #758, opened 2024-11-20
External Transactron

tilk:
This PR finally extracts Transactron to external repo, https://github.com/kuznia-rdzeni/transactron. Depends on #754.

In the same time, the `amaranth-stubs` submodule is removed and replaced by an entry in `requirements.txt`. My reasons are:
1. Installing local packages is not great when `pyproject.toml` is used. I'd like to move Coreblocks to `pyproject.toml` - I already did that in transactron.
2. Workflows can be simplified, as the submodule doesn't need to be fetched before dependency installation.
3. There are no real negative consequences for development - the local `amaranth-stubs` version needed to be installed using `pip` after each change anyway.


piotro888 (approved):

piotro888:
Coreblocks README needs updating; and maybe docs/transactions.md should be preserved, but only with url to Transactron gh-pages docs?

lekcyjna123 (approved):

tilk:
> Would it make sense to rebase transactron repository on top of something like `git-filter-repo` (note: invocation is destructive!) with ` --path transactron/ --path test/transactron/ --path-rename test/transactron/:test/ --path coreblocks/transactions --replace-message <replace file with "#==>https://github.com/kuznia-rdzeni/coreblocks/pull/">` to keep easier to get history with valid PR references in one place?

I just did that, but it unfortunately invalidated all active PRs in Transactron repo.

PR #762, opened 2024-11-26
Delay flush by one cycle

tilk:
This PR simplifies retirement by delaying the start of flushing. This eliminates one use of `condition` and makes the `flushing` signal registered. Looking at benchmarks:

* IPC ~is not significantly impacted~ somehow didn't change at all.
* Fmax did not visibly improve (sadly), but:
* The critical path is now different.
* Transactron reports one less transaction and three less methods.

### DIFF
--- a/coreblocks/backend/retirement.py
+++ b/coreblocks/backend/retirement.py
@@ -114,7 +114,6 @@ class Retirement(Elaboratable):
 
         continue_pc_override = Signal()
         continue_pc = Signal(self.gen_params.isa.xlen)
-        core_flushing = Signal()
 
         with m.FSM("NORMAL") as fsm:
             with m.State("NORMAL"):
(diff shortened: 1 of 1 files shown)


coreblocks/backend/retirement.py
```diff
@@ -180,15 +178,8 @@ def flush_instr(rob_entry):
...
-                    with condition(m) as cond:
-                        with cond(commit):
-                            retire_instr(rob_entry)
-                        with cond():
-                            # Not using default condition, because we want to block if branch is not ready
-                            flush_instr(rob_entry)
-
-                            m.d.comb += core_flushing.eq(1)
+                    with m.If(commit):
+                        retire_instr(rob_entry)
```
piotro888, line 182: This is not valid - when instruction is an exception with `commit=0` (like synchronous exceptions),
RF is not freed (and not pushed to Free RF), but instruction is retired from ROB, and TRAP_FLUSH would flush next instruction.

tilk: Indeed, you're right - this was not thought out enough. Still, this means we have a blind spot in tests for this.

PR #765, opened 2024-11-27
Synchronous register file using MemoryBank

### LINKED ISSUE #737: Register file using synchronous memory
Efficient register file implementation on FPGAs requires the use of block RAM, which is synchronous. Currently register reads are asynchronous: the register values are read during the RS insertion phase. Synchronous register reads could be implemented using a `MemoryBank` with reads initiated during RS selection and completed during RS insertion.

To materialize the gain, #734 needs to be implemented, as the register file has three ports - too much for a single block RAM.

tilk:
This PR implements a synchronous register file. Read requests are done on RS selection, results are read, like previously, on RS insertion. The change will enable using FPGA Block RAMs and multiport RAM structures utilizing block RAMs.

Closes #737.

TODO:
- [x] Merge Transactron PR https://github.com/kuznia-rdzeni/transactron/pull/18, change commit in dependencies.


piotro888 (approved):
lgtm

PR #770, opened 2024-12-05
De-Wishbonized tests

tilk:
LSU and ICache are using an abstract bus interface, so going down to a physical bus is not required. This PR removes the Wishbone dependency in LSU and ICache tests. Also, a PMA test is fixed, I broke it by accident while porting tests.


piotro888 (approved):

Arusekk (approved):
Looks like no rocket science.

PR #774, opened 2024-12-10
Remove func blocks unifier indirections

tilk:
This PR simplifies the announcement mechanism using the dependency system. At the same time, two layers of `Collector`s for accepting results were flattened to a single collector. The `func_blocks_unifier` module became trivial, and it might make sense to remove it later.

Benchmark results dropped slightly for some reason, ~but device utilization also seems to be reduced~.

### DIFF
--- a/coreblocks/func_blocks/fu/common/rs_func_block.py
+++ b/coreblocks/func_blocks/fu/common/rs_func_block.py
@@ -28,9 +28,6 @@ class RSFuncBlock(FuncBlock, Elaboratable):
         RS select method.
     update: Method
         RS update method.
-    get_result: Method
-        Method used for getting single result out of one of the FUs. It uses
-        layout described by `FuncUnitLayouts`.
     """
 
     def __init__(
@@ -67,7 +64,6 @@ class RSFuncBlock(FuncBlock, Elaboratable):
         self.insert = Method(i=self.rs_layouts.rs.insert_in)
         self.select = Method(o=self.rs_layouts.rs.select_out)
         self.update = Method(i=self.rs_layouts.rs.update_in)
-        self.get_result = Method(o=self.fu_layouts.accept)
 
     def elaborate(self, platform):
         m = TModule()
--- a/coreblocks/func_blocks/fu/common/rs_func_block.py
+++ b/coreblocks/func_blocks/fu/common/rs_func_block.py
@@ -1,6 +1,8 @@
 from collections.abc import Collection, Iterable
 from amaranth import *
 from dataclasses import dataclass
+
+from transactron.utils import DependencyContext
 from coreblocks.params import *
 from .rs import RS, RSBase
 from coreblocks.scheduler.wakeup_select import WakeupSelect
@@ -9,6 +11,7 @@ from coreblocks.func_blocks.interface.func_protocols import FuncUnit, FuncBlock
 from transactron.lib import Collector
 from coreblocks.arch import OpType
 from coreblocks.interface.layouts import RSLayouts, FuncUnitLayouts
+from coreblocks.interface.keys import FuncUnitResultKey
 
 __all__ = ["RSFuncBlock", "RSBlockComponent"]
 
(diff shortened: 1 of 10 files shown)


test/backend/test_annoucement.py
```diff
@@ -116,30 +113,24 @@ async def producer(sim: TestbenchContext):
 
     async def consumer(self, sim: TestbenchContext):
         # TODO: this test doesn't do anything, fix it!
```
piotro888, line 115: Is this comment up-to-date?

coreblocks/func_blocks/fu/common/rs_func_block.py
```diff
@@ -87,12 +89,9 @@ def elaborate(self, platform):
             m.submodules[f"func_unit_{n}"] = func_unit
             m.submodules[f"wakeup_select_{n}"] = wakeup_select
 
-        m.submodules.collector = collector = Collector([func_unit.accept for func_unit, _ in self.func_units])
-
         self.insert.proxy(m, self.rs.insert)
         self.select.proxy(m, self.rs.select)
         self.update.proxy(m, self.rs.update)
-        self.get_result.proxy(m, collector.method)
```
piotro888, line 95: `self.get_result` should be removed (+ in docstring too)

piotro888 (approved):

test/backend/test_annoucement.py
tilk: I believe so. The condition in `while` looks to be false in the beginning. Maybe a negation was intended there?

coreblocks/func_blocks/fu/common/rs_func_block.py
tilk: Done.

test/backend/test_annoucement.py
tilk: After adding the negation, the test fails. Created issue #775 for this.

lekcyjna123:
The Fmax drop is a little bit worrying and it looks like the FuncBlockUnifier is on critical path now. I checked the synthesis results (https://github.com/kuznia-rdzeni/coreblocks/actions/runs/12359494095/job/34492356220) and it looks like:
- In sync->sync most of 30ns critical path is travelling between different parts of unifier.
- We should add registers on wishbone input from gpio. It takes 13ns to route data from GPIO and this impact LSU scheduling.
- There is a path from wishbone GPIO, LSU, FuncUnitResultKey_unifier, RF to the CSR. All that in one cycle.

coreblocks/func_blocks/interface/func_blocks_unifier.py
```diff
@@ -18,19 +17,10 @@ def __init__(
     ):
         self.rs_blocks = [(block.get_module(gen_params), block.get_optypes()) for block in blocks]
 
-        self.result_collector = Collector([block.get_result for block, _ in self.rs_blocks])
```
lekcyjna123, line 21: If I correctly see, removing that Collector cause that all FUs get_results methods are joined with the announcement methods (so with RS and RF), which make scheduling more complex and critical path longer. In Collector there is hidden a Forwarder which cut the critical path on data.

coreblocks/interface/keys.py
```diff
@@ -52,6 +52,16 @@ class FetchResumeKey(UnifierKey, unifier=Collector):
     pass
 
 
+@dataclass(frozen=True)
```
lekcyjna123, line 55: Maybe we should start adding the doc strings to our keys? In practice they are a global variables and we haven't documented them...

coreblocks/func_blocks/fu/common/rs_func_block.py
```diff
@@ -87,12 +85,9 @@ def elaborate(self, platform):
             m.submodules[f"func_unit_{n}"] = func_unit
             m.submodules[f"wakeup_select_{n}"] = wakeup_select
 
-        m.submodules.collector = collector = Collector([func_unit.accept for func_unit, _ in self.func_units])
```
lekcyjna123, line 90: This also complicates the transactron network. Probably its is connected with observed IPC drop. Previously, when two results were ready in the same cycle, one have been announced and second stored in Forwarder for a cycle, what made the FU ready to process the next instruction. Now the FU have to stall till it can push out its result.

lekcyjna123 (changes requested):
LNGTM

The changes are affecting Fmax and make transactron network more complicated. Additionaly they removed buffers in announcement.

tilk:
Invalid since #850 (or maybe earlier).

PR #778, opened 2025-01-09
Renode support

Arusekk:
This aims to aid serious software development when FPGAs are unavailable.
Does not work yet, because of lack of debugging interface (required to set initial PC).

But does it otherwise work? No idea. 😄

Feedback welcome.

### DIFF
--- a/renode/README.md
+++ b/renode/README.md
@@ -1,0 +1,8 @@
+# Coreblocks in Renode (via Verilator)
+
+Code based on [`verilated_ibex.resc`](https://github.com/renode/renode/blob/master/scripts/single-node/verilated_ibex.resc) and official Renode samples from https://github.com/antmicro/renode-verilator-integration
+
+Unfortunately needs Renode built from source, does not work in installed environment (https://github.com/renode/renode/issues/726).
+
+Does not work yet, hangs on setting PC to 0 (no debug interface).
+
--- a/renode/build_plugin.sh
+++ b/renode/build_plugin.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+# Usage: renode/build_plugin.sh core.v
+
+set -e
+
+ORIGIN=$(dirname "$0")
+
+if [ -z "$RENODE_ROOT" ]; then
+  echo "RENODE_ROOT not set; trying to guess from PATH"
+  RENODE_ROOT=$(renode --disable-gui --console -e 'python "import os;print(os.getcwd());os._exit(0)"' |tail -1)
+fi
+COSIM_DIR="$RENODE_ROOT/plugins/VerilatorIntegrationLibrary"
+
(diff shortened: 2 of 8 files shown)


Arusekk:
Alright, it is now integrated correctly, and can even dump traces!

The VCD traces and debug log suggest that we supply `0x0, 0x1, 0x2, ...` instead of `0x0, 0x4, 0x8, ...` to wishbone. And it responds with `0x2117A009, 0x002117A0, 0x01002117, 0x13010021, 0x01130100, ...` instead of `0x2117A009, 0x01130100, ...`. Any ideas why?

piotro888:
This is intentional for Wishbone bus, done ex. here  https://github.com/kuznia-rdzeni/coreblocks/blob/a2d56a57831323b2ae0ef5c8aa93a915669f1cef/coreblocks/func_blocks/fu/lsu/lsu_requester.py#L134.


I cant

piotro888:
At the same time this is in spec
```
ADR_I()

The address input array [ADR_I()] is used to pass a binary address. The higher array boundary is specific to the address width of the core, and the lower array boundary is determined by the data port size. For example the array size on a 32-bit data port with BYTE granularity is [ADR_O(n..2)]. In some cases (such as FIFO interfaces) the array may not be present on the interface.
```
I'm confused now

piotro888:
Oh, that makes sense now, they are referring to address bit array (address signal) size and not memory size. It could've been written clearer.
So Wishbone is currently addressed correctly (word addressed, smaller accesses selected via SEL_I)

Arusekk:
With https://github.com/renode/renode/pull/727 it does work now! Even with the same unmodified Ibex LiteX BIOS (verified on `full` core config, since that BIOS needs compressed instructions).

The below should work:
```sh
cd
git clone -b wishbone-adr https://github.com/Arusekk/renode
cd renode
./build.sh
export PATH=$PATH:$HOME/renode
cd ~/coreblocks/renode
../scripts/gen_verilog.py -c full
./build_plugin.sh --trace core.v
./verilated_coreblocks.resc
```
Then type `start` or `s` in Renode window. Voilà!

piotro888:
Thats great! :+1: 

at the same time I see that addressing shift is already adjusted in Renode in wishbone.cpp here https://github.com/renode/renode/blob/bf7eaf6578b2ab3734f7ce55a6c647ae3115c539/src/Plugins/CoSimulationPlugin/IntegrationLibrary/src/buses/wishbone.cpp#L57 
Is this a separate code path from wishobne-initiator or issue is somewhere else?

Arusekk:
If I understand correctly, wishbone.cpp is only used for wishbone peripherals (slaves), and wishbone-initiator.h only for initiators (masters). The funny part is that my build script does not even compile wishbone.cpp code. 😄

Arusekk:
Note: separate repo

PR #779, opened 2025-01-13
Implement Zaamo - atomic memory operations extension

piotro888:
Implements Atomic Memory Operations (part of A extension), that do read/modify/write operations on memory.

It is implemented as DummyLSU wrapper, that itself is another `FuncUnit` that forwards LOAD/STORE instructions, but switches to atomic mode on AMO instructions and internally issues LOAD to LSU, then does arithmetic operation (via own ALU), issue STORE and produces final result.

Atomic requirements are more relaxed in our case because (a) there is only one core, (b) LSU doesn't reoder accesses. Only problem to solve was executing multiple operations one one instruction. I selected this approach instead of expanding to multiple instructions, because that would require additional temporary registers.
Both requirements can be approached (with wrapper solution) in future - (a) with informing LSU that it should lock target on bus (and add bus that supports that like AXI4) and (b) with the same mechanism as FENCE.

TODO: docstring

### DIFF
--- a/coreblocks/arch/isa.py
+++ b/coreblocks/arch/isa.py
@@ -59,6 +59,8 @@ class Extension(enum.IntFlag):
     ZIHPM = auto()
     #: Integer conditional operations
     ZICOND = auto()
+    #: Atomic memory operations
+    ZAAMO = auto()
     #: Misaligned atomic operations
     ZAM = auto()
     #: Half precision floating-point operations (16-bit)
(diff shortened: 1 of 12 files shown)


piotro888:
Is there a reason why `DummyLSU.issue` is `single_caller`?

piotro888:
For future PR: I'm not sure what to do with MMIO regions. Raise exception? Ignore atomic requirements? or is it required to work the same?

tilk (approved):

xThaid (approved):

PR #780, opened 2025-01-16
Older Ubuntu in workflows

tilk:
Quick and dirty fix for the breakage caused by Github Ubuntu version upgrade, see https://github.com/actions/runner-images/issues/10636.

In the long run, we should upgrade the Ubuntu base image in our Docker images too.


piotro888 (approved):

tilk:
Merged with bypass.

PR #782, opened 2025-01-16
Implement Zalrsc - load-reserved/store-conditional instructions extension

piotro888:
Very simplified implementation, because there is only one hart (reservation set size == full address space) -> only matches LR to first SC.
Now we have complete A extension!

~~Depends on #779 (see last commit for diff)~~


xThaid:
solid work! have you tried running `rv32ua` from the `riscv-tests` test suite?

piotro888:
Welcome back!! :)

I planned to do that in separate PR, I can test on in the evening.

tilk (approved):

piotro888:
AMO tests from `riscv-tests` pass.
LRSC test needed to be excluded, because it contains 1024 memory stores that time-out, but I manually verified that it passes 3 first subtests of this test.

tilk (approved):

coreblocks/func_blocks/fu/lsu/lsu_atomic_wrapper.py
```diff
@@ -129,7 +171,22 @@ def _(arg):
...
+                m,
+                {
+                    "result": 1,
+                    "rob_id": atomic_op.rob_id,
+                    "rp_dst": atomic_op.rp_dst,
+                    "exception": 0,
+                },
+            )
+
+        sc_failed_trans.add_conflict(push_lsu_result, priority=Priority.RIGHT)
```
lekcyjna123, line 189: Why you don't add `~sc_failed` as a condition to run `push_lsu_result`? This should have a similar result but without complicating transaction graph.

piotro888: Why is this a problem? `sc_failed_trans` would still conflict with `push_lsu_result` because of `push_result` calling. This is only setting priority to make pending previously issued `LOAD` results to be prioritized over SC failed, which would be reversed in your proposition.

lekcyjna123: I haven't seen that. Approved.

PR #783, opened 2025-01-16
Age-ordered RS

### LINKED ISSUE #290: Improve wakeup-select
As for now our wakeup-select logic is very simply (take the last ready instruction - last in sense of index of RS), so in future it would be great to make it more intelligent.

tilk:
This PR changes RS so that oldest entries are taken first. The idea is that the indexes in `get_ready_list` and `take` are no longer physical RS row numbers, but they are ordered from oldest to newest. The priority encoder in `WakeupSelect` then naturally selects the oldest ready entry.

The main RS test was extended to introduce random delays and testing multiple ready lists. The individual method tests are dropped, for three reasons:

1. These tests don't just verify conformance to some protocol, they need to look into RS internals to work. Because of that, changes to internals invalidate these tests, creating more work for the developer.
2. There is quite a lot of lines of code in these tests, which increases the amount of maintenance work even more.
3. The main test seems good enough, and even if some error is not found there, the tests for the whole core will most probably fail then.

Fixes #290.


lekcyjna123 (approved):
Approved, but I have mixed feelings about this PR. I don't like the fact that RS is reordering ids so that WakeupSelect can choose first one. This break an abstraction, where RS is container and WakeupSelect an algorithm to choose from that container. I would rather have the order presented as part of RS state, so that WakeupSelect can do the decision which element to choose.

Nevertheless I think that it can be changed in the future refactor, so approved.

tilk:
Fixes #290.

coreblocks/func_blocks/fu/common/rs.py
```diff
@@ -28,8 +30,7 @@ def __init__(
         ready_for = ready_for or ((op for op in OpType),)
         self.gen_params = gen_params
         self.rs_entries = rs_entries
-        self.rs_entries_bits = (rs_entries - 1).bit_length()
-        self.layouts = gen_params.get(RSLayouts, rs_entries_bits=self.rs_entries_bits)
+        self.layouts = gen_params.get(RSLayouts, rs_entries=self.rs_entries)
         self.internal_layout = make_layout(
             ("rs_data", self.layouts.rs.data_layout),
             ("rec_full", 1),
```
piotro888, line 36: `rec_reserved` is not used

coreblocks/func_blocks/fu/common/rs.py
```diff
@@ -50,75 +51,83 @@ def __init__(
...
             m.d.sync += record.rec_full.eq(0)
-            self.perf_rs_wait_time.stop(m, slot=rs_entry_id)
+            self.perf_rs_wait_time.stop(m, slot=actual_rs_entry_id)
             out = Signal(self.layouts.take_out)
             m.d.av_comb += assign(out, record.rs_data, fields=AssignType.COMMON)
+            self.log.debug(m, True, "taken entry {} at idx {}", actual_rs_entry_id, rs_entry_id)
             return out
 
         for get_ready_list, ready_list in zip(self.get_ready_list, ready_lists):
+            reordered_list = Cat(ready_list.bit_select(self.order[i], 1) for i in range(self.rs_entries))
```
piotro888, line 123: Shouldn't `ready_list` be anded with `takeable_mask` here too?

coreblocks/func_blocks/fu/common/rs.py
```diff
@@ -50,75 +51,83 @@ def __init__(
...
         )
+        self.log = logging.HardwareLogger(f"backend.rs.{rs_number}")
 
     @abstractmethod
     def elaborate(self, platform) -> TModule:
         raise NotImplementedError
 
-    def _elaborate(self, m: TModule, selected_id: Value, select_possible: Value, take_vector: Value):
+    def _elaborate(self, m: TModule, takeable_mask: ValueLike, alloc: Method, free_idx: Method, order: Method):
         m.submodules += [self.perf_rs_wait_time, self.perf_num_full]
```
piotro888, line 70: Docstring for this _elaborate function and interface would be very helpful

coreblocks/func_blocks/fu/common/rs.py
```diff
@@ -50,75 +51,83 @@ def __init__(
...
-        @def_method(m, self.select, ready=select_possible)
+        @def_method(m, self.select)
         def _() -> RecordDict:
-            m.d.sync += self.data[selected_id].rec_reserved.eq(1)
+            selected_id = alloc(m).ident
+            self.log.debug(m, True, "selected entry {}", selected_id)
             return {"rs_entry_id": selected_id}
 
+        @def_method(m, self.update)
+        def _(reg_id: Value, reg_val: Value) -> None:
```
tilk, line 89: move needed? (update vs insert order)

tilk: The move was done to simplify the update logic - now the insert method overrides the update.

tilk:
> Approved, but I have mixed feelings about this PR. I don't like the fact that RS is reordering ids so that WakeupSelect can choose first one. This break an abstraction, where RS is container and WakeupSelect an algorithm to choose from that container. I would rather have the order presented as part of RS state, so that WakeupSelect can do the decision which element to choose.

This was done so that the deallocation logic is simpler. The `free_idx` method of the `PreservedOrderAllocator` is simpler than `free` (see https://github.com/kuznia-rdzeni/transactron/blob/a2b7d33adad51b8500dc942d92082e2cfc190cdd/transactron/lib/allocators.py#L119). In this PR I had to decide between circuit size and preserving the old behavior, and I chose circuit size. But to be honest, I didn't measure the difference - maybe it doesn't really matter?

piotro888 (approved):

PR #786, opened 2025-01-18
Remove constructor passed methods

tilk:
This PR replaces passing methods to constructors with `proxy`. The module requiring external methods needs to declare their inputs/outputs. This gives better error messages when interfaces mismatch and simplifies mocking in tests. Code size is also somewhat reduced.

Refactored modules are `Retirement`. `ResultAnnouncement`, `WakeupSelect` and `DecodeStage`. The `Scheduler` is not refactored to avoid conflicts with the checkpointing branch.


test/frontend/test_decode_stage.py
```diff
@@ -1,82 +1,85 @@
...
 
-        self.decode = DecodeStage(self.gen_params, fifo_in.read, fifo_out.write)
         self.m = SimpleTestCircuit(
             ModuleConnector(
                 decode=self.decode,
-                fifo_in=fifo_in,
-                fifo_out=fifo_out,
-                fifo_in_write=self.fifo_in_write,
-                fifo_out_read=self.fifo_out_read,
+                fifo_in_write=self.push_raw,
```
lekcyjna123, line 63: Shouldn't be `fifo_in_write` renamed to `push_raw`? Similarly with `out` below.

lekcyjna123 (approved):

coreblocks/backend/annoucement.py
```diff
@@ -17,34 +18,37 @@ class ResultAnnouncement(Elaboratable):
...
     `ManyToOneConnectTrans` to a FIFO.
+
+    Attributes
+    ----------
+    get_result : Method
+        Method which is invoked to get results of next ready instruction,
+        which should be announced in core. This method assumes that results
+        from different FUs are already serialized.
+    rob_mark_done : Method
+        Method which is invoked to mark that instruction ended without exception.
```
piotro888, line 29:
```suggestion
        Method which is invoked to mark that instruction ended.
```

piotro888 (approved):
(For future) I think that some clear distinction between methods declared in module and methods that are expected to be connected by proxy should be made. This can be a little bit confusing now.

test/frontend/test_decode_stage.py
tilk: Fixed.

tilk:
I'll wait with merging to not interfere with #796.

PR #788, opened 2025-01-21
Close path for fpu adder/subtractor

Durchbruchswagen:
Implementation of close path for fpu adder/subtractor based on:
http://i.stanford.edu/pub/cstr/reports/csl/tr/90/442/CSL-TR-90-442.pdf.


coreblocks/func_blocks/fu/fpu/close_path.py
```diff
@@ -0,0 +1,214 @@
...
+        shift_amount_lza_zero = Signal(range(self.params.sig_width))
+        shift_amount_lza_one = Signal(range(self.params.sig_width))
+        shift_amount = Signal(range(self.params.sig_width))
+        bit_shift_amount = Signal(range(self.params.sig_width))
+        check_shift_amount = Signal(range(self.params.sig_width))
+        shifted_sig = Signal(self.params.sig_width)
+        shifted_exp = Signal(self.params.sig_width)
+        final_sig = Signal(self.params.sig_width)
+        final_exp = Signal(self.params.exp_width)
+        final_round = Signal(1)
```
tilk, line 90: `Signal(1)` can be shortened to `Signal()`.

coreblocks/func_blocks/fu/fpu/close_path.py
```diff
@@ -0,0 +1,214 @@
...
+
+            with m.If(is_zero):
+                m.d.av_comb += final_sig.eq(final_result)
+                m.d.av_comb += final_exp.eq(0)
+                m.d.av_comb += final_round.eq(guard_bit)
+            with m.Elif(exp == 0):
+                m.d.av_comb += final_sig.eq(final_result)
+                m.d.av_comb += final_exp.eq(0)
+                m.d.av_comb += final_round.eq(guard_bit)
+
```
tilk, line 168: Confusing empty line - the next one has an `Elif`.

coreblocks/func_blocks/fu/fpu/close_path.py
```diff
@@ -0,0 +1,214 @@
...
+                with m.If(exp == 1):
+                    m.d.av_comb += final_sig.eq(final_result)
+                    m.d.av_comb += final_round.eq(guard_bit)
+                with m.Else():
+                    m.d.av_comb += shift_amount.eq((exp - 1))
+                    m.d.av_comb += bit_shift_amount.eq((exp - 2))
+                    m.d.av_comb += final_sig.eq((final_result << shift_amount) | (shift_in_bit << bit_shift_amount))
+                    m.d.av_comb += final_round.eq(0)
+                m.d.av_comb += final_exp.eq(0)
+
```
tilk, line 179: Confusing empty line.

coreblocks/func_blocks/fu/fpu/close_path.py
```diff
@@ -0,0 +1,214 @@
...
+                with m.If(shift_correction):
+                    with m.If(correct_shift == 0):
+                        m.d.av_comb += final_sig.eq(shifted_sig)
+                        m.d.av_comb += final_round.eq(guard_bit)
+                    with m.Else():
+                        m.d.av_comb += bit_shift_amount.eq(correct_shift - 1)
+                        m.d.av_comb += final_sig.eq(shifted_sig | (shift_in_bit << bit_shift_amount))
+                        m.d.av_comb += final_round.eq(0)
+                    m.d.av_comb += final_exp.eq(shifted_exp)
+
```
tilk, line 197: Confusing empty line.

coreblocks/func_blocks/fu/fpu/close_path.py
```diff
@@ -0,0 +1,214 @@
...
+                with m.Else():
+                    m.d.av_comb += final_round.eq(0)
+                    with m.If(shifted_exp == 1):
+                        with m.If(correct_shift > 0):
+                            m.d.av_comb += bit_shift_amount.eq(correct_shift - 1)
+                            m.d.av_comb += final_sig.eq(shifted_sig | (shift_in_bit << bit_shift_amount))
+                        with m.Else():
+                            m.d.av_comb += final_sig.eq(shifted_sig)
+                        m.d.av_comb += final_exp.eq(0)
+
```
tilk, line 207: Confusing empty line.

tilk (approved):
Left minor comments.

coreblocks/func_blocks/fu/fpu/close_path.py
```diff
@@ -0,0 +1,214 @@
...
+                m.d.av_comb += final_exp.eq(0)
+                m.d.av_comb += final_round.eq(guard_bit)
+
+            with m.Elif(exp <= correct_shift):
+                with m.If(exp == 1):
+                    m.d.av_comb += final_sig.eq(final_result)
+                    m.d.av_comb += final_round.eq(guard_bit)
+                with m.Else():
+                    m.d.av_comb += shift_amount.eq((exp - 1))
+                    m.d.av_comb += bit_shift_amount.eq((exp - 2))
```
piotro888, line 175:
```suggestion
                    m.d.av_comb += shift_amount.eq(exp - 1)
                    m.d.av_comb += bit_shift_amount.eq(exp - 2)
```

coreblocks/func_blocks/fu/fpu/close_path.py
```diff
@@ -0,0 +1,214 @@
...
+                m.d.av_comb += is_zero.eq(lza_is_zero_zero)
+
+            with m.If(is_zero):
+                m.d.av_comb += final_sig.eq(final_result)
+                m.d.av_comb += final_exp.eq(0)
+                m.d.av_comb += final_round.eq(guard_bit)
+            with m.Elif(exp == 0):
+                m.d.av_comb += final_sig.eq(final_result)
+                m.d.av_comb += final_exp.eq(0)
+                m.d.av_comb += final_round.eq(guard_bit)
```
piotro888, line 167:
```suggestion
            with m.If(is_zero | (exp == 0)):
                m.d.av_comb += final_sig.eq(final_result)
                m.d.av_comb += final_exp.eq(0)
                m.d.av_comb += final_round.eq(guard_bit)
```
body duplicated

coreblocks/func_blocks/fu/fpu/close_path.py
```diff
@@ -0,0 +1,214 @@
...
+                    m.d.av_comb += l_flag.eq(rtna_l)
+
+            with m.If(l_flag):
+                m.d.av_comb += final_result.eq(result_add_one)
+                m.d.av_comb += correct_shift.eq(shift_amount_lza_one)
+                m.d.av_comb += is_zero.eq(lza_is_zero_one)
+            with m.Else():
+                m.d.av_comb += final_result.eq(result_add_zero)
+                m.d.av_comb += correct_shift.eq(shift_amount_lza_zero)
+                m.d.av_comb += is_zero.eq(lza_is_zero_zero)
```
piotro888, line 158: There is a lot of unnecessary signals in the code, please try to avoid that.
Maybe it can be done directly with one transaction here? Something like:
```python
with Transaction().body(m):
    m.d.av_comb += final_result.eq(Mux(l_flag, result_add_one, ...)
    lza_resp = Mux(l_flag, zero_lza..., one_lza...)
    m.d.av_comb += lza_resp.eq(lza_resp...)

File: coreblocks/func_blocks/fu/fpu/close_path.py [hunk 9]
```diff
@@ -0,0 +1,214 @@
...
+                with m.Case(RoundingModes.ROUND_UP):
+                    m.d.av_comb += l_flag.eq(up_l)
+                with m.Case(RoundingModes.ROUND_DOWN):
+                    m.d.av_comb += l_flag.eq(down_l)
+                with m.Case(RoundingModes.ROUND_ZERO):
+                    m.d.av_comb += l_flag.eq(zero_l)
+                with m.Case(RoundingModes.ROUND_NEAREST_EVEN):
+                    m.d.av_comb += l_flag.eq(rtne_l)
+                with m.Case(RoundingModes.ROUND_NEAREST_AWAY):
+                    m.d.av_comb += l_flag.eq(rtna_l)
```
piotro888, line 149: Similar here, code could be shortened by defining those conditions in `Case` assignments directly.

coreblocks/func_blocks/fu/fpu/close_path.py
```diff
@@ -0,0 +1,214 @@
...
+                    m.d.av_comb += final_sig.eq(final_result)
+                    m.d.av_comb += final_round.eq(guard_bit)
+                with m.Else():
+                    m.d.av_comb += shift_amount.eq((exp - 1))
+                    m.d.av_comb += bit_shift_amount.eq((exp - 2))
+                    m.d.av_comb += final_sig.eq((final_result << shift_amount) | (shift_in_bit << bit_shift_amount))
+                    m.d.av_comb += final_round.eq(0)
+                m.d.av_comb += final_exp.eq(0)
+
+            with m.Else():
```
piotro888, line 180: Signals can be created locally too - like `shifted_sig` could be declared under Else():, instead of at top of elaborate to increase code clarity.

coreblocks/func_blocks/fu/fpu/close_path.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+class ClosePathModule(Elaboratable):
+    """Close path module
+    Based on http://i.stanford.edu/pub/cstr/reports/csl/tr/90/442/CSL-TR-90-442.pdf.
+    This module computes results for effectiv subtraction,
+    whenever difference of exponents is lesser than 2.
+    Beside computing the result this implementation also perform rounding at the same time
+    as subtraction by using two adders (one computing a+b and the other one computing a+b+1).
+    The correct output is chosen based on flags that are different for each rounding mode.
```
piotro888, line 49:
```suggestion
    Based on http://i.stanford.edu/pub/cstr/reports/csl/tr/90/442/CSL-TR-90-442.pdf.
    This module computes results for efficient subtraction,
    whenever difference of exponents is lesser than 2.
    Besides computing the result, this implementation also performs rounding at the same time
    as subtraction by using two adders (one computing a+b and the other one computing a+b+1).
    The correct output is chosen based on flags that are different for each rounding mode.
```
typos

coreblocks/func_blocks/fu/fpu/close_path.py
Durchbruchswagen: Done, shortened all `Signal(1)` to `Singal()`

coreblocks/func_blocks/fu/fpu/close_path.py
Durchbruchswagen: Done, removed empty line.

coreblocks/func_blocks/fu/fpu/close_path.py
Durchbruchswagen: Done, removed empty line.

coreblocks/func_blocks/fu/fpu/close_path.py
Durchbruchswagen: Done, removed empty line.

coreblocks/func_blocks/fu/fpu/close_path.py
Durchbruchswagen: Done, removed empty line.

coreblocks/func_blocks/fu/fpu/close_path.py
Durchbruchswagen: Done

coreblocks/func_blocks/fu/fpu/close_path.py
Durchbruchswagen: Done, removed `{rounding_mode}_l` flags and moved conditions inside `Switch/Case`.

coreblocks/func_blocks/fu/fpu/close_path.py
Durchbruchswagen: Done

coreblocks/func_blocks/fu/fpu/close_path.py
Durchbruchswagen: Ok, so I did it as you suggested, but I'm curious about one thing. The reason why I did it with two transactions is that I was afraid that if I used `Mux` or put it inside `If/Else` statement the module would have to wait for l_flag value to dispatch the correct request. The main advantage of using `LZA` over `LZC` is that we can run it in parallel to the adder, and it takes roughly the same amount of time as addition; thus, we are saving time over using `LZC`. So I'm curious if by using `Mux` the module, it is going to run those two requests and then only assign the correct response, or if it will wait for `l_flag` ?

piotro888: This is not a problem, as transactions/methods always execute in one cycle.
If you are worried about critical path and timing, in this case it is also fine:
The way it works internally, is that when scheduling in Transactron, method calls under Ifs are always considered as called, so this doesn't affect scheduling. After scheduling, Ifs affect `run` signal of methods, and one of them is selected.  However, this doesn't affect critical path here, as the methods called are purely combinational, and nothing depends on if method is actually run.

PR #792, opened 2025-01-23
Push result from FUs

tilk:
This is my *third* attempt at changing the result direction in FUs from pull to push, and I think this is finally it. The change is very incremental - just enough to achieve the goal and nothing more. The change reduces the number of lines of code and even improves performance a little, thanks to removing some FIFOs. This is a good start for tuning the announcement part of the core.

~Depends on #784.~ Merged.

### DIFF
--- a/scripts/synthesize.py
+++ b/scripts/synthesize.py
@@ -111,19 +111,19 @@ def unit_fu(unit_params: FunctionalComponentParams):
     def unit(gen_params: GenParams):
         fu = unit_params.get_module(gen_params)
         issue_adapter = AdapterTrans(fu.issue)
-        accept_adapter = Adapter(fu.push_result)
+        push_result_adapter = Adapter(fu.push_result)
 
         issue_connector, issue_resources = InterfaceConnector.with_resources(issue_adapter, "adapter", 0)
-        accept_connector, accept_resources = InterfaceConnector.with_resources(accept_adapter, "adapter", 1)
+        push_connector, push_resources = InterfaceConnector.with_resources(push_result_adapter, "adapter", 1)
 
-        resources = append_resources(issue_resources, accept_resources)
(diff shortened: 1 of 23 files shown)


scripts/synthesize.py
```diff
@@ -111,7 +111,7 @@ def unit_fu(unit_params: FunctionalComponentParams):
     def unit(gen_params: GenParams):
         fu = unit_params.get_module(gen_params)
         issue_adapter = AdapterTrans(fu.issue)
-        accept_adapter = AdapterTrans(fu.accept)
+        accept_adapter = Adapter(fu.push_result)
```
Hazardu, line 114: Consider renaming `accept_adapter` for consistency with other changes

Hazardu:
It appears that a lot of code in FuncUnit derived classes is duplicated, 
things like assignment of 
```
self.gen_params = gen_params
self.layouts = layouts = gen_params.get(FuncUnitLayouts)
self.issue = Method(i=layouts.issue)
self.push_result = Method(i=layouts.push_result)
```
are the exact same in almost all cases, with one exception being dummyLsu.py that renamed `layouts` to `fu_layouts`

scripts/synthesize.py
tilk: Thanks for noticing that.

tilk:
@Hazardu

> It appears that a lot of code in FuncUnit derived classes is duplicated,

Indeed. This could probably be resolved by adding a base class for FUs, which would include a common constructor. This is not the subject of this PR, however - maybe you would be interested in doing such a refactor?

Hazardu:
Yes, i'll do it. Seems quite simple.

piotro888 (approved):

Hazardu (approved):

PR #794, opened 2025-01-26
Superscalar RS

tilk:
This PR introduces the possibility to perform multiple RS updates in a single clock cycle. Care is taken to avoid priority logic.

Todo:
* [ ] Replace reduction per bit with something else, as this can make the simulation perform poorly.
* [ ] Extend test.

Based on #783.


coreblocks/func_blocks/fu/common/rs.py
```diff
@@ -94,16 +99,28 @@ def _() -> RecordDict:
...
+            for i, record in enumerate(iter(self.data)):
+                m.d.comb += matches_s1[i][k].eq(record.rs_data.rp_s1 == reg_id)
+                m.d.comb += matches_s2[i][k].eq(record.rs_data.rp_s2 == reg_id)
+
+        # It is assumed that two simultaneous update calls never update the same physical register.
+        for i, record in enumerate(iter(self.data)):
+            with m.If(matches_s1[i].any()):
+                m.d.sync += record.rs_data.rp_s1.eq(0)
+                m.d.sync += record.rs_data.s1_val.eq(
+                    reduce(or_, (Mux(matches_s1[i][k], self.update[k].data_in.reg_val, 0) for k in range(self.rs_ways)))
```
lekcyjna123, line 116: Maybe philosophical question, but wouldn't be a 1 multiplexer with `k` entries, better that `k` multiplexers connected with `or`?

lekcyjna123 (approved):

coreblocks/func_blocks/fu/common/rs.py
```diff
@@ -94,16 +99,28 @@ def _() -> RecordDict:
...
+        matches_s1 = Signal(ArrayLayout(len(self.update), self.rs_entries))
+        matches_s2 = Signal(ArrayLayout(len(self.update), self.rs_entries))
+
+        @def_methods(m, self.update)
+        def _(k: int, reg_id: Value, reg_val: Value) -> None:
+            for i, record in enumerate(iter(self.data)):
+                m.d.comb += matches_s1[i][k].eq(record.rs_data.rp_s1 == reg_id)
+                m.d.comb += matches_s2[i][k].eq(record.rs_data.rp_s2 == reg_id)
+
+        # It is assumed that two simultaneous update calls never update the same physical register.
```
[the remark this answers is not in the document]
piotro888: and I think that constraints about interfaces like that should be mentioned in actual rendered docstrings

coreblocks/func_blocks/fu/common/rs.py
piotro888: If this is the case with reduction performance, we should temporarily put that in transactron one-hot mux and use it here, so if pmux is fixed it could be changed in one place in transactron.

piotro888 (commented):
Left comments on @lekcyjna123 remarks

piotro888 (approved):

PR #796, opened 2025-01-27
Add the stall controller

xThaid:
It made a few changes that will be useful in the future work on the frontend:
- I decoupled the stall logic from the fetch unit
- Func units call the `resume` method instead of defining it

todo:
- [x] comments
- [x] fix the test

~One test in `test_core` is failing because of https://github.com/kuznia-rdzeni/transactron/issues/45~ Issue is fixed in Transactron.


tilk:
#802 should have a fix for https://github.com/kuznia-rdzeni/transactron/issues/45.

tilk (approved):
Looks like a nice refactor.

coreblocks/frontend/stall_controller.py
```diff
@@ -0,0 +1,107 @@
...
+            pcs = [Mux(runs[i], v.pc, 0) for i, v in enumerate(args)]
+
+            return {"pc": reduce(operator.or_, pcs, 0)}
+
+        @def_method(m, self._resume_from_unsafe, nonexclusive=True, combiner=resume_combiner)
+        def _(pc):
+            log.assertion(m, stalled_unsafe)
+            m.d.sync += stalled_unsafe.eq(0)
+
+            with condition(m, nonblocking=True) as branch:
```
piotro888, line 81: I would not set `nonblocking` here. If after some future changes versions `redirect_frontend` may be unready, it will cause resume to be missed.
Otherwise LGTM.

coreblocks/frontend/stall_controller.py
```diff
@@ -0,0 +1,107 @@
...
+    The stall controller is responsible for managing all stall/unstall signals
+    that may be triggered in the core and based on them deciding if and where
+    the frontend should be resumed.
+
+
+    Attributes
+    ----------
+    stall_unsafe : Method
+        Signals that the frontend should be stalled because an unsafe (i.e. causing
+        difficult to handle side effects) was just fetched.
```
lekcyjna123, line 33:
```suggestion
        difficult to handle side effects) instruction was just fetched.
```

coreblocks/frontend/stall_controller.py
```diff
@@ -0,0 +1,107 @@
...
+        stalled_exception = Signal()
+
+        @def_method(m, self.stall_guard, ready=~(stalled_unsafe | stalled_exception), nonexclusive=True)
+        def _():
+            pass
+
+        def resume_combiner(m: Module, args: Sequence[MethodStruct], runs: Value) -> AssignArg:
+            # Make sure that there is at most one caller - there can be only one unsafe instruction
+            log.assertion(m, popcount(runs) <= 1)
+            pcs = [Mux(runs[i], v.pc, 0) for i, v in enumerate(args)]
```
lekcyjna123, line 72: Maybe OneHotMux will be better?

coreblocks/frontend/stall_controller.py
tilk: I think that your remark is not true since https://github.com/kuznia-rdzeni/coreblocks/pull/684.

coreblocks/frontend/stall_controller.py
tilk: Unfortunately, all our `OneHot*` structures are potentially not performing well because of https://github.com/kuznia-rdzeni/transactron/issues/54. `OneHotMux` can potentially be changed to use logic like here, if it turns out to be an issue.

xThaid: there are currently two callers of `resume_from_unsafe` (and probably not too many in the future), so I guess it doesn't matter that much what we do here. Nevertheless, a sum of products of all inputs should be pretty efficient.

coreblocks/frontend/stall_controller.py
xThaid: I assumed that this method will be always ready, which is true in the current and the upcoming implementation. I added an assertion to verify it.

lekcyjna123 (approved):

PR #798, opened 2025-01-28
Improve core graph

Arusekk:
I identified two major core graph culprits that can be fixed on the side of Coreblocks: debug signals and simultaneous transactions. Perhaps a bit hacky, but independent of transactron changes suggested in kuznia-rdzeni/transactron#46.


Arusekk:
Let's leave out simultaneous transactions for the future.

tilk (approved):

qbojj (approved):

PR #800, opened 2025-02-05
Remove conflict between FUs reporting exceptions

tilk:
This PR modifies exception reporting so that each reporting FU gets its private report FIFO. This removes the conflict between exception-raising FUs, allowing them to run concurrently, this also simplifies the conflict graph. The giant connected component containing the LSU, jump-branch unit, privileged unit and other stuff is now gone, replaced by smaller ones.

Question to answer: for correctness, should the FIFOs be cleared after exception handling? Is there a risk a stale report will linger in one of the FIFOs after an exception is handled?
Edit - I believe that in our current sluggish core this is not likely to happen, but nothing seems to forbid that. I've added the clear logic.

TODO:
* [x] Answer the question above.
* [x] Find a simple way to perform the clears.
* [x] Test fix.


piotro888:
> Find a simple way to perform the clears.

Maybe clears could be connected to `ResumeFromExceptionKey`?

Additional buffering before `ExceptionInformationRegister` complicates checkpointing with cleaning exception reports on rollbacks. FIFOs would be needed to replaced later with other buffering structure handling tag-selective clearing (but this is probably ok) (or maybe there will some better solution to this?)

tilk:
Added changes handle method clears, and also fixes an issue which caused FIFOs to be created per call site - which sometimes meant two or three for a single FU.

coreblocks/interface/keys.py
```diff
@@ -53,7 +54,7 @@ class FetchResumeKey(UnifierKey, unifier=Collector):
 
 
 @dataclass(frozen=True)
-class ExceptionReportKey(SimpleKey[Method]):
+class ExceptionReportKey(SimpleKey[Callable[[], Callable[Concatenate[TModule, ...], MethodStruct]]]):
```
lekcyjna123, line 57: Please add docstring to explain what this key does/is used for. In our code keys are near global variables, so they should be properly documented in my opinion.

lekcyjna123 (approved):

coreblocks/interface/keys.py
tilk: Done.

piotro888 (approved):

PR #802, opened 2025-02-10
Update transactron and pyright

tilk:
Turned out to be more complex than expected because of the hashability issues. Depends on https://github.com/kuznia-rdzeni/transactron/pull/58, ~Transactron URL needs to be changed after merging.~ IPC changes come from https://github.com/kuznia-rdzeni/transactron/pull/56.


lekcyjna123 (approved):

tilk:
Merging without second review.

PR #803, opened 2025-02-13
Deduplicated FuncUnit

### LINKED ISSUE #793: Refactor FuncUnit to remove duplicate code
@Hazardu

> It appears that a lot of code in FuncUnit derived classes is duplicated,

Indeed. This could probably be resolved by adding a base class for FUs, which would include a common constructor. This is not the subject of this PR, however - maybe you would be interested in doing such a refactor?

_Originally posted by @tilk in https://github.com/kuznia-rdzeni/coreblocks/issues/792#issuecomment-2613918737_

Hazardu:
The purpose of this PR is to resolve #793 by adding a constructor in FuncUnit

### DIFF
--- a/coreblocks/func_blocks/interface/func_protocols.py
+++ b/coreblocks/func_blocks/interface/func_protocols.py
@@ -11,8 +11,6 @@ class FuncUnit(HasElaborate, Protocol):
     issue: Provided[Method]
     push_result: Required[Method]
 
-    # TODO move this to another class, and
-    # make all but JumpBranchWrapper inherit from it instead
     gen_params: Required[GenParams]
     layouts: Required[FuncUnitLayouts]
 
--- a/coreblocks/func_blocks/interface/func_protocols.py
+++ b/coreblocks/func_blocks/interface/func_protocols.py
@@ -1,7 +1,8 @@
 from typing import Protocol
 from transactron import Method, Provided, Required
 from transactron.utils._typing import HasElaborate
-
+from coreblocks.params import GenParams
+from coreblocks.interface.layouts import FuncUnitLayouts
 
 __all__ = ["FuncUnit", "FuncBlock"]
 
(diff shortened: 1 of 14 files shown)


coreblocks/func_blocks/interface/func_protocols.py
```diff
@@ -10,6 +11,17 @@ class FuncUnit(HasElaborate, Protocol):
     issue: Provided[Method]
     push_result: Required[Method]
 
+    # TODO move this to another class, and
+    # make all but JumpBranchWrapper inherit from it instead
+    gen_params: Required[GenParams]
```
Hazardu, line 16: Jump Branch Wrapper is the only FuncUnit derived class that assigns  issue and push_result differently from others. 
One solution is to add another class that all but JumpBranchWrapper inherit from, 
Or to just have JumpBranchWrapper overwrite the issue and push_result after they get assigned in super().__init__

coreblocks/func_blocks/interface/func_protocols.py
```diff
@@ -11,8 +11,6 @@ class FuncUnit(HasElaborate, Protocol):
     issue: Provided[Method]
     push_result: Required[Method]
 
-    # TODO move this to another class, and
-    # make all but JumpBranchWrapper inherit from it instead
     gen_params: Required[GenParams]
     layouts: Required[FuncUnitLayouts]
```
Hazardu, line 15: Should this be made optional instead of required?

JumpBranchWrapper uses JumpBranchLayouts instead of FuncUnitLayouts

Hazardu: Unless `FuncUnitLayouts` is a base class of `JumpBranchLayouts `

tilk: You seem to have misunderstood the `Provided`/`Required` type annotations. These are supposed to be used with `Method` and data structures containing instances of `Method`. Currently, their only job is to guide adapter creation in `SimpleTestCircuit`.

coreblocks/func_blocks/interface/func_protocols.py
```diff
@@ -10,6 +11,15 @@ class FuncUnit(HasElaborate, Protocol):
     issue: Provided[Method]
     push_result: Required[Method]
 
+    gen_params: Required[GenParams]
+    layouts: Required[FuncUnitLayouts]
+
+    def __init__(self, gen_params: GenParams):
```
tilk, line 17: `FuncUnit` is a `Protocol` class, its job is to aid static typing. It's not a base class, and it shouldn't really have a constructor.

awariac:
hi, would you be interested in further development of this change?

tilk:
Closed due to lack of activity.

PR #804, opened 2025-02-18
Simplify JumpBranchFn

tilk:
The `JumpBranchFn` module unnecessarily repeats the logic already present in the decoder - this can be removed. Intermediate `branch_target` turned out to be unnecessary - `jmp_addr` can be assigned early instead. It's ignored on `auipc`, so there is no need for it being 0 there. The changes might allow the optimizer to reduce the number of adders and multiplexers in the synthesized circuit.

Looks like "LUTs used as carry" decreased slightly.

### DIFF
--- a/coreblocks/func_blocks/fu/jumpbranch.py
+++ b/coreblocks/func_blocks/fu/jumpbranch.py
@@ -74,13 +74,7 @@ class JumpBranch(Elaboratable):
     def elaborate(self, platform):
         m = Module()
 
-        branch_target = Signal(self.gen_params.isa.xlen)
-
-        # Spec: "The 12-bit B-immediate encodes signed offsets in multiples of 2,
-        # and is added to the current pc to give the target address."
-        # Multiplies of 2 are converted to the real offset in the decode stage, so we have 13 bits.
-        m.d.comb += branch_target.eq(self.in_pc + self.in_imm[:13].as_signed())
-
+        m.d.comb += self.jmp_addr.eq(self.in_pc + self.in_imm)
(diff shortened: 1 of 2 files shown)


coreblocks/func_blocks/fu/jumpbranch.py
```diff
@@ -89,39 +83,25 @@ def elaborate(self, platform):
 
         with OneHotSwitch(m, self.fn) as OneHotCase:
             with OneHotCase(JumpBranchFn.Fn.JAL):
-                # Spec: "[...] the J-immediate encodes a signed offset in multiples of 2 bytes.
-                # The offset is sign-extended and added to the pc to form the jump target address."
-                # Multiplies of 2 are converted to the real offset in the decode stage, so we have 21 bits.
-                m.d.comb += self.jmp_addr.eq(self.in_pc + self.in_imm[:21].as_signed())
+                m.d.comb += self.jmp_addr.eq(self.in_pc + self.in_imm)
```
xThaid, line 86: maybe it is worth simplifying it even further to have only one adder? I think it would go something like this (pseudo code):
```
jmp_addr = Mux(self.fn == JumpBranchFn.Fn.JALR, self.in1, self.in_pc) + self.in_imm

next_pc_offset = 4
if Extension.C:
  with m.If(self.in_rvc):
    m.d.comb += next_pc_offset.eq(2)

reg_res = Mux(self.fn == JumpBranchFn.Fn.AUIPC, self.jmp_addr, self.in_pc + next_pc_offset)
```

piotro888 (approved):

kalinf (approved):

PR #811, opened 2025-03-16
Change custom machine cause reg values in isa_consts.py

ArturBieniek4:
Interrupts 16 and 17 are marked as _Reserved_ in the [RISC-V ISA](https://drive.google.com/file/d/17GeetSnT5wW3xNuAHI95-SI1gPGd5sJ_/view). This means they might have a different function assigned in the future. We should change those internal interrupts to instead use interrupt numbers 24 and 25, which are marked as _Designated for custom use_.


tilk:
These are not interrupts, but machine cause register values; but other than that, I believe you are right (Table 14 in the linked spec).

tilk (approved):

piotro888 (approved):

PR #812, opened 2025-03-28
Add PMP CSR registers

ArturBieniek4:
This is the beginning of adding CSR registers for PMP from the RV32 spec.

### DIFF
--- a/coreblocks/priv/csr/csr_instances.py
+++ b/coreblocks/priv/csr/csr_instances.py
@@ -2,7 +2,7 @@ from amaranth import *
 
 from typing import Optional
 from coreblocks.arch import CSRAddress
-from coreblocks.arch.csr_address import MstatusFieldOffsets
+from coreblocks.arch.csr_address import MstatusFieldOffsets, PMPConfigFieldOffsets
 from coreblocks.arch.isa import Extension
 from coreblocks.arch.isa_consts import PrivilegeLevel, XlenEncoding, TrapVectorMode
 from coreblocks.params.genparams import GenParams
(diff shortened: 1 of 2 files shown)


piotro888 (changes requested):
My comment on the meeting was about using `AliasedCSR` for individual bits and I didn't notice that `pmpcfgX` registers are super-registers of `pmpXcfg` ones; sorry for the confusion.
Using separate registers for each `pmpXcfg` would be helpful here and is the intended use case of  `AliasedCSR`.

Please change:
* generate separate `pmpXcfg`, that will be added to externally accessible (self.) array
* prepare code for RV64 like rest of `csr_instances`. Note that odd pmpcfgX are skipped
* restore field offsets, as this will be is useful

EDIT: magic numbers like 16, 64 should not be used directly, but be a constant defined in CSR or ISA constants

coreblocks/priv/csr/csr_instances.py
```diff
@@ -86,6 +86,28 @@ def __init__(self, gen_params: GenParams):
...
+                setattr(self, f"pmpcfg{i}", pmpcfg_i)
+
+        elif gen_params.isa.xlen == 32:
+            for i in range(CSRAddress.PMPCFG_COUNT):
+                pmpcfg_i = AliasedCSR(getattr(CSRAddress, f"PMPCFG{i}"), gen_params)
+                for j in range(4):
+                    pmp_j_cfg = CSRRegister(None, gen_params, width=4)
+                    pmpcfg_i.add_field(j * CSRAddress.PMPXCFG_WIDTH, pmp_j_cfg)
+                    setattr(self, f"pmp{i*4+j}cfg", pmp_j_cfg)
+                setattr(self, f"pmpcfg{i}", pmpcfg_i)
```
piotro888, line 106: Please don't duplicate code.
It can be done in one place ex. by computing `for j in range (pmpcfg_i.width // PMPXCFG_WIDTH)` and doing loop continue on % == 1 when xlen == 64.

coreblocks/priv/csr/csr_instances.py
```diff
@@ -86,6 +86,28 @@ def __init__(self, gen_params: GenParams):
...
 
+        if gen_params.isa.xlen == 64:
+            # In RV64, odd-numbered configuration registers pmpcfg1, ... pmpcfg15 are illegal.
+            for i in range(0, CSRAddress.PMPCFG_COUNT, 2):
+                pmpcfg_i = AliasedCSR(getattr(CSRAddress, f"PMPCFG{i}"), gen_params)
+                for j in range(8):
+                    pmp_j_cfg = CSRRegister(None, gen_params, width=8)
+                    pmpcfg_i.add_field(j * CSRAddress.PMPXCFG_WIDTH, pmp_j_cfg)
+                    setattr(self, f"pmp{i*8+j}cfg", pmp_j_cfg)
+                setattr(self, f"pmpcfg{i}", pmpcfg_i)
```
piotro888, line 97: pmpcfgX will never be accessed, this should not be needed

coreblocks/priv/csr/csr_instances.py
```diff
@@ -86,6 +86,28 @@ def __init__(self, gen_params: GenParams):
...
         )
 
+        if gen_params.isa.xlen == 64:
+            # In RV64, odd-numbered configuration registers pmpcfg1, ... pmpcfg15 are illegal.
+            for i in range(0, CSRAddress.PMPCFG_COUNT, 2):
+                pmpcfg_i = AliasedCSR(getattr(CSRAddress, f"PMPCFG{i}"), gen_params)
+                for j in range(8):
+                    pmp_j_cfg = CSRRegister(None, gen_params, width=8)
+                    pmpcfg_i.add_field(j * CSRAddress.PMPXCFG_WIDTH, pmp_j_cfg)
+                    setattr(self, f"pmp{i*8+j}cfg", pmp_j_cfg)
```
piotro888, line 96: Avoid using `setattr`. This can be done cleaner and easier to use by declaring `self.pmpxcfg = []` array before and appending CSRs to it.

coreblocks/priv/csr/csr_instances.py
ArturBieniek4: Each pmpcfgX needs to be an attribute of self, otherwise it fails with "method not defined".

coreblocks/priv/csr/csr_instances.py
ArturBieniek4: This unfortunately does not work, fails with "Method '_fu_read' not defined".

coreblocks/priv/csr/csr_instances.py
ArturBieniek4: This is done in the latest commit.

piotro888 (approved):
Looks good, please add extra class-global `pmpxcfg` array apart from `setattr` (with references to created CSRs), to make those CSRs easier to access externally.

tilk (approved):

PR #816, opened 2025-05-15
Physical memory protection implementation

ArturBieniek4:
This is the first PMP implementation sketch.

PR #840, opened 2025-09-22
Configurable number of PMP entries

### LINKED ISSUE #817: Make PMP optional
PMP CSR storage takes a lot of space LUT-wise, we might want to make it disable-able it in core configurations.

piotro888:
Resolves #817

Adds configuration to PMP entries to `CoreConfiguration`, and slightly refactors `pmpcfg` registers generation (+buggy width on rv64 fixed).
As tilk mentioned, spec allows it to be set to 64, 16 or disabled.

PMP registers take quite a lot of LUT space,  so reduced 16 size options looks very useful if PMP is required. 
It caused  `basic` configuration to no longer fit on an arty board :/ (and breaks the linux tutorial ://)

#816 would need to be adjusted to this change

### DIFF
--- a/coreblocks/params/configurations.py
+++ b/coreblocks/params/configurations.py
@@ -108,6 +108,8 @@ class _CoreConfigurationDataClass:
         read-only and directly connected to input signal (implementation must provide clearing method)
     user_mode: bool
         Enable User Mode.
+    pmp_register_count: int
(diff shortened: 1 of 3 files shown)


tilk:
Merging for practical reasons mentioned in the description.

PR #843, opened 2025-09-26
Fix LSUAtomicWrapper - Linux regression

### LINKED ISSUE #842: Coreblocks / Linux regression
There is a regression somewhere between previously patched piotro888/piotro/fosdem-2025 https://github.com/piotro888/coreblocks/commit/3b48decb49b28b83c28a3d1cc38178f5dd5e718b (base in master https://github.com/kuznia-rdzeni/coreblocks/commit/320219c1c7d0e45e66af4265dea1891e16be218f) and current master https://github.com/kuznia-rdzeni/coreblocks/commit/b825ce8d0d5a97062e988724921db9efa6a93c8c

Linux boot hangs up just before scheduler intialisation (last messages printed from `mm_core_init();` in https://github.com/torvalds/linux/blob/v6.9/init/main.c#L940), so I suspect interrupt related / flush problems.
 
I ruled out regression in LiteX using original patched LiteX branch with the new core - still not working. The same built kernel image works with the old  (February) gateware. (However error could be still somewhere else I missed)

Current configuration is run on digilent_arty using LiteX https://github.com/piotro888/litex/tree/coreblocks-update-202509 branch, and pythondata-cpu-coreblocks manually updated to master +  #841 applied (in Feb build attributes needed to be manually inserted in core.v)

Any help welcome, I will keep this issue updated.

piotro888:
Fixes #842

Adding `Forwarder` to `push_result` of `LSUAtomicWrapper` fixes the boot.
This `Forwarder` was removed in #792, when refactoring to the push result FU architecture.

However, the problem is, I'm not sure why it fixes it. 
From what I see, both `lsu_push_result` and `sc_failed_trans` depend on  `push_result` and the other AMO issue Transaction is strictly ordered by registers from `lsu_push_result`, so the `Forwarder` change should not change the behaviour.
Both `sc_failed_trans` and `lsu_push_result` would be blocked for a moment without the `Forwarder` and state should not be changed.
@tilk can you take a look at this? second pair of eyes could be helpful here. 
There is always a risk it doesn't fix the core issue but just affect the instruction-cycle timing.

Adding `Forwarder` is the only relevant change here that fixes the boot itself, other is renaming for clarity and extra condition (that shouldn't be necessary, but its better to be safe).

### DIFF
--- a/coreblocks/func_blocks/fu/lsu/lsu_atomic_wrapper.py
+++ b/coreblocks/func_blocks/fu/lsu/lsu_atomic_wrapper.py
@@ -130,7 +130,9 @@ class LSUAtomicWrapper(FuncUnit, Elaboratable):
 
         lsu_push_result = Method(i=self.fu_layouts.push_result)
 
-        m.submodules.result_buff = result = Forwarder(self.fu_layouts.push_result)
+        # Forwarder here is required to workaround Transactron simultaneous transactions issue, when using
+        # conditional method calls, see https://github.com/kuznia-rdzeni/coreblocks/pull/843#issuecomment-3350907439
+        m.submodules.result_fwd = result = Forwarder(self.fu_layouts.push_result)
 
         m.submodules.result_connect = ConnectTrans.create(result.read, self.push_result)
 
@@ -191,7 +193,7 @@ class LSUAtomicWrapper(FuncUnit, Elaboratable):
                 },
             )
 
-        sc_failed_trans.add_conflict(lsu_push_result, priority=Priority.RIGHT)
+        sc_failed_trans.add_conflict(lsu_push_result, priority=Priority.RIGHT)  # only sets the priority for performance
 
         with Transaction().body(m, ready=atomic_in_progress & amo_second_op_issue & ~amo_second_op_issue_finished):
             amo_store = Signal(self.fu_layouts.issue)


tilk:
A short reproduction example would be great, but I imagine you don't have one. I need to think about this a little more, but I noticed two small things.

* The conflict between `sc_failed_trans` and `lsu_push_result`, with a set priority. If the priority is to enforce some ordering, this might not work in all cases because of other dependencies in the system. I now believe that conflicts should be used for correctness (e.g. when sharing a resource), and priorities should only be used for performance reasons. Either way, because the LSU is not pipelined (yet), I don't see why this could cause an issue. One thing to test might be removing `sc_failed_trans` and sending the result early.
* Adding the forwarder might partially decouple `lsu_push_result` and `sc_failed_trans` from the announcement stage, which can make things happen in the core in a different order. Still, I currently don't see why this should impact correctness.

piotro888:
Here are the related waveforms https://nc.kuznia-rdzeni.org/s/eeckoctyXPcxmKa
BTW, https://gitlab.com/surfer-project/surfer is awesome! ;)

piotro888:
> If the priority is to enforce some ordering, this might not work in all cases because of other dependencies in the system. I now believe that conflicts should be used for correctness (e.g. when sharing a resource), and priorities should only be used for performance reasons.

The `add_conflict` is there only to add priority for performance - to push remaining result with data first before sc_failed (that is younger in order, and probably less important). This priority setting doesn't affect the correctness.
Conflict is already there because of shared use of `push_result`.

tilk:
> I think this is an Transactron issue.

Damn. If so, it's pretty important to get to the root of the problem, we might stumble upon this again.

> Here are the related waveforms https://nc.kuznia-rdzeni.org/s/eeckoctyXPcxmKa

That's a pretty giant GTKW file, will probably have to look at it at work later.

tilk:
For now, I'm confirming your observation - in your VCD file indeed it looks like that a spurious announcement with `rob_id` equal to `0` is sent when `atomic_in_progress` is 1 and `atomic_second_reqest` is 0, and the core locks up when the actual `rob_id` is also 0. Unfortunately, I didn't find the root cause. I think a smaller example is in order, I'd start with an asm test which spams the affected instruction.

By the way, it looks like adding more assertions to the core to check various assumptions makes sense. I believe that verifying in ROB that `mark_done` is called only when `done` is 0 would catch this problem early.

piotro888:
Thanks for looking into it!! I also suspected the simultaneous in `Connect`.
I will add a note on why Forwarder was added here, and check in Transactron would be great.
Do you have some some ideas in mind about the transaction manager redesign? Would that be a large change?

tilk:
> Do you have some some ideas in mind about the transaction manager redesign? Would that be a large change?

It might turn out to be a large change, I don't know yet. The possible semantics in various use scenarios and their consequences need to be determined. That's why I'm for postponing this - I'm practically certain that a "quick and dirty" solution will be wrong in some way.

PR #846, opened 2025-10-20
Checkpointing part 2: final integration

awariac:
Checkpointing is **finally** here!!! :tada: (it took a bit of work and debug)

Integrates rollback-on-branch-misprediction flow and use of instruction tags into the core.
Async interrupts and other exceptions are still handled by flushing the old way.

TODO:
fix other unit tests, verify linux boot, verify benchmarks


piotro888:
yoooo

tilk:
> yeah, it is, but I expected it to be bigger (hope that's not a wishful thinking). Did you have any expectations of it?

Also hoped for a bigger difference, but it looks like there are other factors at play. A few of the most obvious for me are:

* Mispredictions still have a cost (filling the pipeline, keeping the functional units busy with not needed operations).
* The core cannot execute dependent instructions cycle after cycle as there is no forwarding from instruction results to RS (and full forwarding would probably be costly to implement). Superscalarity would help discover independent instructions faster.
* According to metrics, quite a lot of instructions seem to linger in RS for some time. There could be multiple reasons for that: maybe the results are not coming in fast enough, or maybe the other end of the pipeline is the limiting factor. I would consider adding more metrics to RS to help understand the behavior.
* Lack of superscalarity means that if multiple instructions in different RS have their operand ready in the same cycle, they still need to complete in sequence.
* The LSU blocks reordering of instructions.

It's hard to decide the importance of these factors on the final performance. There is probably no single biggest reason, I'm starting to believe that OoO performance is complex and comes from a combination of many factors.

tilk:
Given that benchmarks and other full core tests work fine, most probably the issue is related to M-mode and interrupt handling. Indeed RISCV-DV claims to cover this.

piotro888 (commented):
oops wrong PR

awariac:
finally made a merge with superscalarity, new frontend and other changes.
stopped working correctly on core and checkpointing tests.
needs further investigation

awariac:
@xThaid can you tell me how FTQ `commit` and `backend_redirect` should behave on rollbacks (that invalidate some suffix of instructions) and retiring invalidated instructions (that were rolled-back and have tag set to inactive)?
I want to make sure I understand it correctly.

If rollback would call `backend_redirect`, then FTQ head pointer would be set to commit+1, but valid instructions that were in between and are still active, have IDs out of range now. It always flushes the frontend, but I'm not sure how to do the tag cleanup correctly. Is some special handling needed or I misunderstood it?

edit: we know the ftq pointer of branch that triggers rollback. Should it be set to that ptr + 1 then?

xThaid:
> @xThaid can you tell me how FTQ `commit` and `backend_redirect` should behave on rollbacks (that invalidate some suffix of instructions) and retiring invalidated instructions (that were rolled-back and have tag set to inactive)? I want to make sure I understand it correctly.
> 
> If rollback would call `backend_redirect`, then FTQ head pointer would be set to commit+1, but valid instructions that were in between and are still active, have IDs out of range now. It always flushes the frontend, but I'm not sure how to do the tag cleanup correctly. Is some special handling needed or I misunderstood it?
> 
> edit: we know the ftq pointer of branch that triggers rollback. Should it be set to that ptr + 1 then?

I think the behavior of commit shouldn't change (I might be missing something). Presumably rollbacks don't affect what we commit?

For `backend_redirect`, yeah, I agree that in the current shape we can't express that we want to flush a suffix of active instructions. I see that you passed FTQ pointers to RSes so we know how to redirect the FTQ when resuming from an unsafe instruction. I actually need it for branch prediction as well. I quickly put up #1004 - I believe that you will find it useful if I understand checkpointing correctly.

awariac:
@xThaid i think i already did that yesterday here https://github.com/kuznia-rdzeni/coreblocks/pull/846/changes/bab1cf0dfd4c2a3df2f354ae62977d73ba1ee270 + minor change to retirement commit earlier.
Could you look at it?
Your change is probably better polished, but you could change for equivalence. This PR needs to be split in smaller changes anyway i think.

awariac:
meh github doesnt trigger ci on conflicts

awariac:
WOW +0.3

and its green again ^^'

tilk:
Closing PR, superseded by #1050.

PR #858, opened 2026-03-03
On-the-fly conversion and quotient selection function

Durchbruchswagen:
Implementation of: 
- QSF (quotien selection function) that selects next quotient digit based on few MSB bits of divisor and partial remainder.
- On-the-fly conversion that performs conversion between redundant representation of quotient to non-redundant representation, by using two registers, one assuming that next digit is non-negative and the other one assuming next digit is positive, to avoid carry propagation


tilk:
You have circular imports.

coreblocks/func_blocks/fu/fpu/otfc.py
```diff
@@ -0,0 +1,123 @@
...
+        def _(arg):
+            m.d.sync += a_register.eq(0)
+            m.d.sync += b_register.eq(3)
+
+        @def_method(m, self.otfc_add_digit)
+        def _(sign, q):
+            a1 = q[1] | sign
+            a0 = q[0]
+            b1 = ((~q[1]) & (~q[0])) | (sign & q[0])
+            b0 = ~q[0]
```
tilk, line 111: These should be signals as they are used multiple times. Actually, given the usage, you could just declare two-bit signals `a` and `b`.

test/func_blocks/fu/fpu/test_otfc.py
```diff
@@ -0,0 +1,51 @@
...
+                [(0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0)],
+            ]
+            results = [
+                1002,
+                23,
+                1557 + (3 << 12),
+                4093 + (3 << 12),
+                (1 << 11) + (1 << 2),
+                0,
+            ]
```
tilk, line 34: There are just a few cases here, so this doesn't hurt that much, but writing cases like this hurts readability and maintainability. Group test case data, e.g. like you did in `test_fpu_classification`.

coreblocks/func_blocks/fu/fpu/fpu_qsf.py
```diff
@@ -0,0 +1,122 @@
...
+                    with m.If(residual < Const(self.bounds[i][0])):
+                        m.d.av_comb += q.eq(Const(self.digits[0][1]))
+                        m.d.av_comb += sign.eq(Const(self.digits[0][0]))
+                    for j in range(1, len(self.bounds[i])):
+                        with m.Elif(residual < Const(self.bounds[i][j])):
+                            m.d.av_comb += q.eq(Const(self.digits[j][1]))
+                            m.d.av_comb += sign.eq(Const(self.digits[j][0]))
+                    with m.Else():
+                        m.d.av_comb += q.eq(Const(self.digits[len(self.bounds[i])][1]))
+                        m.d.av_comb += sign.eq(Const(self.digits[len(self.bounds[i])][0]))
```
tilk: Are the `Const` uses here really needed? Looks weird.

tilk (changes requested):

coreblocks/func_blocks/fu/fpu/otfc.py
Durchbruchswagen: Done, added two signals `a` and `b` and used variables `a0/1`, `b0/1` to set value of those signals

test/func_blocks/fu/fpu/test_otfc.py
Durchbruchswagen: Done, grouped test cases and results in a single dataclass with two fields

coreblocks/func_blocks/fu/fpu/fpu_qsf.py
Durchbruchswagen: No, they are not needed, but considering table values are constant, I thought they should be put in `Const` to indicate that, but if this looks weird we may get rid of those `Consts`. 
For now I removed `Consts` out of `fpu_qsf.py`.

tilk (approved):

PR #869, opened 2026-03-13
Implement initial Fetch Target Queue (FTQ)

xThaid:
The Fetch Target Queue (FTQ) is a buffer that decouples branch prediction from instruction fetch. The branch predictor writes predicted fetch addresses (along with some prediction metadata) into the FTQ ahead of time, while the fetch unit consumes entries from the queue to retrieve the corresponding instruction blocks. FTQ stores prediction information so it can be later sent back to the BPU for predictor training after instruction commit. FTQ maintains the complete lifecycle of instructions from prediction to commit.

In this PR I'm adding an initial implementation of FTQ. It doesn't do anything meaningful at this point: it simply stores basic predictions from the branch prediction unit and then sends them to the fetch unit.

TODO in this PR:
- [x] some FTQ unit tests
- [x] documentation

### DIFF
--- a/coreblocks/frontend/bpu/bpu.py
+++ b/coreblocks/frontend/bpu/bpu.py
@@ -1,0 +1,50 @@
+from amaranth import *
+
+from transactron.core import *
+from transactron.utils.transactron_helpers import make_layout
+from transactron.lib import logging
+from transactron.lib.connectors import Pipe
+
+from coreblocks.params import *
+from coreblocks.frontend import FrontendParams
+from coreblocks.interface.layouts import CommonLayoutFields
+from coreblocks.interface.layouts import BranchPredictionLayouts
+
+log = logging.HardwareLogger("frontend.bpu")
+
+
+class BranchPredictionUnit(Elaboratable):
+    request: Provided[Method]
+    write_prediction: Required[Method]
+    flush: Provided[Method]
+
+    def __init__(self, gen_params: GenParams) -> None:
+        self.gen_params = gen_params
+        self.layouts = gen_params.get(BranchPredictionLayouts)
+
+        self.request = Method(i=self.layouts.request)
+        self.write_prediction = Method(i=self.layouts.write_prediction)
+        self.flush = Method()
+
+    def elaborate(self, platform):
+        m = TModule()
+
+        fparams = self.gen_params.get(FrontendParams)
+        fields = self.gen_params.get(CommonLayoutFields)
+
+        # TODO: 'clear' method of Pipe creates conflicts with other methods. This blocking behavior is not
+        # needed here and creates a lot of unnecessary logic and delays
+        m.submodules.pipe = pipe = Pipe(layout=make_layout(fields.pc, fields.ftq_ptr))
+
+        @def_method(m, self.request)
+        def _(pc, ftq_ptr):
+            pipe.write(m, pc=fparams.pc_from_fb(fparams.fb_addr(pc) + 1, 0), ftq_ptr=ftq_ptr)
+
+        with Transaction(name="BPU_Stage1").body(m):
+            self.write_prediction(m, pipe.read(m))
+
+        @def_method(m, self.flush, nonexclusive=True)
+        def _():
+            pipe.clear(m)
+
+        return m
--- a/coreblocks/frontend/bpu/bpu.py
+++ b/coreblocks/frontend/bpu/bpu.py
@@ -1,0 +1,50 @@
+from amaranth import *
+
+from transactron.core import *
+from transactron.utils.transactron_helpers import make_layout
+from transactron.utils import logging
+from transactron.lib.connectors import Pipe
+
+from coreblocks.params import *
+from coreblocks.frontend import FrontendParams
+from coreblocks.interface.layouts import CommonLayoutFields
+from coreblocks.interface.layouts import BranchPredictionLayouts
+
+log = logging.HardwareLogger("frontend.bpu")
+
+
+class BranchPredictionUnit(Elaboratable):
+    request: Provided[Method]
+    write_prediction: Required[Method]
+    flush: Provided[Method]
+
+    def __init__(self, gen_params: GenParams) -> None:
+        self.gen_params = gen_params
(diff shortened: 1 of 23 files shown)


coreblocks/interface/keys.py
```diff
@@ -132,6 +132,9 @@ class SFenceVMAKey(UnifierKey, unifier=MethodProduct.create):
 class L1TLBBackingDeviceKey(SimpleKey["TLBBackingDevice"]):
     """Used to provide a component that can be used as a backing device for the L1 TLB."""
 
+
+@dataclass(frozen=True)
+class FTQCommitEntry(SimpleKey[Method]):
```
qbojj: It is not present in `__all__`

coreblocks/interface/layouts.py
```diff
@@ -26,9 +27,21 @@
...
 
 
+class FTQPtrLayout(CircularBufferPointer.Layout):
+    def __init__(self, gp: GenParams):
+        super().__init__(size_log=gp.ftq_size_log)
+
+
+class FTQPtr(CircularBufferPointer):
+    def __init__(self, target=None, *, gp: GenParams, **kwargs):
+        super().__init__(layout=FTQPtrLayout(gp), target=target, **kwargs)
```
qbojj: All other places call the argument `gen_params`

tilk:
Looks like there is some issue with compressed instructions?

Because it's now exam season, my review will probably be delayed.

qbojj (changes requested):

qbojj:
I will continue the review when I find some time. I think Fmax should also be addressed.

xThaid:
re fmax: conflicts created by the `clear` method of `Pipe` module seems to be the problem

coreblocks/frontend/bpu/bpu.py
```diff
@@ -0,0 +1,82 @@
...
+
+from coreblocks.params import *
+from coreblocks.frontend import FrontendParams
+from coreblocks.interface.layouts import CommonLayoutFields
+from coreblocks.interface.layouts import BranchPredictionLayouts
+
+log = logging.HardwareLogger("frontend.bpu")
+
+
+class _Pipe(Elaboratable):
```
tilk: Note: remove after Transactron 0.9.

coreblocks/frontend/ftq.py
```diff
@@ -0,0 +1,300 @@
...
+    BranchPredictionLayouts,
+    FTQPtr,
+    FetchTargetQueueLayouts,
+)
+from coreblocks.interface.keys import PredictedJumpTargetKey, BranchResolveKey, FTQCommitEntry
+
+log = logging.HardwareLogger("frontend.ftq")
+
+
+class FetchAddressUnit(Elaboratable):
```
tilk: Maybe move this to a separate file? This is already moved (from `frontend.py`) which complicates the diff.

coreblocks/interface/views.py
```diff
@@ -0,0 +1,57 @@
+from amaranth.lib import data
+from typing import Optional
+from amaranth_types import ValueLike
+from amaranth import *
+
+
+class CircularBufferPointer(data.View):
```
tilk, line 7: This looks useful, it could serve as a base for alternative FIFO implementation. I wonder what advantages/disadvantages has this over how BasicFifo is implemented. If this is useful enough, it should go to Transactron.

Docstrings needed,

coreblocks/frontend/ftq.py
```diff
@@ -0,0 +1,300 @@
...
+
+class FetchTargetQueue(Elaboratable):
+    """
+    Tracks speculative fetch targets in a circular buffer indexed by FTQ pointer.
+    Each entry represents one fetch block: the FTQ allocates an entry per fetch, sends the PC to
+    both the IFU and BPU, and advances the fetch pointer. Entries are freed at retirement via
+    ``commit`` and the buffer stalls allocation when full.
+    """
+
+    stall_lock: Required[Method]
```
tilk: Huh, why the name change? It's called `stall_guard` in `stall_controller`, which feels like a better name. This method does not lock anything, it does nothing but stops things from execution sometimes. (Also, `guard` looks like a good general naming convention for things like this, I feel that e.g. `precommit` should be renamed,)

tilk:
Fmax is much better, but looks like the critical path still goes through the prediction checker.
Edit: looks like the critical path goes through old code, in particular handling of instructions crossing fetch blocks. Fix not needed here.

coreblocks/frontend/ftq.py
xThaid: I initially thought that `lock` would be better for some reason, but now I totally buy your point - I renamed it.

coreblocks/interface/views.py
xThaid: There's probably not much benefit to using this to implement a basic FIFO. It's more useful when we have a circular buffer and want to operate on its pointers (arithmetic, comparison, etc.). The main benefit of this over a simple power-of-two pointer is that there's no ambiguity around the wrap point (thanks to the parity bit).

I think the ROB (and soon the FTQ) are great use-cases for this. For example, currently, we have to resort to shenanigans just to compare ROB pointers:
```
rob_start_idx = self.rob_get_indices(m).start
m.d.comb += should_write.eq(
    (rob_id - rob_start_idx).as_unsigned() < (self.rob_id - rob_start_idx).as_unsigned()
)
```

if this we were using CircularBufferPointer, it would be more like this:
```
m.d.comb += should_write.eq(rob_id < self.rob_id)
```
and we don't have to access the global "start" pointer. The tradeoff is obviously the extra parity bit carried by every pointer.

coreblocks/interface/keys.py
```diff
@@ -132,6 +133,9 @@ class SFenceVMAKey(UnifierKey, unifier=MethodProduct.create):
...
```
tilk: All keys have `Key` at the end of their name. Let's not break established conventions for no reason.

tilk (approved):

qbojj (approved):

PR #870, opened 2026-03-16
Implement Physical Memory Protection

dawidpawliczek4:


tilk (changes requested):

piotro888:
Also it could be verified (in favorite gui synthesis view) if generated 
Verilog RTL and later GL looks sane for internal PMP checking.

qbojj:
Address is now computed a cycle before it is used in PMP - before adder was directly feeding into PMP's comparators. At least that is what changed since the last benchmark

tilk (approved):

qbojj (approved):

PR #873, opened 2026-03-23
Superscalar register allocation

tilk:
This PR implements superscalar register allocation. Groups of instructions are now passed from the frontend to the scheduler, which are then serialized after register allocation. In-flight instruction counter is modified to allow changing its value by more than 1 each cycle.

Originally I wanted to superscalarize the entire scheduler in one PR, but I think I'll do it piecewise instead.

TODO:
- [x] Superscalar reg allocation
- [x] Fix tests

### DIFF
--- a/coreblocks/priv/traps/instr_counter.py
+++ b/coreblocks/priv/traps/instr_counter.py
@@ -1,46 +1,51 @@
 from amaranth import *
 from coreblocks.params.genparams import GenParams
 from coreblocks.interface.layouts import CoreInstructionCounterLayouts
-from transactron.core import Method, TModule, def_method
+from transactron.core import Method, Provided, TModule, def_method
 
 
 class CoreInstructionCounter(Elaboratable):
     """
     Counts instructions currently processed in core.
     Used in exception handling, to wait for core flush to finish.
+    """
 
-    Attributes
-    ----------
-    increment : Method
-        Increments the counter. Should be called when new instruction leaves fetch stage.
-    decrement : Method
-        Decrements the counter, and returns if the counter will be equal to zero after that cycle (it was the
-        last instruction in core and no new instruction is fetched). Should be called when instruction is retired.
+    increment: Provided[Method]
+    """Increments the counter. Should be called when new instruction leaves fetch stage."""
+
+    decrement: Provided[Method]
+    """Decrements the counter. Should be called when instruction is retired.
+    Returns a boolean saying if the counter will be equal to zero after that cycle (it was the
+    last instruction in core and no new instruction is fetched).
     """
 
     def __init__(self, gen_params: GenParams):
         self.gen_params = gen_params
 
-        self.increment = Method()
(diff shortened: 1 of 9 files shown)


coreblocks/scheduler/scheduler.py
```diff
@@ -24,7 +24,7 @@ class RegAllocation(Elaboratable):
 
     get_instr: Required[Method]
     push_instr: Required[Method]
```
piotro888: `*_instrs`?

coreblocks/scheduler/scheduler.py
```diff
@@ -38,22 +38,21 @@ def __init__(self, *, gen_params: GenParams):
...
-            instr = self.get_instr(m)
-            with m.If(instr.regs_l.rl_dst != 0):
-                reg_id = self.get_free_reg(m)
-                m.d.comb += free_reg.eq(reg_id)
+            instrs = self.get_instr(m)
+            m.d.av_comb += assign(data_out, instrs)
+
+            for i in range(self.gen_params.frontend_superscalarity):
+                with m.If((i < instrs.count) & (instrs.data[i].regs_l.rl_dst != 0)):
+                    m.d.av_comb += data_out.data[i].regs_p.rp_dst.eq(self.get_free_reg[i](m).ident)
```
piotro888: Should we care about a case when number of free regs < super-scalar instructions available?
That's some performance/complexity tradeoff that needs to be carefully reviewed (or a paramterization option).

coreblocks/priv/traps/instr_counter.py
```diff
@@ -1,46 +1,51 @@
...
-        self.decrement = Method(o=gen_params.get(CoreInstructionCounterLayouts).decrement)
+        layouts = gen_params.get(CoreInstructionCounterLayouts)
+        self.increment = Method(i=layouts.increment_in)
+        self.decrement = Method(i=layouts.decrement_in, o=layouts.decrement)
 
     def elaborate(self, platform) -> TModule:
         m = TModule()
 
-        count = Signal(self.gen_params.rob_entries_bits + 1)
+        counter = Signal(self.gen_params.rob_entries_bits + 1)
```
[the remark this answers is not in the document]
tilk: Not the subject of this PR.

coreblocks/scheduler/scheduler.py
tilk: * Currently we use 64 physical registers. In 2-way superscalarity, at most one of them will end up unused - 1/64 doesn't feel like much.
* Currently we don't have much register contention. Metrics show maximum 54 registers allocated in all benchmarks, with the average of around 33. Obviously this will change with superscalarity, but it doesn't seem alarming (yet).
* Currently we use only 14 DP16KD block RAM elements, out of 208. And the register file is not that big - 64 registers on RV32 is just 2048 bits, while a single DP16KD is 18 Kbit in size. I don't think enlarging the register file is going to be much of a problem - it's just a drop compared to the needs of L1 caches, for example.

All in all, I really don't think this is a problem at all.

PR #877, opened 2026-03-30
use more modular Zc* extensions instead of C and implement Zcb

qbojj:
Implements the Zc* modular extensions in place of monolithic C extension.

Additionally implements Zcb for more compressed instructions.


coreblocks/arch/isa.py
```diff
@@ -81,6 +81,14 @@ class Extension(enum.IntFlag):
     ZHINX = auto()
     #: Extended shift operations
     ZBA = auto()
+    #: Compressed integer instructions (16-bit, formerly part of C)
+    ZCA = auto()
+    #: Compressed single-precision floating-point instructions (formerly part of C if F enabled)
+    ZCF = auto()
+    #: Compressed floating-point double instructions (formerly part of C if D enabled)
+    ZCD = auto()
+    #: Compressed simple code-size reduction instructions
```
xThaid: not super important, but the comment seems to be a bit vague. How about something like
```
Compressed byte/half-word loads, stores, and bit manipulation
```

piotro888 (commented):
Please check if changes are not needed for ISA string generation (in gen_params) from selected FU configurations.
Related code:
- https://github.com/kuznia-rdzeni/coreblocks/blob/cd4276bfb4f90cf4a1a4cbba632d37c5b07c16b8/coreblocks/params/fu_params.py#L77
- https://github.com/kuznia-rdzeni/coreblocks/blob/cd4276bfb4f90cf4a1a4cbba632d37c5b07c16b8/coreblocks/params/genparams.py#L25
- https://github.com/kuznia-rdzeni/coreblocks/blob/master/test/params/test_configurations.py

tilk (commented):
Looks much cleaner now.

tilk (approved):

piotro888 (approved):

PR #878, opened 2026-03-30
implement S mode CSRs

### LINKED ISSUE #528: Add CSR aliasing support
Extend CSR component with a wrapper which will provide a possibility to connect 2 or more small CSRs into a one bigger CSR. The goal of these changes is to support such cases as `fflags` and `frm` which create `fcsr` and `cycle` CSR which should be alias to `time`. If possible your wrapper should have an interface compatible with `CSRRegister` class.

qbojj:
closes #863


coreblocks/arch/csr_address.py
piotro888: undefined?

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -71,15 +73,49 @@ def __init__(self, gen_params: GenParams):
...
             | (((1 << gen_params.interrupt_custom_count) - 1) << 16)
         )
+        if gen_params.supervisor_mode:
+            mie_writeable |= smode_interrupts
+
         self.mie = CSRRegister(CSRAddress.MIE, gen_params, ro_bits=~mie_writeable)
         self.mip = CSRRegister(CSRAddress.MIP, gen_params, fu_write_priority=False, ro_bits=~self.edge_reported_mask)
 
+        if gen_params.supervisor_mode:
+            # TODO: implement logic for delegation and S-mode interrupts
```
piotro888: If it is not supported (for now) fields should be RO 0.

coreblocks/func_blocks/csr/csr.py
```diff
@@ -135,19 +136,44 @@ def elaborate(self, platform):
...
-                        current_priv_mode = (
-                            self.dependency_manager.get_dependency(CSRInstancesKey()).m_mode.priv_mode.read(m).data
-                        )
+                        access_valid = Signal(1, init=1)
+
+                        if counteren_bit is not None:
+                            with m.If(~mcounteren[counteren_bit] & (current_priv_mode < PrivilegeLevel.MACHINE)):
+                                m.d.comb += access_valid.eq(0)
+                            with m.If(~scounteren[counteren_bit] & (current_priv_mode < PrivilegeLevel.SUPERVISOR)):
+                                m.d.comb += access_valid.eq(0)
```
piotro888: "The setting of a bit in mcounteren does not affect whether the corresponding bit in
scounteren is writable. However, **U-mode may only access a counter if the corresponding bits
in scounteren and mcounteren are both set**"

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -71,15 +73,49 @@ def __init__(self, gen_params: GenParams):
...
+        smode_interrupts = (
+            (1 << InterruptCauseNumber.SSI)
+            | (1 << InterruptCauseNumber.STI)
+            | (1 << InterruptCauseNumber.SEI)
+            | (((1 << gen_params.interrupt_custom_count) - 1) << 16)
+        )
+
+        smode_delegable_exceptions = 0
+        for cause in ExceptionCause:
+            if cause in [ExceptionCause._COREBLOCKS_ASYNC_INTERRUPT, ExceptionCause._COREBLOCKS_MISPREDICTION]:
```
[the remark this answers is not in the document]
tilk: > I would match on names starting with `_` to be more future-proof.

Would add a class method to `ExceptionCause` to be even more future-proof.

coreblocks/priv/csr/csr_instances.py
```diff
@@ -1,28 +1,66 @@
...
+    for counter in counter_list:
+        mask |= 1 << counter
+    return mask
+
+
+def counteren_access_filter(gen_params: GenParams, counteren_bit: int):
+    def _filter(m: TModule, priv_mode: Value) -> ValueLike:
+        csr_instances = DependencyContext.get().get_optional_dependency(CSRInstancesKey())
+        if csr_instances is None:
+            return C(1)
```
awariac: shouldn't that be an error?

coreblocks/arch/isa_consts.py
```diff
@@ -165,6 +165,18 @@ class ExceptionCause(IntEnum, shape=5):
...
+    def smode_delegable_mask(cls, xlen: int) -> int:
+        mask = 0
+        for cause in cls:
+            if cause.name.startswith("_"):
+                continue
+            if cause == cls.ENVIRONMENT_CALL_FROM_M:
+                continue
+            mask |= 1 << cause.value
+
+        return mask
```
awariac, line 181: Is this needed?
What's your opinion on my comment:
> Maybe it would be easier to only blacklist traps 11, 16 as in spec?
Machine code can query if interrupts exists by writing to ie bits. I don't think there is much benefit in further filtering here (but I may be wrong here).

That diff is OK; I'm only curious on your opinion.

qbojj: The spec quote:
> For exceptions that cannot occur in less privileged modes, the corresponding medeleg bits should be read-only zero. In particular, medeleg[11] is read-only zero.
> 
> The medeleg[16] is read-only zero as double trap is not delegatable.


I am not really keen on relying on informal parts of the spec and the part about bit 11 at least looks informal to me. Other than that, just masking bits 11 and 16 would be OK.

coreblocks/priv/csr/csr_instances.py
qbojj: yes

piotro888 (approved):

coreblocks/arch/isa_consts.py
awariac: "**cannot occur** in less privileged modes" ok got it this time

coreblocks/arch/isa_consts.py
```diff
@@ -202,3 +214,23 @@ class PMPAFlagEncoding(IntEnum, shape=2):
     TOR = 1
     NA4 = 2
     NAPOT = 3
+
+
+@unique
+class SatpModeEncoding(IntEnum):
```
tilk: Just `SatpMode`? Add `shape`.

coreblocks/arch/isa_consts.py
```diff
@@ -202,3 +214,23 @@ class PMPAFlagEncoding(IntEnum, shape=2):
...
+    BARE = 0
+    SV32 = 1
+    SV39 = 8
+    SV48 = 9
+    SV57 = 10
+    SV64 = 11
+
+    @classmethod
+    def valid_modes(cls, xlen: int) -> frozenset["SatpModeEncoding"]:
```
tilk: Bikeshedding: would just use `collections.abc.Set` for typing.

coreblocks/arch/csr_address.py
```diff
@@ -458,8 +465,113 @@ class MstatusFieldOffsets(IntEnum):
...
+    HPM25 = 25  # HPMCOUNTER25 enable
+    HPM26 = 26  # HPMCOUNTER26 enable
+    HPM27 = 27  # HPMCOUNTER27 enable
+    HPM28 = 28  # HPMCOUNTER28 enable
+    HPM29 = 29  # HPMCOUNTER29 enable
+    HPM30 = 30  # HPMCOUNTER30 enable
+    HPM31 = 31  # HPMCOUNTER31 enable
+
+
+sstatus_field_subset = {
```
tilk: Maybe use `frozenset` to make sure it's immutable, or type as `Set`, or both?

coreblocks/func_blocks/csr/csr.py
```diff
@@ -135,19 +136,44 @@ def elaborate(self, platform):
...
-                    with m.Case(csr_number):
+                    counteren_bit = None
+                    if csr_number in [CSRAddress.CYCLE, CSRAddress.CYCLEH]:
+                        counteren_bit = 0
+                    elif csr_number in [CSRAddress.TIME, CSRAddress.TIMEH]:
+                        counteren_bit = 1
+                    elif csr_number in [CSRAddress.INSTRET, CSRAddress.INSTRETH]:
+                        counteren_bit = 2
+
+                    with branch(instr.csr == csr_number):
```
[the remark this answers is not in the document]
tilk: Leave a comment about kuznia-rdzeni/transactron#10.

coreblocks/params/configurations.py
```diff
@@ -169,6 +178,11 @@ def __post_init__(self):
     interrupt_custom_edge_trig_mask: int = 0
 
     user_mode: bool = True
+    supervisor_mode: bool = True
+
+    asidlen: int = 0
+    supported_vm_schemes: set[SatpModeEncoding] = field(default_factory=lambda: {SatpModeEncoding.BARE})
```
tilk: Just use `Collection`, like in `func_units_config`, and initialize with a tuple, so that `field` is not needed.

coreblocks/params/vmem_params.py
```diff
@@ -0,0 +1,74 @@
...
+from amaranth.lib.data import StructLayout
+
+from coreblocks.arch.isa_consts import SatpModeEncoding
+
+__all__ = [
+    "VirtualMemoryParams",
+]
+
+
+class SatpLayout(StructLayout):
```
tilk: This is arch-related, shouldn't it be somewhere in `arch`?

coreblocks/params/vmem_params.py
```diff
@@ -0,0 +1,74 @@
...
+                        "ppn": 44,
+                        "asid": 16,
+                        "mode": 4,
+                    }
+                )
+            case _:
+                raise ValueError(f"Unsupported XLEN for SATP layout: {xlen}")
+
+
+class VirtualMemoryParams:
```
tilk: Rename to `VirtualMemoryParameters` to match naming of other two. (I know that `GenParams` has `Params`, but this PR is for features and is not supposed to change established conventions.)

tilk (changes requested):
Minor style-related comments.

coreblocks/arch/isa_consts.py
qbojj: I don't really know how to correctly handle that satp mode has different lengths for RV32 and RV64. could always use RV64 length

coreblocks/params/configurations.py
```diff
@@ -111,6 +111,14 @@ class _CoreConfigurationDataClass:
         read-only and directly connected to input signal (implementation must provide clearing method)
     user_mode: bool
         Enable User Mode.
+    supervisor_mode: bool
+        Enable Supervisor Mode.
+    asidlen: int
+        Number of writable ASID bits in SATP.
+    supported_vm_schemes: Collection[SatpModeEncoding]
```
tilk:
```suggestion
    supported_vm_schemes: Collection[SatpMode]
```

tilk (approved):

tilk:
Fmax seems worrying, but the critical path seems unrelated.

awariac:
it can be verified with other toolchains or/and we should add option for yosys to print more top worst paths (it might hide behind the optimization target).

tilk:
> it can be verified with other toolchains or/and we should add option for yosys to print more top worst paths

This seems like the best course of action.

Still, I don't want this to be a blocker for further S-mode work, and I'm willing to merge now and worry later. What do you think?

qbojj:
It seems like my comment during the meeting was not correct:
- current critical path (~27 ns) is entirely inside of the frontend, not even touching CSRUnit
- CSRUnit is in the <async> -> sync domain (~25 ns) is: bus_master_data_adapter -> atomic wrapper -> announcement -> `CSRUnit_update0_*` -> reservation station

awariac:
`CSRUnit_update0_*` - looks like its also a RS thing  (internal in CSRUnit, no connection with CSR processing)

qbojj:
I have made a version that would not have the `condition` workaround (https://github.com/qbojj/coreblocks/pull/1), the results are not far from what is seen here.

The critical path there also doesn't touch CSRs - it goes around RS and ALU, so I have no idea what caused the loss of FMax.

Maybe the resource usage started to impact the FMax due to contention?

awariac:
mayhaps

but the increase is not that large, and in runs on this PR it differs a lot compared to that - ~3k change, but Fmax doesn't seem to correlate with that.

Utilzization now:
```
 Info: 	          TRELLIS_FF:    8491/  83640    10%
Info: 	        TRELLIS_COMB:   46336/  83640    55%
```

Due to previous comments, I would merge
and we can try later with larger FPGA, comparing this commit to previous one.
But this need to be analyzed more in-depth soon.

tilk:
Merge now and worry later, then.

awariac:
wait, did it disappear on master? https://kuznia-rdzeni.github.io/coreblocks/dev/benchmark/

maybe we forgot how bad it is on full config and this was not an issue. Device utilization may be the case.
we need to observe next commits to see if it still fits in noise floor.

PR #880, opened 2026-04-02
Superscalar rename prep

tilk:
Implementing superscalar rename with CRAT will be simpler if the jump instruction will be the last one.


qbojj (approved):

tilk:
As you are not attending the meetings, I'm going to assume a week without a review to be a silent approval. We need to move faster to achieve project goals.

PR #881, opened 2026-04-04
Integrate PMP

dawidpawliczek4:


coreblocks/priv/csr/csr_instances.py
```diff
@@ -112,7 +112,20 @@ def __init__(self, gen_params: GenParams):
...
+                return Mux(a_field == PMPAFlagEncoding.OFF, value & ~grain_mask, value)
+
+            reg0 = CSRRegister(CSRAddress.PMPADDR0, gen_params, fu_read_map=pmpaddr0_fu_read_map)
+            self.pmpaddrx.append(reg0)
+            setattr(self, "pmpaddr0", reg0)
+
+        for i in range(1, gen_params.pmp_register_count):
             reg = CSRRegister(getattr(CSRAddress, f"PMPADDR{i}"), gen_params)
             self.pmpaddrx.append(reg)
             setattr(self, f"pmpaddr{i}", reg)
```
qbojj: What you have done is taken the NOTE in spec as the rule itself. The notes are non-normative and should only be used for clarifying.

What this note was referencing is:
> When G ≥ 2 and pmpcfgi.A[1] is set, i.e. the mode is NAPOT, then bits pmpaddri[G-2:0] read as all ones. When G ≥ 1 and pmpcfgi.A[1] is clear, i.e. the mode is OFF or TOR, then bits pmpaddri[G-1:0] read as all zeros.

coreblocks/priv/csr/csr_instances.py
qbojj, line 145: The spec says
> When G ≥ 1, the NA4 mode is not selectable.

qbojj (changes requested):
If you could, could you also look at #888 (mostly the part with `MPRV`, but the locking itself could also be done)

coreblocks/priv/csr/csr_instances.py
```diff
@@ -97,23 +97,62 @@ def __init__(self, gen_params: GenParams):
...
             setattr(self, f"pmpcfg{i}", pmpcfg)
 
         self.pmpaddrx = []
+
         for i in range(gen_params.pmp_register_count):
-            reg = CSRRegister(getattr(CSRAddress, f"PMPADDR{i}"), gen_params)
+
+            def make_pmpaddr_fu_read_map(idx: int):
+                def pmpaddr_fu_read_map(_: TModule, value: Value):
+                    a_field = self.pmpxcfg[idx].value[3:5]
```
qbojj: same here, use `PMPCfgLayout`

coreblocks/priv/pmp.py
```diff
@@ -72,13 +71,24 @@ def elaborate(self, platform) -> HasElaborate:
...
-                    napot_mask = addr_val ^ (addr_val + 1)
-                    m.d.comb += entry_match.eq((self.addr[2:] & ~napot_mask) == (addr_val & ~napot_mask))
+                    if grain >= 2:
+                        # Spec: bits [G-2:0] read as all ones in NAPOT mode
+                        forced_val = Signal.like(addr_val, name=f"napot_forced_{i}")
+                        m.d.comb += forced_val.eq(addr_val | ((1 << (grain - 1)) - 1))
+                    else:
+                        forced_val = addr_val
+                    napot_mask = forced_val ^ (forced_val + 1)
+                    m.d.comb += entry_match.eq((self.addr[2:] & ~napot_mask) == (forced_val & ~napot_mask))
```
qbojj: If you just ignore the lower bits, the effect would be the same:
```py
start_bit = max(0, grain - 1)
napot_mask = addr_val[start_bit:] ^ (addr_val[start_bit:] + 1)
m.d.comb += entry_match.eq((self.addr[2 + start_bit:] & ~napot_mask) == (addr_val[start_bit:] & ~napot_mask))
```

coreblocks/priv/csr/csr_instances.py
```diff
@@ -97,23 +97,62 @@ def __init__(self, gen_params: GenParams):
...
+            # When G >= 1, the NA4 mode is not selectable (changed to OFF)
+            to_off_mask = 0b11100111 # Clear A field bits
+            if gen_params.pmp_grain >= 1:
+                a_field = v[3:5]
+                filtered_v = Mux(
+                    a_field == PMPAFlagEncoding.NA4,
+                    v & to_off_mask,
+                    v,
+                )
+                return C(1), filtered_v
```
qbojj: Please use `PMPCfgLayout`

qbojj (changes requested):

qbojj (approved):

coreblocks/frontend/fetch/fetch.py
```diff
@@ -131,13 +133,25 @@ def flush():
         # Fetch - stage 0
         # ================
         # - send a request to the instruction cache
+        # - check PMP execute permission (if PMP is enabled)
         #
+        csr = DependencyContext.get().get_dependency(CSRInstancesKey())
```
awariac: why request it here? `PMPChecker` should do that inside instead

coreblocks/frontend/fetch/fetch.py
```diff
@@ -250,6 +264,19 @@ def _(pc):
...
+            fault = pmp_fault_fifo.read(m)
+            fb_addr = params.fb_addr(fault.pc)
+            s1_s2_pipe.write(
+                m,
+                fb_addr=fb_addr,
+                access_fault=1,
+                instr_valid=0,
+                rvc=0,
+                instrs=[C(0, self.gen_params.isa.ilen)] * fetch_width,
+                instr_block_cross=0,
```
awariac: you set instr_block_cross always to 0.
Please not that with Compressed instructions enabled, regular instructions could cross fetch blocks, and access error an happen only on second half of instructions.

We have special flag that should be emitted for that case here: https://github.com/dawidpawliczek4/coreblocks/blob/e99207937d13ba3b9e9c2312f4464c736b8b3777/coreblocks/frontend/fetch/fetch.py#L393

Because you set instr_block_cross always as zero it would not happen.

piotro888 (changes requested):

coreblocks/params/configurations.py
```diff
@@ -113,6 +113,8 @@ class _CoreConfigurationDataClass:
         Enable User Mode.
     pmp_register_count: int
         Number of Physical Memory Protection CSR entries. Valid values are: 0, 16, and 64.
+    pmp_grain: int
+        Grain parameter of Physical Memory Protection. pmp_grain must be >= icache_line_bytes_log + 2
```
awariac:
```suggestion
        Grain parameter of Physical Memory Protection. `pmp_grain` must be >= icache_line_bytes_log + 2 if PMP registers are enabled.
```

awariac: additionally explain here what it exactly translates you.
G parameter isn't even properly directly explained in spec.

It would be better to define this parameter instead as:
a) Grain in bytes. Then check if it is power of two, cache lines, small values. 
b) logarithm of that = count of const 0 as the end. Check if the value >=2 (?)

G could be computed then in gen_params internally.

choose what you think would fit here better.

Remember that `configurations.py` are end-user facing options that should have extra attention to clarity, and assume less knowledge of spec details.

coreblocks/priv/csr/csr_instances.py
```diff
@@ -97,23 +98,59 @@ def __init__(self, gen_params: GenParams):
         self.pmpxcfg = []
         pmpcfg_subregisters = gen_params.isa.xlen // PMPXCFG_WIDTH
         pmpcfgx_cnt = gen_params.pmp_register_count // pmpcfg_subregisters
+
+        def filter_na4(_: TModule, v: Value):
+            # When G >= 1, the NA4 mode is not selectable (changed to OFF)
+            if gen_params.pmp_grain >= 1:
+                cfg = data.View(PMPCfgLayout(), v)
+                filtered_a = Mux(cfg.A == PMPAFlagEncoding.NA4, PMPAFlagEncoding.OFF, cfg.A)
+                filtered_v = Cat(cfg.R, cfg.W, cfg.X, filtered_a, cfg.reserved, cfg.L)
```
awariac:
```
The R, W, and X fields form a collective WARL field for which the
combinations with R=0 and W=1 are reserved.
```

From what i understand in the spec reserved = illegal (as in WLRL definition)
*WARL* should return any valid encoding in that case, ex. RW=00

coreblocks/priv/csr/csr_instances.py
```diff
@@ -97,23 +98,59 @@ def __init__(self, gen_params: GenParams):
...
+
+            def make_pmpaddr_fu_read_map(idx: int):
+                def pmpaddr_fu_read_map(_: TModule, value: Value):
+                    cfg = data.View(PMPCfgLayout(), self.pmpxcfg[idx].value)
+                    a_field = cfg.A
+
+                    if gen_params.pmp_grain >= 2:
+                        # When G >= 2 and pmpcfgi.A[1] is set, then bits pmpaddri[G-2:0] read as all ones.
+                        napot_mask = (1 << (gen_params.pmp_grain - 1)) - 1
+                        napot_value = value | napot_mask
```
awariac:
```suggestion
     napot_value = Cat(C(1).replicate(gen_params.pmp_grain-1), value[gen_params.pmp_grain-1:])
```

would me more readable.

`C(0, gen_params.pmp_grain)` in second case.

test/asm/pmp_fetch.asm
```diff
@@ -0,0 +1,33 @@
...
+    csrr x1, mcause
+    li x2, 1
+    bne x1, x2, fail  # Check if it is a ACCESS fault
+    j pass
+
+fail:
+    j fail
+
+pass:
+    j pass
```
awariac, line 34: this test doesn't execute!
it needs to be added to `test_core.py`

coreblocks/priv/csr/csr_instances.py
```diff
@@ -97,23 +98,59 @@ def __init__(self, gen_params: GenParams):
...
+                        napot_value = value
+
+                    if gen_params.pmp_grain >= 1:
+                        # When G >= 1 and pmpcfgi.A[1] is clear, then bits pmpaddri[G-1:0] read as all zeros.
+                        tor_mask = (1 << gen_params.pmp_grain) - 1
+                        tor_value = value & ~tor_mask
+                    else:
+                        tor_value = value
+
+                    return Mux(a_field[1], napot_value, tor_value)
```
awariac: use explicit field names instead.

coreblocks/priv/pmp.py
```diff
@@ -72,13 +71,21 @@ def elaborate(self, platform) -> HasElaborate:
...
+                    else:
+                        m.d.comb += entry_match.eq(0)
                 with m.Case(PMPAFlagEncoding.NAPOT):
-                    napot_mask = addr_val ^ (addr_val + 1)
-                    m.d.comb += entry_match.eq((self.addr[2:] & ~napot_mask) == (addr_val & ~napot_mask))
+                    start_bit = max(0, grain - 1)
+                    napot_mask = addr_val[start_bit:] ^ (addr_val[start_bit:] + 1)
+                    m.d.comb += entry_match.eq(
+                        (self.addr[2 + start_bit :] & ~napot_mask) == (addr_val[start_bit:] & ~napot_mask)
+                    )
```
awariac, line 93: it takes a while to see what happens here.
Please explain what the computation does in a comment.

test/asm/user_mode.asm
```diff
@@ -4,6 +4,12 @@
 # trap_handler verifies trap cause and priv mode change
 
 _start:
+    # Configure PMP: allow all access for U-mode
+    li x1, 0x1F
+    csrw pmpcfg0, x1
+    li x1, -1
+    csrw pmpaddr0, x1
```
awariac, line 11: why is it needed? Isn't PMP off by default?

coreblocks/priv/csr/csr_instances.py
```diff
@@ -97,23 +98,59 @@ def __init__(self, gen_params: GenParams):
...
+
+                return pmpaddr_fu_read_map
+
+            reg = CSRRegister(getattr(CSRAddress, f"PMPADDR{i}"), gen_params, fu_read_map=make_pmpaddr_fu_read_map(i))
```
awariac, line 175: `L` bit is not implemented. Writes should be ignored.

coreblocks/priv/csr/csr_instances.py
```diff
@@ -97,23 +98,59 @@ def __init__(self, gen_params: GenParams):
...
+
         for i in range(0, pmpcfgx_cnt):
             # In RV64, odd-numbered configuration registers pmpcfg1, ... pmpcfg15 are illegal.
             pmpcfg_index = i * 2 if gen_params.isa.xlen == 64 else i
             pmpcfg = AliasedCSR(getattr(CSRAddress, f"PMPCFG{pmpcfg_index}"), gen_params)
 
             # pmpcfgX CSR contains a range of pmpYcfg, pmpY+1cfg, ... fields that correspond to pmpaddrY entries
             for j in range(pmpcfg_subregisters):
-                pmpcfg_sub = CSRRegister(None, gen_params, width=PMPXCFG_WIDTH)
+                pmpcfg_sub = CSRRegister(None, gen_params, width=PMPXCFG_WIDTH, fu_write_filtermap=filter_na4)
```
awariac: `L` bit is not implemented. Writes should be ignored.

piotro888 (changes requested):

coreblocks/priv/csr/csr_instances.py
qbojj: Also bits 5-6 should be read only 0

test/asm/user_mode.asm
dawidpawliczek4: we use `full_core_config`, so we have 16 PMP registers and by default user mode has no access

spec: 
> If no PMP entry matches an M-mode memory operation, the operation succeeds. If no PMP entry matches an S-mode or U-mode memory operation, but at least one PMP entry is implemented, the operation fails.

and note:
> If at least one PMP entry is implemented, but all PMP entries’ A fields are set to OFF, then all S-mode and U-mode memory accesses will fail.

coreblocks/frontend/fetch/fetch.py
```diff
@@ -250,6 +262,25 @@ def _(pc):
...
+                m,
+                fb_addr=fb_addr,
+                access_fault=1,
+                instr_valid=Cat(pmp_instr_block_cross, C(0, fetch_width - 1)),
+                rvc=0,
+                instrs=[C(0, self.gen_params.isa.ilen)] * fetch_width,
+                instr_block_cross=pmp_instr_block_cross,
+            )
+
+            m.d.sync += prev_half_v.eq(0)
```
qbojj: Now that I think about it we cannot have 2 independent fifos, as that could lead to messing up the order

qbojj (changes requested):

coreblocks/frontend/fetch/fetch.py
```diff
@@ -173,82 +184,101 @@ def _(pc):
         prev_half_v = Signal()
         with Transaction(name="Fetch_Stage1").body(m):
             fetch_request = fetch_requests.read(m)
-            cache_resp = self.icache.accept_res(m)
```
qbojj, line 176: I think only changing:
```py
access_fault = Signal()
with condition(m) as branch:
    with branch(~fetch_request.pmp_fault):
        cache_resp = self.icache.accept_res(m)
        m.d.comb += access_fault.eq(cache_resp.error)
    with branch(fetch_request.pmp_fault):
        m.d.comb += access_fault.eq(1)
```

where every later reference to `cache_resp.error` is changed to `access_fault` would be enough

awariac: +1

coreblocks/priv/csr/csr_instances.py
awariac: not that `Additionally, if PMP entry i is locked and pmpicfg.A is
set to TOR, writes to pmpaddri-1 are ignored`
This turns out to be a bit more complex, you can make L R/O 0 for now and implement it in other PR if you want.

qbojj (approved):

coreblocks/frontend/fetch/fetch.py
```diff
@@ -173,13 +184,21 @@ def _(pc):
...
             # The address of the fetch block.
             fetch_block_addr = params.fb_addr(fetch_request.pc)
             # The index (in instructions) of the first instruction that we should process.
             fetch_block_offset = params.fb_instr_idx(fetch_request.pc)
 
+            # Conditionally read from icache or mark PMP fault.
+            access_fault = Signal()
+            with condition(m) as branch:
+                with branch(~fetch_request.pmp_fault):
+                    cache_resp = self.icache.accept_res(m)
```
awariac: it's not a great (and elegant) idea to create a python var like that under branch.
Declare a `Signal` with responese layout outside the `condition` block. Then `av_comb` `assign` to it here (due to `av_comb`, performance would be the same)

coreblocks/frontend/fetch/fetch.py
awariac, line 236: I'm now worried about `access_fault` handling:
What if cache response pattern is that `instr_start` turns out 0?
Then due to serializer no instruction would be passed out of fetch, fetch would be stalled, but because of that no exception would be risen, because there is no information. Core hang-ups.
Anyway thats for separate PR.

I will try to look more closely into it in next days, leaving issue for now.

piotro888 (commented):
everything else looks good

tilk:
Please resolve merge conflicts, benchmarks need to be re-checked before merging.

piotro888 (approved):

tilk:
Well, this is not great, the critical path in the `full` config goes through the fetcher now.

PR #882, opened 2026-04-05
Superscalar rename and ROB allocation

tilk:
This PR introduces superscalar register renaming and ROB allocation. This is probably the most involved change in the frontend/scheduler refactor.

~Unfortunately there is a bug somewhere. Regression tests point to jump instructions.~ No bug, just a consequence of borked tagging.

TODO:
* [x] Find the bug.
* [x] Consult CRAT interface change.
* [x] Consult jump handling.
* [x] Fix checkpoint tagging.
* [x] Fix tests.


coreblocks/scheduler/scheduler.py
```diff
@@ -63,39 +62,47 @@ class InstructionTagger(Elaboratable):
...
 
-        data_out = Signal(self.push_instr.layout_in)
+        data_out = Signal(self.push_instrs.layout_in)
 
         with Transaction().body(m):
-            instr = self.get_instr(m)
+            instrs = self.get_instrs(m)
+
+            idx = Signal(range(self.gen_params.frontend_superscalarity))
+            idx.eq(instrs.count - 1)
```
awariac: ummm
```suggestion
            m.d.av_comb += idx.eq(instrs.count - 1)
```

coreblocks/core_structs/crat.py
lekcyjna123, line 1: Please remember about updating doc string

coreblocks/scheduler/scheduler.py
```diff
@@ -121,33 +129,52 @@ def __init__(self, *, gen_params: GenParams):
...
-                rl_dst=instr.regs_l.rl_dst,
-                rp_dst=instr.regs_p.rp_dst,
-                tag=instr.tag,
-                commit_checkpoint=instr.commit_checkpoint,
+            instrs = self.get_instrs(m)
+
+            m.d.av_comb += data_out.count.eq(instrs.count)
+
+            idx = Signal(range(self.gen_params.frontend_superscalarity))
+            idx.eq(instrs.count - 1)
```
awariac: same

coreblocks/scheduler/scheduler.py
```diff
@@ -121,33 +129,52 @@ def __init__(self, *, gen_params: GenParams):
...
+
+            self.crat_commit_checkpoint(
+                m, tag=instrs.data[idx].tag, commit_checkpoint=instrs.data[idx].commit_checkpoint
```
awariac: for correctness is the same, but could be `instrs.data[0].tag`. Would be more accurate and simpler in hw.

piotro888 (commented):
>    All instructions in one group should have the same tag.
>    The tag can only increment in the first instruction of the group.
>   The checkpoint can be taken of the final FRAT state after one cycle, just like before.

The approach you proposed makes sense, I verified the code that is should do it and work well with CRAT.

> The tag handling. I'm not sure if I understand all the corner cases, so please look on that.

looks like it is fine

> The interface between stages. Right now, some fields are repeated between multiple instructions of the group, even though they are guaranteed to be equal. This may change later, though.

I think it can be done in another PR (but is worth doing it).

> I've split the CRAT rename method into two parts, separating checkpoint handling from renaming. This makes sense in the current design, as there is at most one checkpoint taken, but multiple renames are possible. Even later, with possible multiple checkpoints per cycle, it makes sense to have a lower number of checkpoints than instructions because of the amount of data moving required in making checkpoints, 

That makes a lot of sense for now. It can stay as longer-term soulution for now, and we should revisit it later with some measurements when rest of superscalarity is finished and checkpointing with rollbacks is fully implemented. 

> and very long sequences of jumps are improbable, too.

I would be not so sure about that.

coreblocks/core_structs/crat.py
lekcyjna123, line 182: Side note. After refactor, nearly whole logic is in the below `condition`. So it looks like something to be refactored, but after 30min of thinking, I don't see a possibility to do that.

coreblocks/core_structs/crat.py
```diff
@@ -166,11 +167,13 @@ def make_new_checkpoint(*, from_tag: Value):
...
 
-        @def_method(m, self.rename)
-        def _(rp_dst: Value, rl_dst: Value, rl_s1: Value, rl_s2: Value, tag: Value, commit_checkpoint: Value):
+        @def_method(m, self.commit_checkpoint)
+        def _(tag: Value, commit_checkpoint: Value):
+            m.d.av_comb += group_tag.eq(tag)
             # don't overwrite freshly restored FRAT with flushed inactive instructions
-            tag_valid = Signal()
             m.d.av_comb += tag_valid.eq(~frat_lock | (tag == frat_unlock_tag))
             with m.If(tag_valid):
```
lekcyjna123, line 180:
```suggestion
            s_unlock = Signal()
            m.d.top_comb += s_unlock.eq(tag == frat_unlock_tag)
            m.d.av_comb += tag_valid.eq(~frat_lock | s_unlock)
            with m.If(s_unlock):
```

lekcyjna123 (approved):
>The interface between stages. Right now, some fields are repeated between multiple instructions of the group, even though they are guaranteed to be equal. This may change later, though.

I would suggest to leave it duplicated for now and open an issue to not forget about this. Having this done under separate ticket, would allow to analyze if and how big the impact of such optimization is.

> I've split the CRAT rename method into two parts, separating checkpoint handling from renaming. 

Yes. That has sense. I would assume that in first step we will have something like 4 renames, 1 checkpoint. And later 8 renames, 2 checkpoints.

> The tag handling. I'm not sure if I understand all the corner cases, so please look on that.

I checked both in general and for equivalence to the previous code and haven't found any issues.

coreblocks/core_structs/crat.py
awariac: why is this suggestion?

lekcyjna123: Because it is equivalent and slightly shorten path.

coreblocks/scheduler/scheduler.py
tilk: 🤦

coreblocks/core_structs/crat.py
tilk: This logic was unchanged in this PR, the original code is readable, and I don't think this place is in danger of being performance limiting. Would opt for not applying the change here.

coreblocks/core_structs/crat.py
tilk: I also thought about ways of eliminating this use of `condition`, but didn't find any good ideas.

coreblocks/scheduler/scheduler.py
tilk: Done

coreblocks/scheduler/scheduler.py
tilk: Done

piotro888 (approved):

coreblocks/core_structs/crat.py
lekcyjna123: ok

PR #884, opened 2026-04-09
Implement S-mode traps

qbojj:
This builds up on #878 and implements trap handling.

This PR implements:
- interrupt and exception delegation
- `SRET` and blank `SFENCE.VMA`
- tests for S-mode traps


coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -175,19 +265,34 @@ def _():
...
+        if self.gen_params.supervisor_mode:
+
+            @def_method(m, self.sret)
+            def _():
+                log.info(m, True, "Supervisor interrupt handler return")
+                pass
+
         @def_method(m, self.entry)
-        def _():
+        def _(target_priv):
```
awariac: maybe target_priv can be logged here?
also there is some leftover pass, please remove it (unless its somehow needed)

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -130,19 +145,94 @@ def elaborate(self, platform):
...
+            self.interrupt_insert.eq(m_interrupt_insert | s_interrupt_insert),
+            interrupt_pending.eq((mie & mip).any()),
+        ]
+
+        with m.If(m_interrupt_insert):
+            m.d.comb += selected_pending.eq(pending_m)
+        with m.Elif(s_interrupt_insert):
+            m.d.comb += selected_pending.eq(pending_s)
+        with m.Else():
+            m.d.comb += selected_pending.eq(0)
```
awariac: It could be just a `Mux`, 0 case is not needed

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -130,19 +145,94 @@ def elaborate(self, platform):
...
-        m.d.comb += self.interrupt_insert.eq(interrupt_pending & interrupt_enable)
+        m_interrupt_insert = Signal()
+        s_interrupt_insert = Signal()
+        interrupt_pending = Signal()
+
+        m.d.comb += [
+            m_interrupt_insert.eq(pending_m.any() & interrupt_enable_m),
+            s_interrupt_insert.eq(pending_s.any() & interrupt_enable_s),
+            self.interrupt_insert.eq(m_interrupt_insert | s_interrupt_insert),
+            interrupt_pending.eq((mie & mip).any()),
```
awariac: this signal could be merged with previous `pending`, but declared at module-level

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -175,19 +265,34 @@ def _():
...
+        # this is split here to avoid complicated call graphs and conflicts that are not handled well by Transactron
         with Transaction().body(m):
             with m.If(self.entry.run):
-                self.mstatus_mie.write(m, {"data": 0})
-                self.mstatus_mpie.write(m, self.mstatus_mie.read(m).data)
-                self.mstatus_mpp.write(m, priv_mode.read(m).data)
-                priv_mode.write(m, PrivilegeLevel.MACHINE)
+                with m.Switch(self.entry.data_in.target_priv):
+                    if self.gen_params.supervisor_mode:
+                        with m.Case(PrivilegeLevel.SUPERVISOR):
```
awariac, line 272: maybe assertion here that trap is from user | supervisor mode?

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -130,19 +145,94 @@ def elaborate(self, platform):
...
+                    ),
+                ]
+            else:
+                m.d.comb += [
+                    pending_m.eq(pending),
+                    pending_s.eq(0),
+                    interrupt_enable_s.eq(0),
+                ]
+
+            m.d.comb += interrupt_enable_m.eq(self.mstatus_mie.read(m).data | (priv < PrivilegeLevel.MACHINE))
```
awariac: use `av_comb` in transactions that should always run

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -130,19 +145,94 @@ def elaborate(self, platform):
...
+            m.d.av_comb += Cat(cause_num, is_interrupt).eq(cause)
+
+            edeleg = Signal(64)
+            ideleg = self.mideleg.read(m).data
+            current_priv = priv_mode.read(m).data
+
+            if self.medelegh:
+                m.d.av_comb += edeleg.eq(Cat(self.medeleg.read(m).data, self.medelegh.read(m).data))
+            else:
+                m.d.av_comb += edeleg.eq(self.medeleg.read(m).data)
```
awariac: I think we need some wraper for RV32 to group `*h` registers together and provide single `read`.
Out of scope here.

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -130,19 +145,94 @@ def elaborate(self, platform):
...
+
+            target_priv = Signal(PrivilegeLevel)
+            with m.If(is_interrupt):
+                m.d.av_comb += target_priv.eq(
+                    Mux(ideleg.bit_select(cause_num, 1), PrivilegeLevel.SUPERVISOR, PrivilegeLevel.MACHINE)
```
awariac:
```suggestion
                    Mux(ideleg[cause_num], PrivilegeLevel.SUPERVISOR, PrivilegeLevel.MACHINE)
```
nit: a bit simpler

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -130,19 +145,94 @@ def elaborate(self, platform):
 
         priv_mode = self.dm.get_dependency(CSRInstancesKey()).m_mode.priv_mode
 
-        interrupt_enable = Signal()
+        @def_method(m, self.get_trap_target_priv)
+        def _(cause):
```
awariac: I don't like that its feeding back already computed cause (but only for interrupts), but I don't have ideas how to improve it. I see this is necessary for exceptions, and to be done at last possible point.
Some things I was thinking about - maybe `trap_entry` should return target priv instead of requiring it? Or maybe this would fit more into `ExceptionInformationRegister` that we need to call too?
The split about ways of  exceptions and interrupts is probelmatic here.

If we don't come up with any better ideas, this could be left as-is.


Also in case of interrupts this is duplicating computation of top interrupt. Could this be unified?

coreblocks/backend/retirement.py
```diff
@@ -176,11 +188,24 @@ def flush_instr(rob_entry):
...
+                            with m.Switch(target_priv):
+                                if self.gen_params.supervisor_mode:
+                                    with m.Case(PrivilegeLevel.SUPERVISOR):
+                                        assert s_csr is not None
+                                        set_trap_csrs(s_csr.scause, s_csr.sepc, s_csr.stval)
+                                with m.Case(PrivilegeLevel.MACHINE):
+                                    set_trap_csrs(m_csr.mcause, m_csr.mepc, m_csr.mtval)
+
+                            m.d.sync += trap_target_priv.eq(target_priv)
+                            self.trap_entry(m, target_priv=target_priv)
```
piotro888: I think it would make more sense to get `target_priv` as output form `trap_entry` instead of wiring new method through the core

coreblocks/params/configurations.py
```diff
@@ -233,7 +233,7 @@ def replace(self, **kwargs) -> Self:
                 ShiftUnitComponent(),
                 JumpComponent(),
                 ExceptionUnitComponent(),
-                PrivilegedUnitComponent(),
+                PrivilegedUnitComponent(supervisor_enable=True),
```
awariac, line 252: It's not very convenient; this should be inferred form `gen_params`

coreblocks/interface/layouts.py
```diff
@@ -767,6 +768,14 @@ def __init__(self, gen_params: GenParams):
 
         self.data_layout = self.rs.data_layout
 
+        self.imm_layout = make_layout(
+            ("imm", 5),
+            ("_1", gen_params.isa.xlen - 5 - 2 * gen_params.isa.reg_cnt_log),
+            ("rd", gen_params.isa.reg_cnt_log),
+            ("rs1", gen_params.isa.reg_cnt_log),
+        )
+        assert self.imm_layout.size == gen_params.isa.xlen
```
awariac, line 807: please document this layout here (and reason for it)

unrelated - yay, finally it can be done nicely with amaranth Views :)

piotro888 (commented):
Very nice work!

3/4 part of review - some comments and discussions:

coreblocks/priv/traps/interrupt_controller.py
qbojj: `TypeError: Cannot index value with a value; use Value.bit_select() instead`

awariac: oooh, sorry i forgot about that

coreblocks/params/configurations.py
qbojj: I had a problem with `get_decoder_manager` not getting gen params

tilk: Huh. The mechanism with `get_decoder_manager` was designed with more local effects in mind (e.g. ALU: parameter added -> new instructions supported -> new extension automatically inferred). I see two ways of avoiding redundancy in config:

* Treat this like extensions - declaring `supervisor_enable` on any `PrivilegedUnit` automatically makes the whole core support supervisor mode.
* Pass `gen_params` to `get_decoder_manager`, and therefore allow the set of supported instructions to change after enabling supervisor mode in the configuration.

Don't have a strong opinion on this, maybe I like the second option slightly more.

awariac: second one would be more consistent with `user_mode`.
Maybe it could accept core_config instead of gen_params?
The problem is gen_params's ISA string is determined from FU configs, so that's kind of circular dependency. It's not a problem in this case, but it could create problems for other units if we add this to standard interface.
@tilk what do you think about that?

tilk: You are right about the circular depdendency, passing the configuration seems like the best choice here.

awariac:
there are also architectural tests for supervisor mode that need to be enabled. you can do it here on in other PR, but its worth checking for correctness

awariac:
in one (or both) of tests/external/{riscv-dv, riscof}, i don't remember now

awariac:
There are some macros in `test/external/riscof/coreblocks/env/model_test.h` that look like they may be needed to implement (copy from other riscof usages for other cores) - if the issue doesn't resolve with some trivial fix.
You can move riscof to separate PR to merge CSR changes sooner.

qbojj:
I think I will leave riscof to another PR

coreblocks/params/configurations.py
qbojj: the problem is that this object is the part of the configuration...

coreblocks/priv/traps/interrupt_controller.py
qbojj: I have redone the interface - now `entry` returns the privilege. There is still duplication of interrupt target logic - I was wandering if we could just have it saved like `interrupt_cause`, but I'm not sure if it would not be some sort of TOC-TOU broblem

coreblocks/params/configurations.py
awariac: ok, I think you can add assert to `PrivilegedUnitComponent` that `supervisor_mode == gen_params.supervisor_mode` and maybe we will come up with something better later

coreblocks/priv/traps/interrupt_controller.py
awariac: I'm also a bit worried about TOCTOU issues here.

The only part I was referring to was that we already do `selected_pending` computation based on `m_interrupt_insert` and that's equivalent to the code here.
`get_interrupt_cause` happens at the same time in the retirement as `trap_entry` and represents the state in the same cycle.

However, I have no idea how to connect that cleanly, and hardware cost is not that large either (one xlenwide mux).
The retirement implementation could change in the future as well.

It's fine to leave it as its now, its much better.

awariac (approved):

qbojj:
It looks like S-mode changes themselves pass riscof tests https://github.com/kuznia-rdzeni/coreblocks/pull/921. The only problem seems to be the Zifencei

PR #889, opened 2026-04-13
Superscalar RS insertion

### LINKED ISSUE #890: Generalize and refactor FU instruction count metrics
Currently, some FUs (ALU, JumpBranch, Priv) have metric counters on processed instructions. These count all instructions, including rollbacked ones (#832), but there are other problems too:

* Code duplication.
* Names are constants - which don't allow multiple instances of a FU, which blocks superscalarity.

### LINKED ISSUE #905: Superscalar instruction scheduling
Modify the instruction scheduler so that it can process multiple instructions in one cycle.

tilk:
This PR implements superscalar RS selection and insertion, making the whole scheduler superscalar.

~At this point, instructions are already in ROB and there is no reason to keep them ordered anymore. The pipeline is therefore made branched.~

TODO:
* [x] Redesign the PR so that RS selection is done in-order.
* [x] Add another ALU func block
* [x] Fix RS select test
* [x] Fix #890 
* [x] Add assert to RS insertion stage

Closes #905.

### DIFF
--- a/coreblocks/scheduler/scheduler.py
+++ b/coreblocks/scheduler/scheduler.py
@@ -368,16 +368,15 @@ class RSInsertion(Elaboratable):
             core_state = DependencyContext.get().get_dependency(CoreStateKey())
             flushing = core_state(m).flushing
 
-            tag_inactive = ~self.crat_active_tags(m).active_tags[instrs.data[0].tag]
-
-            skip_source_registers = flushing | tag_inactive
-
+            active_tags = self.crat_active_tags(m)
             rs_entry_id: list[Value] = []
             rs_selected: list[Value] = []
             rs_datas = []
 
             for i in range(self.gen_params.frontend_superscalarity):
                 instr = instrs.data[i]
+                tag_inactive = ~active_tags.active_tags[instr.tag]
+                skip_source_registers = flushing | tag_inactive
 
                 # RS insertion guarantees RF response for present instructions
                 # Nested transaction used to avoid locking
--- a/coreblocks/scheduler/scheduler.py
+++ b/coreblocks/scheduler/scheduler.py
@@ -292,7 +292,8 @@ class RSSelection(Elaboratable):
                 for j, (alloc, block_params) in enumerate(zip(self.rs_select, self.gen_params.func_units_config)):
                     # checks if RS can perform this kind of operation
                     optype_matches = Cat(lookup == op for op in block_params.get_optypes()).any()
-                    with Transaction().body(m, ready=(i < instrs.count) & prev_insert & optype_matches):
+                    tr = Transaction(name=f"RSSelection_{i}_{j}")
+                    with tr.body(m, ready=(i < instrs.count) & prev_insert & optype_matches):
                         allocated_field = alloc(m)
 
                         m.d.comb += instr_out.rs_entry_id.eq(allocated_field.rs_entry_id)
--- a/coreblocks/scheduler/scheduler.py
+++ b/coreblocks/scheduler/scheduler.py
@@ -4,6 +4,7 @@ from amaranth import *
 
 from transactron import Method, Methods, Required, Transaction, TModule
 from transactron.lib import Connect, Pipe, WideFifo
+from transactron.lib import logging
 from transactron.utils import OneHotSwitchDynamic, assign, AssignType
 from transactron.utils.dependencies import DependencyContext
 
@@ -13,6 +14,9 @@ from coreblocks.arch.optypes import OpType
 from coreblocks.interface.keys import CoreStateKey
 
 __all__ = ["Scheduler"]
+
+
+log = logging.HardwareLogger("frontend.scheduler")
 
 
 class RegAllocation(Elaboratable):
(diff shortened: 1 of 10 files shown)


tilk:
Can't reproduce the timeout locally.

tilk:
I think I know what is wrong. My original assumption was incorrect: letting later instruction into RSs before earlier ones might lead to overfilling the RSs, which in some cases might lead to core lock-up. For this reason, RS selection still needs to be done in-order.

tilk:
Good news: superscalar frontend is now complete and the core now eats multiple instructions per cycle 🎉
Bad news: bad branch prediction basically eats most of the benefits 😞

tilk:
Welp, this hit kuznia-rdzeni/transactron#132!

coreblocks/scheduler/scheduler.py
```diff
@@ -345,37 +368,66 @@ def elaborate(self, platform):
...
+                    source2 = self.rf_read_resp[2 * i + 1](m, reg_id=instr.regs_p.rp_s2)
+                log.assertion(m, tr.run == (i < instrs.count), f"invalid RF response for instr {i}")
+
+                rs_data = Signal(self.gen_params.get(RSFullDataLayout).data_layout)
+                m.d.av_comb += assign(
+                    rs_data,
+                    {
+                        # when operand value is valid the convention is to set operand source to 0
+                        "rp_s1": Mux(source1.valid | skip_source_registers, 0, instr.regs_p.rp_s1),
+                        "rp_s2": Mux(source2.valid | skip_source_registers, 0, instr.regs_p.rp_s2),
```
qbojj, line 400: `skip_source_registers` is using tag of the first instruction in the block. Shouldn't it be `flushing | ~active_tags[instrs.data[i].tag]` or `flushing | (i < instrs.count)`?

qbojj (changes requested):

coreblocks/scheduler/scheduler.py
tilk: I believe you're right. Because of the `WideFifo` before RS selection/insertion, we can no longer assume that the tag is the same for all instructions in the group.

coreblocks/scheduler/scheduler.py
```diff
@@ -345,37 +368,65 @@ def elaborate(self, platform):
...
+
                 arg = Signal.like(rs_insert.data_in)
-                m.d.comb += assign(arg, data, fields=AssignType.COMMON)
-                # this assignment truncates signal width from max rs_entry_bits to target RS specific width
-                m.d.comb += arg.rs_entry_id.eq(instr.rs_entry_id)
+                for i in OneHotSwitchDynamic(m, matches):
+                    # connect only matching fields
+                    m.d.av_comb += assign(arg.rs_data, rs_datas[i], fields=AssignType.COMMON)
+                    # this assignment truncates signal width from max rs_entry_bits to target RS specific width
+                    m.d.av_comb += arg.rs_entry_id.eq(rs_entry_id[i])
```
piotro888: does it work, because of RSSelection conflict on `alloc`s we would never insert two instructons in the same package to the same RS?
If yes this should be commented in `RSSelection` and maybe here too. It's non-obvious.

coreblocks/scheduler/scheduler.py
```diff
@@ -272,26 +276,45 @@ def decode_optype_set(self, optypes: set[OpType]) -> int:
...
+                next_insert = Signal()
+                instr = instrs.data[i]
+                instr_out = data_out.data[i]
+                lookup = Signal(OpType)  # lookup of currently processed optype
+                m.d.av_comb += lookup.eq(instr.exec_fn.op_type)
+                m.d.av_comb += assign(instr_out, instr)
+
+                for j, (alloc, block_params) in enumerate(zip(self.rs_select, self.gen_params.func_units_config)):
+                    # checks if RS can perform this kind of operation
+                    optype_matches = Cat(lookup == op for op in block_params.get_optypes()).any()
```
awariac, line 299: we could add an assertion that for each instruction some RS must match. this caused bugs in the past.

coreblocks/scheduler/scheduler.py
```diff
@@ -272,26 +276,45 @@ def decode_optype_set(self, optypes: set[OpType]) -> int:
...
+                    tr = Transaction(name=f"RSSelection_{i}_{j}")
+                    with tr.body(m, ready=(i < instrs.count) & prev_insert & optype_matches):
+                        allocated_field = alloc(m)
```
awariac, line 304: please add an comment that for each instruction set, we will never assign two to the same RS, because of inner transactions' `alloc` conflicts.
This is non-obvious at the first read.

coreblocks/scheduler/scheduler.py
```diff
@@ -272,26 +276,45 @@ def decode_optype_set(self, optypes: set[OpType]) -> int:
...
+                        m.d.comb += instr_out.rs_entry_id.eq(allocated_field.rs_entry_id)
+                        m.d.comb += instr_out.rs_selected.eq(j)
 
-        for i, (alloc, block_params) in enumerate(zip(self.rs_select, self.gen_params.func_units_config)):
-            # checks if RS can perform this kind of operation
-            optype_matches = Cat(lookup == op for op in block_params.get_optypes()).any()
-            with Transaction().body(m, ready=optype_matches):
-                instr = self.get_instr(m)
-                allocated_field = alloc(m)
+                        m.d.comb += next_insert.eq(1)
```
awariac, line 309: Question: I see the bug you described in the original implementation.
Do you think it is worth taking all instructions that don't conflict on the same RS instead of up-to-the first conflict, but doing it in a similar manner that is done here?
This way, when looking per-RS, everything is still in order and the core should not lock-up.

That would unfortunately require some more complex/wasteful data structure, or somehow coordinating between independent branches.

awariac (approved):

qbojj (approved):

coreblocks/scheduler/scheduler.py
tilk: Conflicts on calls is how Transactron works, and the RS `alloc` call is the first line of the inner transaction. This absolutely should be obvious! But will add the comment regardless.

coreblocks/scheduler/scheduler.py
tilk: Don't know, but worth remembering.

PR #891, opened 2026-04-13
Prepare for MMU implementation

qbojj:
This builds up on #878 and prepares for virtual memory implementation.

Things that this PR tries to do:
- separate physical and virtual addresses:
  - allow for configuration of physical address length independent of XLEN - up to 34 bit for rv32 and 56 bit for rv64
  - rename `addr` to `paddr` where physical addresses are involved
- implement blank address translation module for later implementation of other addressing modes

### DIFF
--- a/coreblocks/func_blocks/fu/lsu/lsu_requester.py
+++ b/coreblocks/func_blocks/fu/lsu/lsu_requester.py
@@ -106,17 +106,23 @@ class LSURequester(Elaboratable):
         m = TModule()
 
         m.submodules.args_fifo = args_fifo = BasicFifo(
-            [("addr", self.gen_params.isa.xlen), ("funct3", Funct3), ("store", 1)], self.depth
+            [
+                ("paddr", self.gen_params.phys_addr_bits),
+                ("addr", self.gen_params.isa.xlen),
+                ("funct3", Funct3),
+                ("store", 1),
+            ],
+            self.depth,
         )
 
         @def_method(m, self.issue)
-        def _(addr: Value, data: Value, funct3: Value, store: Value):
+        def _(paddr: Value, addr: Value, data: Value, funct3: Value, store: Value):
             exception = Signal()
             cause = Signal(ExceptionCause)
 
-            aligned = self.check_align(m, funct3, addr)
-            bytes_mask = self.prepare_bytes_mask(m, funct3, addr)
-            bus_data = self.prepare_data_to_save(m, funct3, data, addr)
+            aligned = self.check_align(m, funct3, paddr)
+            bytes_mask = self.prepare_bytes_mask(m, funct3, paddr)
+            bus_data = self.prepare_data_to_save(m, funct3, data, paddr)
 
             self.log.debug(
                 m,
@@ -131,12 +137,12 @@ class LSURequester(Elaboratable):
 
             with condition(m, nonblocking=True) as branch:
                 with branch(aligned & store):
-                    self.bus.request_write(m, addr=addr >> 2, data=bus_data, sel=bytes_mask)
+                    self.bus.request_write(m, addr=paddr >> 2, data=bus_data, sel=bytes_mask)
                 with branch(aligned & ~store):
-                    self.bus.request_read(m, addr=addr >> 2, sel=bytes_mask)
+                    self.bus.request_read(m, addr=paddr >> 2, sel=bytes_mask)
 
             with m.If(aligned):
-                args_fifo.write(m, addr=addr, funct3=funct3, store=store)
+                args_fifo.write(m, paddr=paddr, addr=addr, funct3=funct3, store=store)
             with m.Else():
                 m.d.av_comb += exception.eq(1)
                 m.d.av_comb += cause.eq(
@@ -163,7 +169,7 @@ class LSURequester(Elaboratable):
                     fetched = self.bus.get_read_response(m)
                     m.d.comb += err.eq(fetched.err)
                     m.d.top_comb += data.eq(
-                        self.postprocess_load_data(m, request_args.funct3, fetched.data, request_args.addr)
+                        self.postprocess_load_data(m, request_args.funct3, fetched.data, request_args.paddr)
                     )
 
             with m.If(err):
--- a/coreblocks/func_blocks/fu/lsu/lsu_requester.py
+++ b/coreblocks/func_blocks/fu/lsu/lsu_requester.py
@@ -8,7 +8,7 @@ from transactron.lib import BasicFifo
 from coreblocks.params import *
 from coreblocks.arch import Funct3, ExceptionCause
 from coreblocks.peripherals.bus_adapter import BusMasterInterface
-from coreblocks.interface.layouts import LSULayouts
+from coreblocks.interface.layouts import CommonLayoutFields, LSULayouts
 
 
 class LSURequester(Elaboratable):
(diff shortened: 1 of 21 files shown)


qbojj:
I will open this once #881 is merged

qbojj:
as #881 doesn't seem to be getting any activity, I am opening this PR

coreblocks/interface/layouts.py
```diff
@@ -717,7 +740,7 @@ def __init__(self, gen_params: GenParams):
 
         self.store: LayoutListField = ("store", 1)
 
-        self.issue = make_layout(fields.addr, fields.data, fields.funct3, self.store)
+        self.issue = make_layout(fields.paddr, fields.addr, fields.data, fields.funct3, self.store)
```
awariac: maybe paddr and vaddr for clarity if both are used,
same for `AddressTranslationLayouts` and LSU. Idk if it is not too problematic though

coreblocks/func_blocks/fu/lsu/lsu_requester.py
awariac: log paddr?

coreblocks/params/configurations.py
```diff
@@ -117,6 +117,8 @@ class _CoreConfigurationDataClass:
         Number of writable ASID bits in SATP.
     supported_vm_schemes: Collection[SatpMode]
         SATP MODE values accepted by this core.
+    phys_addr_bits: int | None
+        Width of physical addresses in bits. If not set, defaults to 34 for RV32 and 56 for RV64.
```
awariac: this should be determined from max of `supported_vm_schemes`.
It it a reason to do it in separate parameter?
Does spec permit smaller address space?

coreblocks/priv/vmem/translation.py
```diff
@@ -0,0 +1,176 @@
...
+                {
+                    "V": 1,
+                    "R": 1,
+                    "W": 1,
+                    "X": 1,
+                    "U": 1,
+                    "G": 1,
+                    "A": 1,
+                    "D": 1,
+                    "_rsv": 2,
```
awariac: thats explicitly names RSW in spec, i think it should stay like that

coreblocks/priv/vmem/translation.py
```diff
@@ -0,0 +1,176 @@
...
+                    m.d.av_comb += effective_priv_mode.eq(priv_mode)
+
+            effective_satp_mode = Signal(SatpMode, init=SatpMode.BARE)
+
+            if self.gen_params.supervisor_mode:
+                with m.If(effective_priv_mode < PrivilegeLevel.MACHINE):
+                    m.d.av_comb += effective_satp_mode.eq(csr.s_mode.satp_mode)
+
+            poffset = Signal(12)
+            ppn = Signal(self.gen_params.phys_addr_bits - 12)
```
awariac: page offset should be in isa consts and `12` replaced everywhere

coreblocks/priv/vmem/translation.py
```diff
@@ -0,0 +1,176 @@
...
+            max_ppn = 1 << (self.gen_params.phys_addr_bits - 12) - 1
+
+            vpn_invalid = Signal()
+            with m.Switch(effective_satp_mode):
+                for mode in self.gen_params.vmem_params.supported_schemes - {SatpMode.BARE}:
+                    vpn_len = bits_per_level(mode) * level_count(mode)
+
+                    with m.Case(mode):
+                        # virtual modes require sign-extended vaddr
+                        m.d.av_comb += vpn_invalid.eq(vpn[vpn_len:].any() & ~vpn[vpn_len:].all())
```
awariac, line 148: can you point where is that in spec? Can't find it now

coreblocks/priv/vmem/translation.py
```diff
@@ -0,0 +1,176 @@
...
+                with m.If(effective_priv_mode < PrivilegeLevel.MACHINE):
+                    m.d.av_comb += effective_satp_mode.eq(csr.s_mode.satp_mode)
+
+            poffset = Signal(12)
+            ppn = Signal(self.gen_params.phys_addr_bits - 12)
+            vpn = Signal(self.gen_params.isa.xlen - 12)
+
+            m.d.av_comb += poffset.eq(addr[:12])
+            m.d.av_comb += vpn.eq(addr[12:])
+
```
awariac, line 138: Cat(poffset, vpn).eq?

coreblocks/priv/vmem/translation.py
```diff
@@ -0,0 +1,176 @@
...
+                    "N": 1,
+                }
+            )
+        case _:
+            raise ValueError(f"Unsupported SATP mode: {mode}")
+
+
+def bits_per_level(mode: SatpMode) -> int:
+    num_entries = (1 << 12) // page_table_entry_format(mode).as_shape().width
+    return num_entries.bit_length() - 1
```
awariac: leave some comment what this computation is for

coreblocks/priv/vmem/translation.py
qbojj: > 18.1.2.1. Definitions
Address validity: The RISC-V privileged spec defines validity of addresses based on the address translation mode that is currently in use (e.g., Sv57, Sv48, Sv39, etc.). For a virtual address to be valid, all bits in the unused portion of the address must be the same as the Most Significant Bit (MSB) of the used portion. For example, when page-based 48-bit virtual memory (Sv48) is used, load/store effective addresses, which are 64 bits, must have bits 63–48 all set to bit 47, or else a page-fault exception will occur. For physical addresses, validity means that bits XLEN-1 to PABITS are zero, where PABITS is the number of physical address bits supported by the processor.

Also there is a NOTE for the virtual memory

coreblocks/params/configurations.py
qbojj: spec permits some physical addresses to be invalid. If we wanted it could be setup as less than spec defined maximum to reduce resource usage.

coreblocks/priv/vmem/translation.py
qbojj: Also:
For SV39
> Instruction fetch addresses and load and store effective addresses, which are 64 bits, must have bits 63–39 all equal to bit 38, or else a page-fault exception will occur.

Similar for greater vmem schemes

coreblocks/params/configurations.py
awariac: ok, thanks. Can we change it so default it max of satp modes? - then it would only need to be set when resource reduction is needed.

awariac: dostring not updated

piotro888 (approved):

coreblocks/params/vmem_params.py
```diff
@@ -1,11 +1,19 @@
...
+from coreblocks.arch.isa_consts import SatpMode, SatpLayout, PAGE_SIZE_LOG
 
 
 class VirtualMemoryParameters:
     """Parameters for virtual memory support."""
 
+    @staticmethod
+    def max_physical_address_bits(xlen: int, supported_schemes: Collection[SatpMode]) -> int:
+        if xlen == 32 and supported_schemes == {SatpMode.BARE}:
+            return 32  # without virtual memory only lower 32 bits of physical addresses are reachable
```
awariac, line 12: wouldn't it be also 64 if BARE is enabled on xlen == 64?

awariac: or is it intentional as only virtual memory use is 'sane' at 64bit space? That would be also good.

qbojj: actually ISA explicitly defines physical address space as 56 bits for RV64.

qbojj: This can be seen e.g. in `pmpaddr` (upper bits of this register on RV64 are R/O 0)

awariac: > actually ISA explicitly defines physical address space as 56 bits for RV64.

uh i was trying to find that, looks like i'm bad at searching today. In both cases thats resolved

tilk:
Just merged the PMP PR, there is some conflict resolving to do unfortunately.

tilk (approved):

PR #895, opened 2026-04-14
add all fixes pre-commit wants to make

qbojj:
Effectively I had run:
```
pre-commit run --all-files
```

Mostly fixing end-of-line whitespaces, but also docstring in long division not being at the top of the file, trailing empty lines and sorting of requirements.txt

I have also excluded `test/external/**` from the pre-commit - as `./scripts/lint.sh` doesn't like some files there


tilk:
Can this be checked in CI so that this never has to be done again?

tilk:
Also, can this and #894 be done also in Transactron?

.github/workflows/main.yml
```diff
@@ -355,6 +355,9 @@ jobs:
           python -m pip install --upgrade pip
           pip3 install -r requirements-dev.txt
 
+      - name: Check pre-commit hooks
+        run: pre-commit run --all-files --show-diff-on-failure
```
tilk: This runs formatting and typing, too. How is error reporting?

I see two options:

* Run this after `check_format` and `check_types`, so that any errors found will be reported in earlier steps.
* If there is no regression on error reporting, maybe remove `check_format` and `check_types` completely.

@lekcyjna123 thoughts?

.pre-commit-config.yaml
tilk: Probably needs to be removed as in Transactron.

awariac:
what's wrong with images? I think it should be skipped on binary files (and svg in this case being a edge-case i guess).

I'm not a fan of dependency (or any other) sorting, it can get a bit annoying to fix for contributors, that don't want to use precommit hook (like me).

tilk:
> what's wrong with images? I think it should be skipped on binary files (and svg in this case being a edge-case i guess).

Final newline stuff I think. SVGs are NOT binary files, but they are usually not edited with text editors, so they should be left alone.

> I'm not a fan of dependency (or any other) sorting, it can get a bit annoying to fix for contributors, that don't want to use precommit hook (like me).

Indeed, this is such a minor thing, too.

qbojj:
It looks like pre-commit on error even tells:
```
If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.
```

tilk (approved):
LGTM

PR #899, opened 2026-04-16
Introduce data cache

dawidpawliczek4:


coreblocks/params/configurations.py
```diff
@@ -161,6 +169,11 @@ def __post_init__(self):
     icache_sets_bits: int = 7
     icache_line_bytes_log: int = 5
 
+    dcache_enable: bool = True
+    dcache_ways: int = 2
+    dcache_sets_bits: int = 7
+    dcache_line_bytes_log: int = 5
```
tilk, line 185: Put in dataclass?

awariac: (prob. together with icache)

PR #903, opened 2026-04-17
Publish to PyPI

tilk:
With `pyproject.toml` set, we can now publish the package to PyPI. This workflow is adapted from Transactron and publishes to both PyPI and TestPyPI. The TestPyPI deployments happen on every master commit and serve as convenience and verification of the deployment setup. The PyPI deployments happen on adding a `v*` tag. As agreed during the meeting, we will use `0.Y.M.D` versioning scheme.


awariac:
maybe `0.0.Y.M.D` to not get right to `1.*` when we would change to stable versioning

awariac:
Don't we want automatic weekly / bi-weekly releases (if changes are present)?
I would not like to introduce the burden of deciding if we should tag a commit or not.

tilk:
Having tags on releases is generally a good practice. The tags could be generated automatically in the workflow, too.

Changing to draft for now, will set as ready to review after we agree on the implementation.

tilk:
> maybe `0.0.Y.M.D` to not get right to `1.*` when we would change to stable versioning

A little overkill? @Arusekk @lekcyjna123 your thoughts?

Arusekk:
0.Y.M.D should be enough. I personally prefer unstable projects with 1.0 versions to projects with 0.x stable versions.

tilk:
If Linus is cool enough for AI, so am I ;) Slopped together a workflow which should:

- Create weekly version tags.
- Publish to PyPI with those tags.
- Allow creating a new tag and publishing a new version with workflow dispatch.
- Publish every master commit to TestPyPI.

Looks fine to me, but I would like a second pair of eyes to check.

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,189 @@
...
+
+on:
+  push:
+    branches:
+      - 'master'
+    tags:
+      - 'v**'
+  schedule:
+    # Every Sunday at 00:00 UTC
+    - cron: '0 0 * * 0'
```
lekcyjna123: Please add some offset to don't start at 0:00 (there is probably big traffic then). Is the `h` specifier available in GithubActions (h = hasz)? If not please start at something like 0:14 (a guessed random number).

lekcyjna123 (approved):

PR #904, opened 2026-04-18
Big CSR refactor - CSR Register ABC, Protocol, misc clean-ups

### LINKED ISSUE #897: Define CSR protocol
Src: TODO in code

piotro888:
- Removes duplication on creation of typical CSR methods
- Defines how CSR register should look like
- Removes dubious use of `CSRRegister` as a base class
- Defines what we expect from FU registerable CSR 
- Clean-ups, move of classes to new files, changes to unify things

- [X] todo docstrings

### DIFF
--- a/coreblocks/func_blocks/csr/csr_unit.py
+++ b/coreblocks/func_blocks/csr/csr_unit.py
@@ -1,8 +1,10 @@
 from amaranth import *
 from amaranth.lib.data import StructLayout
-from dataclasses import dataclass
 
-from transactron import Method, Methods, def_method, def_methods, Transaction, TModule
+from dataclasses import dataclass
+from typing import Protocol
+
+from transactron import Method, Methods, def_method, def_methods, Transaction, TModule, Provided
 from transactron.utils import assign
 from transactron.utils.data_repr import bits_from_int
 from transactron.utils.dependencies import DependencyContext
@@ -22,6 +24,18 @@ from coreblocks.interface.keys import (
     ExceptionReportKey,
     AsyncInterruptInsertSignalKey,
 )
+
+__all__ = [
+    "CSRUnit",
+    "CSRBlockComponent",
+    "RegisteredCSRProtocol",
+]
+
+
+class RegisteredCSRProtocol(Protocol):
+    _fu_read: Provided[Method]
+    _fu_write: Provided[Method]
+    _fu_access_valid: Provided[Method]
 
 
 def csr_access_privilege(csr_addr: int) -> tuple[PrivilegeLevel, bool]:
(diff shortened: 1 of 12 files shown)


awariac:
reminder that `git rebase HEAD~<> --onto=master` is a good trick :)

awariac:
weird that local run of lint check_types doesn't report these obvious errors

awariac:
doing docstrings so i can fall asleep

coreblocks/func_blocks/csr/csr_unit.py
```diff
@@ -80,17 +85,31 @@ def __init__(self, gen_params: GenParams):
...
-        for csr in self.dependency_manager.get_dependency(CSRListKey()):
-            assert csr.csr_number is not None
-            if csr.csr_number in self.regfile:
-                raise RuntimeError(f"CSR number {csr.csr_number} already registered")
-            self.regfile[csr.csr_number] = (csr._fu_read, csr._fu_write, csr.access_valid)
+        # Fills `self.regfile` with CSR registers provided by `CSRListKey` dependency.
+        for csr_number, csr in self.dependency_manager.get_dependency(CSRListKey()):
+            if csr_number in self.regfile:
+                raise RuntimeError(
+                    f"CSR number {csr_number} already registered (at {self.regfile[csr_number].src_loc} {csr.src_loc}"
```
qbojj:
```suggestion
                    f"CSR number {csr_number} already registered (at {self.regfile[csr_number].src_loc}) {csr.src_loc}"
```

qbojj (approved):

qbojj:
It may also be a good idea to separate logical CSRs (on RV32 some registers have two numbers) from CSR numbers and allowing read/read_comb/write to have the width of actual register. Not sure if it should be in this PR though

coreblocks/func_blocks/csr/csr_unit.py
```diff
@@ -80,17 +85,31 @@ def __init__(self, gen_params: GenParams):
...
+                )
+            self.regfile[csr_number] = csr
+
+    def _csr_access_privilege(self, csr_addr: int) -> tuple[PrivilegeLevel, bool]:
```
tilk: Why was it moved inside the class? It's used only here, but it's not a real method, the `self` argument is not used. Annotate as static or move back.

coreblocks/func_blocks/csr/csr_unit.py
```diff
@@ -23,19 +26,21 @@
...
 )
 
+__all__ = [
+    "CSRUnit",
+    "CSRBlockComponent",
+    "RegisteredCSRProtocol",
+]
+
+
+class RegisteredCSRProtocol(Protocol):
```
tilk: How about moving the protocol to a separate file? This typically avoids dependency problems in the long run.

tilk (approved):
Left minor comments, please consider before merging.

tilk:
> It may also be a good idea to separate logical CSRs (on RV32 some registers have two numbers) from CSR numbers and allowing read/read_comb/write to have the width of actual register. Not sure if it should be in this PR though

The `DoubleCounterCSR` thingy looks like a hack, I believe this should be a 64-bit CSR regardless of mxlen, with two windows into it on RV32. Not only counters are 64-bit CSRs in RV32 - there is e.g. `mcyclecfg` of Smcntrpmf.

awariac:
> It may also be a good idea to separate logical CSRs (on RV32 some registers have two numbers) from CSR numbers and allowing read/read_comb/write to have the width of actual register. Not sure if it should be in this PR though

as discussed with qbojj, removed default width argument from CSR layouts, as it creates more confusion. Virtual CSRs already implement this behaviour

coreblocks/func_blocks/csr/csr_unit.py
awariac: oops, meant to put `and` here, corrected

PR #910, opened 2026-04-20
fetch: more explicit access fault handling, test improvements

### LINKED ISSUE #901: potential core hang - `access_fault` handling in fetch
I'm now worried about `access_fault` handling:
What if cache response pattern is that `instr_start` turns out 0?
Then due to serializer no instruction would be passed out of fetch, fetch would be stalled, but because of that no exception would be risen, because there is no information. Core hang-ups.
Anyway thats for separate PR.

I will try to look more closely into it in next days, leaving issue for now.

_Originally posted by @piotro888 in https://github.com/kuznia-rdzeni/coreblocks/pull/881#discussion_r3096934590_

piotro888:
When investigating #901 I found following things:
1 - fetch access fault test never produces `ACCESS_FAULT_ON_SECOND_HALF` case with 
ZCA

2 - issue claim was wrong, core will never hang because valid instr mask generation doesn't in fact check instruction validity, but only produces positions and will be always non-zero.
I feel this claim is very fragile and it's better to do it explicitly.

There may be a bug instead that will report wrong pc address on access fault, with prepared input data and RVC, however tests didn't catch that.
I think it is possible for this one `(fetch_block_offset <= i) & (~instr_start[i - 1] | is_rvc[i - 1])` to be false for first instruction (`i == fetch_block_offset`) if previous instruction was determined as 2-aligned start and non-rvc on random data. That's another reason for change, but I'm not entirely sure if this can happen now.

Anyway I think depending on that much implicit logic is not great, and may break in future changes.
I made the vector for access fault explicit to emit only one instruction with requested pc, that will have access fault set. 

TLDR
core will never hang, because instr_start is always non-zero.
however, it can possibly have wrong address if first requested one is determined invalid. PR fixes that and makes it explicit.

Closes #901

### DIFF
--- a/coreblocks/frontend/fetch/fetch.py
+++ b/coreblocks/frontend/fetch/fetch.py
@@ -212,7 +212,7 @@ class FetchUnit(Elaboratable):
                 else:
                     m.d.av_comb += expanded_instr[i].eq(cache_resp.fetch_block[i * 32 : (i + 1) * 32])
 
-            # Mask denoting at which offsets an instruction starts
+            # Mask denoting at which offsets expected instructions start (depends on rvc indication and start address)
             instr_start = [Signal() for _ in range(fetch_width)]
             for i in range(fetch_width):
                 if Extension.ZCA in self.gen_params.isa.extensions:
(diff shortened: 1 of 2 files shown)


tilk:
> There may be a bug

Is it there, or is it not? If it's not, leaving a comment documenting the behavior is better than increasing the complexity of the hardware.

qbojj:
>  I think it is possible for this one (fetch_block_offset <= i) & (~instr_start[i - 1] | is_rvc[i - 1]) to be false for first instruction (i == fetch_block_offset) if previous instruction was determined as 2-aligned start and non-rvc on random data. That's another reason for change, but I'm not entirely sure if this can happen now.

As every term of `instr_start` has `fetch_block_offset <= i`, when `fetch_block_offset == i` - `instr_start[i - 1]` is trivially 0. As such we always have `instr_start[fetch_block_offset] == 1`.

awariac:
>  As such we always have instr_start[fetch_block_offset] == 1.

ah right. this should be documented.

I think about one other thing:
It is worth inserting extra useless instructions to the core that would need to be rolled back (solution on master)
or
adding one 8-or-something-bit-wide Mux to always insert one instruction at right position (on this PR)?

qbojj (approved):

PR #912, opened 2026-04-22
Generalized instruction metric

### LINKED ISSUE #890: Generalize and refactor FU instruction count metrics

tilk:
Adds superscalar-ready, generalized instruction count metrics.

Accuracy is not yet solved - the module counts all issued instructions, including rollbacked ones.

TODO:
* [x] Fix tests.

Closes #890. Required for #889.

### DIFF
--- a/coreblocks/func_blocks/fu/jumpbranch.py
+++ b/coreblocks/func_blocks/fu/jumpbranch.py
@@ -230,11 +230,9 @@ class JumpBranchFuncUnit(FuncUnitBase[JumpBranchFn]):
                 exception=exception,
             )
 
-        @def_method(m, self.issue)
+        @def_method(m, self.issue_decoded)
         def _(arg):
-            m.d.top_comb += self.decoder.exec_fn.eq(arg.exec_fn)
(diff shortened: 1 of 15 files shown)


coreblocks/func_blocks/fu/jumpbranch.py
```diff
@@ -230,11 +224,9 @@ def elaborate(self, platform):
                 exception=exception,
             )
 
-        @def_method(m, self.issue)
+        @def_method(m, self.issue_decoded)
```
awariac, line 227: i would prefer the `issue` name to stay with extra field, but i see this is required for naming conflict

piotro888 (approved):

coreblocks/func_blocks/fu/jumpbranch.py
tilk: Indeed, this is a slight downside of this design, which I otherwise believe to be pretty elegant.

PR #915, opened 2026-04-24
Superscalar flush

tilk:
Another step towards superscalar retirement: superscalar instruction flushing.

The flush rate is currently limited because of CRAT limitation.

TODO:
- [x] Add done count to ROB test
- [x] Fix tests 

Based on #914.


coreblocks/core.py
```diff
@@ -63,15 +63,15 @@ def __init__(self, *, gen_params: GenParams):
 
         self.rf_allocator = PriorityEncoderAllocator(
             gen_params.phys_regs, gen_params.frontend_superscalarity, init=2**gen_params.phys_regs - 2
```
qbojj: `gen_params.frontend_superscalarity` should probably be `max(gen_params.frontend_superscalarity, gen_params.retirement_superscalarity)` - as it is later `frontend_superscalarity` items from `alloc` and `retirement_superscalarity` from `free` are used.

qbojj (changes requested):

coreblocks/core.py
tilk: Good catch.

coreblocks/backend/retirement.py
```diff
@@ -228,15 +238,26 @@ def set_trap_csrs(cause_reg, epc_reg, tval_reg):
...
-                    self.rob_retire(m, count=1)
 
-                    with m.If(rob_entry.rob_data.tag_increment):
+                    count = Signal(range(self.gen_params.retirement_superscalarity + 1))
+                    tag_incr_mask = Signal(self.gen_params.retirement_superscalarity)
+                    sel_mask = Signal.like(tag_incr_mask)
+                    # CRAT can currently deallocate at most one tag per cycle, this logic reduces the flush rate
+                    # TODO: improve
+                    m.d.av_comb += tag_incr_mask.eq(Cat(entry.rob_data.tag_increment for entry in rob_entries.entries))
+                    m.d.av_comb += sel_mask.eq(tag_incr_mask & (tag_incr_mask - 1) | (-1 << rob_entries.done_count))
```
awariac: why `-1 <<`  instead of `1 <<` if we only `ctz`? That would make one factor easier to follow.
Also `sel_mask` could be maybe `first_unsafe_to_retire_mask` or something more descriptive like that.

awariac (approved):
looks good

qbojj (approved):

coreblocks/backend/retirement.py
tilk: Because it generates smaller circuits, both in FPGA and gate synthesis.

PR #916, opened 2026-04-24
make PMP MPRV aware

### LINKED ISSUE #888: PMP is not MPRV aware
PMP checks for explicit memory operations should be made with current privilege level or if `MPRV` bit of `mstatus` is 1, than with privilege of `MPP` bits of `mstatus`.

The best way would be to just give PMP priviledge level as an argument, as MMU will always have S-mode PMP, instructions always equal to `priv_mode` CSR and LSU depends.

qbojj:
Closes #888

### DIFF
--- a/coreblocks/priv/pmp.py
+++ b/coreblocks/priv/pmp.py
@@ -1,5 +1,6 @@
 from amaranth import *
 from amaranth.lib import data
+from amaranth.lib.enum import Enum, auto, unique
 from amaranth_types import HasElaborate
 from transactron.core import TModule
 from transactron.utils import DependencyContext, assign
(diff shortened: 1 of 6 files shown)


tilk (approved):

coreblocks/priv/pmp.py
```diff
@@ -14,14 +15,26 @@ def __init__(self):
...
 class PMPChecker(Elaboratable):
     """
     Implementation of physical memory protection checker.
     This is a combinational circuit with return value read from `result` output.
 
-    M-mode accesses bypass PMP by default (result = 1/1/1) unless a matching
-    locked entry (L=1) is found. S/U-mode accesses default to no access (0/0/0)
-    if no entry matches.
+    Effective mode depends on `mode`:
+    - LSU: MPRV-aware, using MPP when (MPRV=1 and current mode is M)
```
awariac: did you mean:
```suggestion
    - LSU: MPRV-aware (using MPP when MPRV=1 and current mode is M)
```
?

coreblocks/priv/vmem/translation.py
```diff
@@ -117,9 +117,7 @@ def _(addr: Value):
                     mprv = csr.m_mode.mstatus_mprv.read(m).data
                     mpp = csr.m_mode.mstatus_mpp.read(m).data
 
-                    m.d.av_comb += effective_priv_mode.eq(
-                        (Mux(mprv & (priv_mode == PrivilegeLevel.MACHINE), mpp, priv_mode))
-                    )
+                    m.d.av_comb += effective_priv_mode.eq(Mux(mprv, mpp, priv_mode))
```
awariac, line 120: i see that `mprv & (priv_mode == PrivilegeLevel.MACHINE)` is not explicitly stated by spec, and if i understand correctly mprv & priv_mode != M cannot happen in reality, so it doesn't need to be checked.

Can we add an hardware assertion to `~mprv | priv_mode == Machine` to `csr_registers.py` to catch possible implementation edge-cases?

awariac (approved):

coreblocks/priv/vmem/translation.py
qbojj: Yes `mprv` implies `priv_mode == MACHINE`

tilk (approved):

PR #921, opened 2026-04-29
Enable privileged ISA tests in riscof

qbojj:


tilk:
I see there was a lot of experimentation here, but you gave no explanation for the final results. Why are the final ISA strings like this? Any idea why Zicond and Zifencei fails?

awariac (approved):

qbojj:
Zicond seems to fail because version of riscof / some other part of the CI didn't support that extension.
As for Zifencei it looks like the coreblocks test fails with an exception that there is an invalid access - it needs to have executable section made writeable

qbojj:
> The RISCOF repository is now deprecated, please use ACT 4.0(https://github.com/riscv/riscv-arch-test) instead

BTW It looks like riscof has been deprecated in april

tilk:
> > The RISCOF repository is now deprecated, please use ACT 4.0(https://github.com/riscv/riscv-arch-test) instead
> 
> BTW It looks like riscof has been deprecated in april

Great, another toolchain change to do. @qbojj  Could you make an issue out of that?

tilk:
> Zicond seems to fail because version of riscof / some other part of the CI didn't support that extension. As for Zifencei it looks like the coreblocks test fails with an exception that there is an invalid access - it needs to have executable section made writeable

Do you wish to merge this PR now and leave fixing this to another PR?

qbojj:
Yes

PR #923, opened 2026-04-29
Fix WishboneMuxer select signal

### LINKED ISSUE #919: Linux regression again
Linux doesn't boot _again_ on `master` (more precisely I checked at https://github.com/kuznia-rdzeni/coreblocks/commit/5a9d8f5165c0da5dbc9315b21c19704f9eba228e , so it doesn't include S-mode changes yet) :c

I will try to bisect it later, more detail to follow (but it's quite painful due to synthesis times. I need to set-up Linux regression testing on FPGA SOONer! (pls remind me of that from time to time)

Core hangs on later kernel boot process (iirc after first `io scheduler` log).

awariac:
image
See Wishbone spec.

TAG is allowed to change in a block read (no de-assertion of STB) and holds its value, so extra logic here was not needed and caused a bug.

Fixes #919


awariac:
Linux bug appeared as a very specific case, that there were two `request`s without cycle break that were acessing different memory regions. It is a valid behaviour, but Muxer was too restrictive without need.

tilk (approved):

qbojj (approved):

PR #928, opened 2026-05-01
Add config file option to gen_verilog.py

### LINKED ISSUE #757: Accept custom `CoreConfiguration` in `gen_verilog` script
(from meeting?) It would be great to support reading custom CoreConfiguration derived configurations from external input file in `gen_verilog` script, so Coreblocks source wouldn't need to be modified to customize core.

jseaton:


scripts/gen_verilog.py
tilk: Remove this. Split `coreblocks.params.configurations` into two files, `coreblocks.params.core_configuration` with the `CoreConfiguration` class and related definitions, and `coreblocks.params.configurations`, containing just the configurations. Rename configurations to their short names (`basic` instead of `basic_core_config`, etc.). In modules where the configurations are used, use `from coreblocks.params import configurations` and replace `basic_core_config` with `configurations.basic`, etc.

scripts/gen_verilog.py
```diff
@@ -100,10 +109,19 @@ def main():
...
-        raise KeyError(f"Unknown config '{args.config}'")
+    if args.configfile:
+        configfile = SourceFileLoader("custom_config", args.configfile).load_module()
+
+        if args.config not in dir(configfile):
+            raise KeyError(f"Unknown config '{args.config}'")
+
+        config = getattr(configfile, args.config)
+    else:
+        if args.config not in str_to_coreconfig:
```
tilk: Instead of using `str_to_coreconfig` here, use the `configurations` module instead of the module loaded using the `SourceFileLoader`. Keep the rest of the logic (the existence check and looking up the config from the module) common.

tilk (changes requested):
Thank you for your contribution. It's a great opportunity to also do a minor refactor, to keep the configuration selection logic common between built-in and external configurations. Remember to fix formatting using `scripts/lint.sh format`.

scripts/gen_verilog.py
```diff
@@ -72,6 +73,14 @@ def main():
         + f"Available configurations: {', '.join(list(str_to_coreconfig.keys()))}. Default: %(default)s",
     )
 
+    parser.add_argument(
+        "-f",
+        "--configfile",
+        action="store",
+        default=None,
+        help="Select custom config file for core configuration, -c must be a variable name"
```
awariac: Please mention what the file should contain - `CoreConfiguration`s under appropriate variable names.

That ideally would be in some docs page, but i see we don't have one :p

scripts/gen_verilog.py
```diff
@@ -100,10 +109,19 @@ def main():
...
 
-    if args.config not in str_to_coreconfig:
-        raise KeyError(f"Unknown config '{args.config}'")
+    if args.configfile:
+        configfile = SourceFileLoader("custom_config", args.configfile).load_module()
+
+        if args.config not in dir(configfile):
+            raise KeyError(f"Unknown config '{args.config}'")
+
+        config = getattr(configfile, args.config)
```
awariac: add an assertion here that we expect instanceof `CoreConfiguration`, to raise an early error in case of some mistake.

awariac (commented):
Hi, thanks for you contribution! I think I meant something like that :3

~~you can run `./scripts/lint.sh format` to auto-fromat code before push~~ nvm, i missed the previous review

tilk:
Tip: you can use `scripts/lint.sh verify` to use the typechecker to help with the refactor.

jseaton:
Sorry shouldn't have pushed that 🫠 should be good noww

tilk (approved):
Thank you for your contribution.

awariac (approved):
meow!


<sub>nice refactoring</sub>

qbojj (approved):

PR #934, opened 2026-05-01
Fix tag invalidation bug

tilk:
I was looking into CRAT and found this. The value `tag + 1` does not have `tag_bits` size, it has `tag_bits + 1`, and it does not overflow. Similar with `tags_tail - 1`. This makes the comparison and the `cyclic_mask` calculation invalid.

### DIFF
--- a/coreblocks/core_structs/crat.py
+++ b/coreblocks/core_structs/crat.py
@@ -353,16 +353,21 @@ class CheckpointRAT(Elaboratable):
 
             # Invalidate tags on wrong speculaton path (suffix), but don't free them for instruction validity tracking
             tag_plus_1 = Signal(self.gen_params.tag_bits)
-            latest_tag = Signal(self.gen_params.tag_bits)
+            alloc_tags_bound = Signal(self.gen_params.tag_bits)
             m.d.av_comb += tag_plus_1.eq(tag + 1)
-            m.d.av_comb += latest_tag.eq(tags_tail - 1)
+            m.d.av_comb += alloc_tags_bound.eq(tags_tail - 1)
             with m.If(tag_plus_1 == tags_tail):
                 log.debug(m, True, "rollback to 0x{:x}. no tags to invalidate", tag)
             with m.Else():
-                invalidate_mask = cyclic_mask(2**self.gen_params.tag_bits, tag_plus_1, latest_tag)
+                invalidate_mask = cyclic_mask(2**self.gen_params.tag_bits, tag_plus_1, alloc_tags_bound)
                 m.d.comb += active_tags_reset_mask_0.eq(invalidate_mask)
                 log.debug(
-                    m, True, "rollback to 0x{:x}. invalidate tags from 0x{:x} to 0x{:x}", tag, tag_plus_1, latest_tag
+                    m,
+                    True,
+                    "rollback to 0x{:x}. invalidate tags from 0x{:x} to 0x{:x}",
+                    tag,
+                    tag_plus_1,
+                    alloc_tags_bound,
                 )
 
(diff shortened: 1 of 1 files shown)


awariac:
Maybe truncation could be done inside `cyclic_mask`? It seems like an often use case to +/- from value there

tilk:
> Maybe truncation could be done inside `cyclic_mask`? It seems like an often use case to +/- from value there

Looked at it, and I don't have a good idea how that could be done in the general case. At first sight, this seems to require calculating modulo, which is notoriously hard in hardware.

awariac:
sure, in this case we can add assertions if value is greater than max or less than zero at least for simulation

coreblocks/core_structs/crat.py
```diff
@@ -352,13 +352,17 @@ def _(tag: Value):
             m.d.sync += rollback_tag_s1.eq(tag)
 
             # Invalidate tags on wrong speculaton path (suffix), but don't free them for instruction validity tracking
-            with m.If((tag + 1 == tags_tail)):
+            tag_plus_1 = Signal(self.gen_params.tag_bits)
+            latest_tag = Signal(self.gen_params.tag_bits)
+            m.d.av_comb += tag_plus_1.eq(tag + 1)
+            m.d.av_comb += latest_tag.eq(tags_tail - 1)
```
awariac: the names are not actually true.
`tags_tail - 1` is just a way to select (mask) all possible tags forward (until the oldest allocated tag).
I'm not sure what the best name would be, maybe `all_future_tags`?
`tag_plus_1` could be `rollback_tag_plus_1`

awariac (approved):
naming change request in comment, otherwise ok

i think bound assertion in transactron would be very helpful for future.

I only wonder why it wasn't catched by randomized tests, I remember I run on a multiples of more iterations on local machine and got no issues, iirc they did seem rather solid.

awariac:
the problem with assertion, even the Amaranth one (for sim only), is that it would require additional `m` argument.
Do you think it is worth it? It has a potential of saving a lot of debug time,

coreblocks/core_structs/crat.py
tilk: `tag_plus_1` is literally the value of `tag` plus 1. Adding `rollback` is just verbosity for no reason.

I agree that I named `latest_tag` wrong. But `all_future_tags` seems weird, this looks like a set name, but this is no set, this is a tag number. But I'm confused now. `tags_tail` is the oldest allocated tag. `tags_tail - 1` is... not a tag number? Why is it used as a range bound, both in `cyclic_mask` and the log text? This looks confusing and is not explained in any comment.

Generally I believe that C-RAT was prematurely merged, its logic feels pretty involved and I don't understand it fully. Which is a sign there may be more monsters lurking here.

tilk:
> I only wonder why it wasn't catched by randomized tests, I remember I run on a multiples of more iterations on local machine and got no issues, iirc they did seem rather solid.

I had similar doubts. I think this should be investigated, a test which detects this bug could in principle detect other bugs too.

> i think bound assertion in transactron would be very helpful for future.

The module requirement makes this unwieldy, unfortunately. But I agree, this is not the first place I wish for a simple assertion check in library code.

Anyway, this function (and many others) is missing a docstring, explaining its behavior and input assumptions.

coreblocks/core_structs/crat.py
tilk: > Why is it used as a range bound

Ah, is it supposed to be the bound of the range of (as you said) all tags in the future of `tag`, including (for some reason) the ones that were not allocated yet?

PR #917, opened 2026-04-26
Implement MMU

qbojj:
Based on #884 and #916

Implement MMU featuring:
- [x] L1D and L1I TLB storing entries available for same-cycle translation - fully associative
- [x] L2 TLB - set-associative
- [x] PTW - implementing `Svade` samantics (`Svadu` would require LSU atomics redesign)

PR #942, opened 2026-05-08
[MMU 1/3] Implement L1 TLB - fully associative page table entry cache

qbojj:
Part 1 of 3 of MMU implementation (#917)

Implements fully associative L1 TLB and logic to drive it from address translator.
This also establishes the page translation interface, `SFENCE.VMA` handling, randomized TLB testing.

Some notable features:
- L1 TLB can operate in single-cycle - same as the address translation itself
- compatible with Svade and Svadu semantics.
- easy addition of Svnapot extension.


coreblocks/interface/layouts.py
```diff
@@ -169,18 +169,56 @@ def __init__(self, gen_params: GenParams):
...
+            ("x", 1),
+            ("u", 1),
+            ("d", 1),
+            ("g", 1),
+        )
+        self.size_class = ("size_class", gen_params.vmem_params.tlb_size_class_bits)
 
+        self.request = make_layout(
+            fields.addr,
+            ("write_aspect", 1),
```
awariac: maybe rename to something like `is_store` or otherwise document.

coreblocks/params/core_configuration.py
```diff
@@ -119,6 +120,8 @@ class _CoreConfigurationDataClass:
         SV32 enabled, 32 for RV32 with only BARE mode and 56 for RV64.
     hpm_counters_count: int
         Number of implemented HPM counters (mhpmcounter3..mhpmcounter31).
+    tlb_config: TLBCacheConfiguration
+        Grouped TLB configuration for L1I, L1D, and shared L2 - only applicable when SV* modes are supported.
```
awariac: Is shared L2 in scope of current coreblocks and TLB?

coreblocks/priv/vmem/tlb.py
```diff
@@ -0,0 +1,243 @@
...
+            found_entry = Signal(TLBEntry(self.gen_params))
+
+            for way in range(self.entries):
+                with m.If(cam.full_match[way]):
+                    m.d.av_comb += found_entry.eq(cam.ways_data[way])
+
+            ask_backing = Signal()
+            with m.If(~cam.full_match.any()):
+                m.d.av_comb += ask_backing.eq(1)
+            with m.Elif(write_aspect & ~found_entry.permissions.d):
```
awariac: what about PTE A bit with Svade?

coreblocks/priv/vmem/tlb.py
```diff
@@ -0,0 +1,243 @@
...
+                # We have found an entry, but it doesn't have the dirty bit set
+                # NOTE: currently we would never actually need to ask the backing device,
+                #   as we only implement Svade, so if this happens, we are almost sure
+                #   the re-walk will not fix the issue.
+                m.d.av_comb += ask_backing.eq(1)
```
awariac: This is step 9. for Svadu, right? if so, the comment is a bit confusing.

If we have Svade we should not re-access the memory, but "stop and raise a page-fault exception corresponding to the original access type".
We can modify the entry in memory manually, but is it required to do `sfence.vma` afterwards, or it is not required and this refresh attempt is somehow needed?

Or is it because we don't have fault handling here? Shouldn't we short-circuit page faults from here (in case of Svade), instead of relying on backing module?

sorry for so many questions, i'm a bit confused with the spec and logic

coreblocks/priv/vmem/tlb.py
```diff
@@ -0,0 +1,243 @@
...
+            m.d.comb += cam.checked_asid.eq(self.sfence_vma.data_in.asid)
+            m.d.comb += cam.checked_vpn.eq(self.sfence_vma.data_in.vaddr >> PAGE_SIZE_LOG)
+        with m.Elif(request_in_flight):
+            m.d.comb += cam.checked_asid.eq(current_asid)
+            m.d.comb += cam.checked_vpn.eq(requested_vpn)
+        with m.Else():
+            m.d.comb += cam.checked_asid.eq(current_asid)
+            m.d.comb += cam.checked_vpn.eq(self.request.data_in.vpn)
+
+        @def_method(m, self.request, ready=~request_in_flight)
```
awariac, line 206: if it doesn't cause some comb loop issues (schdule_before may be used) it could save one cycle by `~request_in_flight | refill_transaction.run`

coreblocks/priv/vmem/tlb.py
```diff
@@ -0,0 +1,243 @@
...
+                        match_bits = bits_per_level * sz_class
+                        m.d.comb += self.addr_match[way].eq(
+                            self.checked_vpn[match_bits:] == self.ways_data[way].vpn[match_bits:]
+                        )
+
+        m.d.comb += self.full_match.eq(self.valid_match & self.addr_match & (self.asid_match | self.global_match))
+
+        # Replacement candidate is the first invalid entry, or the round-robin entry if all are valid
+        # The order of checks: currently present entry -> invalid entry -> round-robin.
+        m.d.comb += self.replace_candidate.eq(self.replacement_data)
```
awariac: `_data` may be not the best name for index

coreblocks/priv/vmem/tlb.py
```diff
@@ -0,0 +1,243 @@
...
+                "size_class": gen_params.vmem_params.tlb_size_class_bits,
+                "vpn": gen_params.vmem_params.max_tlb_vpn_bits,
+                "ppn": gen_params.phys_addr_bits - PAGE_SIZE_LOG,
+                "permissions": gen_params.get(AddressTranslationLayouts).permissions,
+            }
+        )
+
+
+class TLBCAM(Elaboratable):
+    """A single set of a TLB CAM, containing multiple ways."""
```
awariac: Document external interface signals.
This is not what you would directly expect as a CAM

awariac (commented):
nice work

coreblocks/params/core_configuration.py
qbojj: The next patch includes L2 TLB infrastructure (set-associative TLB).
I think should be L1I TLB, L1D TLB and L2 TLB

coreblocks/priv/vmem/tlb.py
qbojj: Only entries with A bit set can get inside of the TLB infrastructure regardless of the Svad* semantics.
As the spec allows for speculative setting of A bit - this is allowed

coreblocks/priv/vmem/tlb.py
qbojj: I made it so to make defining other replacement policies easier - it would need only changing the TLBCAM definition.
EDIT: now that I think about it - it would not be that simple

coreblocks/priv/vmem/tlb.py
qbojj: All manual changes to page table structure require `sfence.vma`. For exception based A/D bit manipulation (non Svade) semantics we actually don't need this check and could just return the non-dirty entry and handle it in address translation - making `write_aspect` not used in the TLB infrastructure.
I just made it so to not have multiple different implementations for multiple semantics.

coreblocks/priv/vmem/tlb.py
qbojj: the `request` and `refill_transaction` cannot run at the same time, as both set the `checked_vpn` signal of the `TLBCAM`. The request to get the correct entry and the refill to overwrite potentially present stale entry

coreblocks/priv/vmem/tlb.py
```diff
@@ -0,0 +1,266 @@
...
+                    m.d.av_comb += found_entry.eq(cam.ways_data[way])
+
+            ask_backing = Signal()
+            with m.If(~cam.full_match.any()):
+                m.d.av_comb += ask_backing.eq(1)
+            if self.gen_params.vmem_params.supports_svade:
+                with m.Elif(is_store & ~found_entry.permissions.d):
+                    # We have found an entry, but it doesn't have the dirty bit set
+                    # Ask the backing resolver to set it (Svade semantic)
+                    m.d.av_comb += ask_backing.eq(1)
```
awariac, line 224: Wait, I still don't follow.
Thats 'Svadu' behaviour to set A/D bits.
" When the Svade extension is **not** implemented, the following scheme applies.
[...]
When the virtual page is written and the D bit is clear, the PTE is updated to set the D bit.
"
"If the Svade extension is implemented, **stop** and raise a page-fault exception corresponding to the
original access type'

qbojj: `Svadu` is an extension allowing software control of presense of `Svade` extension, so:
- no `Svade`, no `Svadu` - always exceptions on invalid A/D
- `Svade`, no `Svadu` - always automatic A/D management
- `Svadu` - either automatic or exceptions on invalid A/D - depending on `menvcfg.ADUE`

awariac: oh thanks for clarifying, but isn't it the other way around for implemented/not implemented `Svade`?

qbojj: yes

awariac (approved):
thanks for splitting it in parts

coreblocks/arch/isa_consts.py
```diff
@@ -254,6 +259,23 @@ def mode_dependencies(cls, mode: "SatpMode") -> Set["SatpMode"]:
             case _:
                 raise ValueError(f"Unsupported SATP mode: {mode}")
 
+    @classmethod
+    def level_count(cls, mode: "SatpMode") -> int:
```
tilk: Why is this (and the previous one) a class method, and not a normal method?

coreblocks/params/vmem_params.py
```diff
@@ -14,6 +29,14 @@ def max_physical_address_bits(xlen: int, supported_schemes: Collection[SatpMode]
         satp_layout = SatpLayout(xlen)
         return satp_layout["ppn"].width + PAGE_SIZE_LOG
 
+    @staticmethod
+    def vpn_bits_for_mode(xlen: int, mode: SatpMode) -> int:
+        return SatpMode.bits_per_page_table_level(xlen) * SatpMode.level_count(mode)
```
tilk: This looks like it could go to `SatpMode`. Add docstring.

coreblocks/priv/vmem/tlb.py
```diff
@@ -0,0 +1,266 @@
...
+    valid_match: Out, bit-vector of valid entries.
+    addr_match: Out, bit-vector of ways matching the address (taking into account size class).
+    asid_match: Out, bit-vector of ways matching the ASID (only non-global entries).
+    global_match: Out, bit-vector of global entries.
+
+    full_match: Out, bit-vector of ways matching inputs - valid and address and (ASID or global).
+
+    replacement_rr_index: In, index for round-robin replacement.
+    replace_candidate: Out, index of the way to replace on a miss.
+    next_replacement_rr_index: Out, value of replacement_rr_index after replacement.
```
tilk: Doesn't match neither the old style (numpy-style) nor the new style (docstrings on definitions) of attr docs.

coreblocks/priv/vmem/tlb.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+class TLBEntry(StructLayout):
+    valid: Value
+    asid: Value
+    vpn: Value
+    ppn: Value
+    size_class: Value
+    permissions: Value
+
```
tilk: Huh. Does this make any sense? The fields are present on `View` instances, not `StructLayout` instances.

coreblocks/priv/vmem/iface.py
```diff
@@ -0,0 +1,15 @@
...
+class TLBBackingDevice(Protocol):
+    """Protocol for devices that can be used as a backing device for TLBs."""
+
+    request: Provided[Method]
+    """Method for requesting a translation.
+    Expected layout is `AddressTranslationLayouts.tlb_request`."""
+
+    accept: Provided[Method]
+    """Method for accepting a translation response.
+    Expected layout is `AddressTranslationLayouts.tlb_accept`."""
```
tilk, line 15: Does not require immediate action, but, a minor remark:

Due to our previous experience with refactoring and testing, I would like to minimize passing methods as arguments to constructors, and use `provide` instead. But I am aware this introduces some syntactic overhead when multiple related methods are to be passed. I believe we ultimately need some mechanism not unlike `amaranth.lib.wiring` in Transactron.

tilk (commented):
Will need to have a second look at the logic, but I see nothing weird right now. I'm leaving some coding style remarks.

coreblocks/arch/isa_consts.py
qbojj: didn't want to write `SatpMode`...

test/priv/vmem/test_tlb.py
```diff
@@ -0,0 +1,393 @@
...
+        test_cases.extend((random.randint(0, 0xFFFFF), random.randint(0, 0xF)) for _ in range(64))
+        random.shuffle(test_cases)
+
+        for vpn, asid in test_cases:
+            sim.set(self.csr_instances.s_mode.satp_asid, asid)
+            await sim.tick()
+
+            await self.dut.request.call(sim, vpn=vpn, is_store=0)
+            response = await self.dut.accept.call(sim)
+            assert self.matches_lookup(response, vpn, asid)
```
tilk, line 377: The TLB module seems to be capable of pipelining, and even of single-cycle operation, thanks to the use of `Forwarder` inside. Yet this is not tested anywhere.

qbojj: This test set is designed to be agnostic to the implementation. I will write a specialised test for same-cycle operation

tilk (approved):

PR #943, opened 2026-05-08
`gen_verilog` script updates

### LINKED ISSUE #941: proposal: move gen_verilog inside `coreblocks` and pipx
It is the only script in `scripts` that is directly related to project usage.
This way it can be registered as command in `pyproject.toml` so it simpler to run it externaly, or possible to invoke  it directly (with creating other files) with ex. `pipx`.

As we tightly depend on some dependencies that may conflict i believe running in isolated environment on pipx should be the recommended simple option for verilog generation. It would be also easier to integrate into Litex.
In this case we could also add `[project.entry-points."pipx.run"]` to specify default entrypoint to gen_verilog for pipx invocations

awariac:
Closes #941 (see discussion)
this is needed to be able to generate verilog from packaged `coreblocks`, `scripts` directory is not packaged.
`gen_verilog.py` was also the only script directly related to core usage in `scripts`.

Rationale for including it at root tree and providing it as main `coreblocks` script, is that this script is really the only thing the end-user is interested in from our project. After #928 everything can be done with this script.

When script is named `coreblocks`, then it would be run as default when using project with dependency isolating single run commands like `pipx run` or `uvx`, that in my opinion should be the default usage for our project (for less-advanced use without hacking on internals).

Prerequisite for Litex packaging simplifications (and wold be nice for physical FPGA CI i'm working on)

Bonus: added nice version string to gen_verilog, because it takes long time to finish and could be confusing as installed script.


awariac:
@tilk is this some issue with metrics dependency contexts / imports? This only happens on `--strip-debug`, i'm not sure where to look at

tilk (approved):

PR #945, opened 2026-05-10
correctly handle writeable fields of mip and sip

### LINKED ISSUE #938: mip/sip writes are broken
custom `mip` logic misses CSR written software interrupts (or any other else) and needs rewrite. They are marked by as-level triggered, but in reality in case of a CSR write they should stay forever. (don't trust me with this and read spec) - outside level triggered interrupts need to be somehow integrated with manual writes.

qbojj:
handle mip/sip writes correctly.

Closes #938

### DIFF
--- a/coreblocks/priv/traps/interrupt_controller.py
+++ b/coreblocks/priv/traps/interrupt_controller.py
@@ -19,6 +19,7 @@ from transactron.core import Method, TModule, def_method
 from transactron.core.transaction import Transaction
 from transactron.utils import logging
 from transactron.utils.dependencies import DependencyContext
+from transactron.lib.transformers import MethodMap
 
 log = logging.HardwareLogger("core.interrupt_controller")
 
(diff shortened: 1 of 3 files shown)


coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -76,27 +77,52 @@ def __init__(self, gen_params: GenParams):
...
+            assert self.sip_writeable & ~self.mip_writeable == 0
 
-        self.mie = CSRRegister(CSRAddress.MIE, gen_params, ro_bits=~mie_writeable)
-        self.mip = CSRRegister(CSRAddress.MIP, gen_params, fu_write_priority=False, ro_bits=~self.edge_reported_mask)
+        self.mie = CSRRegister(CSRAddress.MIE, gen_params, ro_bits=~self.mie_writeable)
+        self.mip = CSRRegister(CSRAddress.MIP, gen_params, fu_write_priority=False, ro_bits=~self.mip_writeable)
 
         if gen_params.supervisor_mode:
-            self.mideleg = CSRRegister(CSRAddress.MIDELEG, gen_params, ro_bits=~smode_interrupts)
+            self.mideleg = CSRRegister(CSRAddress.MIDELEG, gen_params, ro_bits=~self.sie_writeable)
```
awariac: fun fact: there is no rule restricting delegation of M-mode interrupts.

There is even a note (weird that they chose to provide non exact Figures):
```
Bits 3, 7, and 11 of sip and sie correspond to the machine-mode software, timer, and external
interrupts, respectively. Since most platforms will choose not to make these interrupts
delegatable from M-mode to S-mode, they are shown as 0 in Figure 54 and Figure 55.
```

We could support it as extra config option.
That would make SIP/SIE the same writeablity rules as MIP/MIE

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -76,27 +77,52 @@ def __init__(self, gen_params: GenParams):
...
+
+        if gen_params.supervisor_mode:
+            # mip_stip_no_stimecmp_acc
+            # mip_ssip_acc
+            # mip_seip_acc
+            self.mip_writeable |= (
+                (1 << InterruptCauseNumber.STI) | (1 << InterruptCauseNumber.SSI) | (1 << InterruptCauseNumber.SEI)
+            )
+
+        self.sie_writeable = (
```
awariac: maybe it would be easier to follow if this would be `mideleg_wrietable`, as SIE/SIP are shadows of M*

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -195,27 +235,25 @@ def elaborate(self, platform):
...
+            for i in range(self.gen_params.isa.xlen):
+                if self.edge_reported_mask & (1 << i):
+                    # edge-triggered bit
+                    m.d.av_comb += new_data[i].eq(mip_value[i] | new_interrupts_value[i])
+                elif self.mip_writeable & (1 << i):
+                    # software-bit
+                    m.d.av_comb += new_data[i].eq(mip_value[i])
+                else:
+                    # level-triggered bit
+                    m.d.av_comb += new_data[i].eq(new_interrupts_value[i])
```
awariac, line 261: there are special rules for SEIP - "Supervisor-level external interrupts are made pending based on the logical-OR of the software-writable SEIP bit and the signal from the external interrupt controller" (level signal).

We want to support external supervisor interrupts (even currently).

qbojj: I was thinking about this, but the biggest hurdle is the interaction with `csrrc` and `csrrs`, and that would require some refactor/special casing in the CSR unit

awariac: i think i designed the CSR interface with subfields, maps and priority including this use case in mind, but I might be wrong.
For now please provide it as an external level signal instead of software bit - for integration with PLIC and leave a FIXME

awariac (approved):

PR #946, opened 2026-05-10
remove condition() from CSRUnit

### LINKED ISSUE #918: Convert some uses of `condition` back to `m.If`
Thanks to kuznia-rdzeni/transactron#141, some uses of `condition` - done due to a common method being called in both branches - can now be safely converted back to `m.If`.

qbojj:
part of #918 - all other places seem to actually require condition.

For completeness those are places left with `condition`:
```
coreblocks/frontend/fetch/fetch.py:218
coreblocks/frontend/fetch/fetch.py:419
coreblocks/frontend/stall_controller.py:87
coreblocks/backend/retirement.py:226
coreblocks/core_structs/crat.py:183
coreblocks/core_structs/crat.py:326
coreblocks/func_blocks/csr/csr_unit.py:153
coreblocks/func_blocks/fu/lsu/dummyLsu.py:171
coreblocks/func_blocks/fu/lsu/lsu_requester.py:139
coreblocks/func_blocks/fu/lsu/lsu_requester.py:165
coreblocks/func_blocks/fu/priv.py:145
```

### DIFF
--- a/coreblocks/func_blocks/csr/csr_unit.py
+++ b/coreblocks/func_blocks/csr/csr_unit.py
@@ -7,7 +7,6 @@ from transactron import Method, Methods, def_method, def_methods, Transaction, T
 from transactron.utils import assign
 from transactron.utils.data_repr import bits_from_int
 from transactron.utils.dependencies import DependencyContext
-from transactron.lib.simultaneous import condition
 
 from coreblocks.arch import OpType, Funct3, ExceptionCause, PrivilegeLevel
 from coreblocks.arch.isa_consts import Opcode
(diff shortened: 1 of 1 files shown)


tilk (approved):

tilk:
Thought that there was a `condition` use inside LSU which was not strictly necessary, but it has a big "BUT": it's not necessary if readiness for loads and stores is the same. That's true for Wishbone, but might not be true for e.g. AXI. I looked at all the other uses of `condition` and it looks like none of them can be changed to `If`. So #918 can probably be closed with this PR.

awariac (approved):

PR #947, opened 2026-05-10
[MMU 2/3] Implement L2 TLB - set associative page table entry cache

qbojj:
Continuation of #942.

Implements set-associative L2, that will be acting as a backing device for L1 TLBs. It is slower, but uses less resources.

### DIFF
--- a/coreblocks/priv/vmem/tlb.py
+++ b/coreblocks/priv/vmem/tlb.py
@@ -1,9 +1,10 @@
 from amaranth import *
 from amaranth.lib.data import StructLayout, ArrayLayout, View
+import amaranth.lib.memory as memory
 
 from transactron import Method, TModule, def_method, Priority, Transaction
 from transactron.utils import DependencyContext, mod_incr
-from transactron.lib import Forwarder, condition, HwCounter, FIFOLatencyMeasurer
+from transactron.lib import Forwarder, Pipe, condition, HwCounter, FIFOLatencyMeasurer
 
 from coreblocks.arch.isa_consts import PAGE_SIZE_LOG, SatpMode
 from coreblocks.interface.layouts import AddressTranslationLayouts
@@ -14,6 +15,7 @@ from coreblocks.priv.vmem.iface import TLBBackingDevice
 
 __all__ = [
     "FullyAssociativeTLB",
+    "SetAssociativeTLB",
 ]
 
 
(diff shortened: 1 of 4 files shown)


test/priv/vmem/test_tlb.py
```diff
@@ -166,7 +167,16 @@ def elaborate(self, platform):
         return m
 
 
+@parameterized_class(
```
tilk: Prefer `pytest.mark.parametrize`.

coreblocks/priv/vmem/tlb.py
```diff
@@ -275,3 +277,277 @@ def _(vaddr, asid, all_vaddrs, all_asids):
...
+        self.sfence_vma.add_conflict(self.request, Priority.LEFT)
+
+        with Transaction(name="flush").body(m, ready=flushing & ~slow_path):
+            with m.If(~flush_fetched):
+                m.d.comb += set_rd.addr.eq(flush_set)
+                m.d.comb += set_rd.en.eq(1)
+                m.d.sync += cam.checked_asid.eq(flush_asid)
+                m.d.sync += cam.checked_vpn.eq(flush_vpn)
+
+                m.d.sync += flush_fetched.eq(1)
```
tilk: What is the reason why this is done in a different cycle than the `sfence_vma` call?

coreblocks/priv/vmem/tlb.py
```diff
@@ -275,3 +277,277 @@ def _(vaddr, asid, all_vaddrs, all_asids):
...
+
+            request_pipe.write(m, vpn=vpn, is_store=is_store)
+
+        slow_path = Signal()
+
+        with Transaction(name="TLBLookup").body(m, ready=~slow_path):
+            req = request_pipe.peek(m)
+
+            valid_vec = Signal(self.ways)
+            m.d.av_comb += valid_vec.eq(cam.valid_match & cam.addr_match & (cam.asid_match | cam.global_match))
```
qbojj: should be `cam.full_match`

coreblocks/priv/vmem/tlb.py
```diff
@@ -275,3 +277,277 @@ def _(vaddr, asid, all_vaddrs, all_asids):
...
+            m.d.sync += flush_all_vaddrs.eq(all_vaddrs)
+            m.d.sync += flush_all_asids.eq(all_asids)
+
+            m.d.sync += flush_set.eq(Mux(all_vaddrs, 0, vpn_to_set_idx(vaddr >> PAGE_SIZE_LOG, 0)))
+            m.d.sync += flush_fetched.eq(0)
+            m.d.sync += flush_size_class.eq(0)
+
+        self.sfence_vma.add_conflict(self.request, Priority.LEFT)
+
+        with Transaction(name="flush").body(m, ready=flushing & ~slow_path):
```
qbojj: request_pipe must also be empty

coreblocks/priv/vmem/tlb.py
qbojj: The reason is twofold:
- after the reset, we clear the whole tlb - this needs a place to initialise the flushing
- as I am waiting for ongoing operations to finish, and they can use the `checked_asid`/`checked_vpn` as well as read from `set_rd` I need a way to prevent collisions.

tilk: Huh; `request` and `sfence_vma` cannot be called in the same cycle because of the conflict, and these are the only places where `checked_asid` and `checked_vpn` are set.

As for `set_rd`, I don't understand what is supposed to happen if `sfence_vma` is called while `TLBLookup` and the FSM transactions are working.

qbojj: ok, I was just stupid

tilk (approved):

PR #948, opened 2026-05-10
[MMU 3/3] Implement page table walker

qbojj:
Continuation of #942.
This PR implements page table walker. It also implements randomized testing.
Even if this is called part 3 it can be merged independent of #947.

What will be left after this is just integrate the MMU into the core.

### DIFF
--- a/coreblocks/priv/vmem/walker.py
+++ b/coreblocks/priv/vmem/walker.py
@@ -175,7 +175,7 @@ class PageTableWalker(TLBBackingDevice, Elaboratable):
                         m.d.sync += access_fault.eq(1)
                         m.next = "DONE"
                     with m.Else():
-                        self.bus.request_read(m, addr=(pte_addr >> offset_bits), sel=~0)
+                        self.bus.request_read(m, addr=pte_addr[offset_bits:], sel=~0)
                         m.next = "EVAL"
             with m.State("EVAL"):
                 with Transaction().body(m):
--- a/coreblocks/priv/vmem/walker.py
+++ b/coreblocks/priv/vmem/walker.py
@@ -1,0 +1,266 @@
+"""Page Table Walker - Hardware implementation of RISC-V page table walking."""
+
+from amaranth import *
+from amaranth.lib.data import StructLayout, View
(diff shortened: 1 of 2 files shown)


awariac:
note for (future) consideration: we now have two possible read/write sources, and LSU "simplified" atomic operations may be incorrect on page table regions.
This would require some shared lock signal. (or maybe spec permits that?).

I think `fence` only orders explicit accesses so current implementation would be fine (but don't trust my memory of spec).

qbojj:
for now page table walker is read-only, so no need for any locking

coreblocks/priv/vmem/walker.py
```diff
@@ -0,0 +1,273 @@
...
+                    m.next = "IDLE"
+            with m.State("ISSUE"):
+                with Transaction().body(m):
+                    with m.If(~pmp_checker.result.r):
+                        log.debug(m, 1, "PMP check failed for PTE address {:x}", pte_addr)
+                        m.d.sync += access_fault.eq(1)
+                        m.next = "DONE"
+                    with m.Else():
+                        log.debug(m, 1, "Issuing bus request for PTE at address {:x}", pte_addr)
+                        self.bus.request_read(m, addr=pte_addr[offset_bits:], sel=~0)
```
tilk, line 179: It looks like there is some assumption missing. Is `offset_bits` equal to `word_width_bytes`?

Also, I believe we currently have a half-assed bus abstraction layer in Coreblocks. As different buses are differently addressed (Wishbone is word-addressed, AXI is byte-addressed), the abstraction layer should settle on one of the options. Currently, the layer basically forwards everything to Wishbone, so the convention is to use word addressing. (The AXI adapter seems to be broken.) I don't know if it's a good decision, maybe using byte addresing + sizes in the abstraction layer makes more sense. This needs an issue. (Edit: #963)

qbojj: I do assert on `pte_layout.as_shape().width == self.gen_params.isa.xlen`, I will change this to check of `te_layout.as_shape().width == self.bus.bus_params.data_width`

tilk (approved):

PR #949, opened 2026-05-11
Implement PLIC in CoreSocks

awariac:
Platform Level Interrupt Controller

It's unfortunately ratified and I hate it: https://docs.riscv.org/reference/plic/_attachments/riscv-plic.pdf

useful for Linux and when we reach multi-core


qbojj:
:skull:

coreblocks/peripherals/wishbone.py
```diff
@@ -316,9 +316,8 @@ class WishboneMuxer(Component):
     ----------
     wb_params: WishboneParameters
         Parameters for bus generation.
-    num_slaves: int
-        Number of slave devices to multiplex.
     ssel_tga: Signal
+        Nuber of slave devices is determied from `ssel_tga` bit width.
```
tilk:
```suggestion
        Number of slave devices is determied from `ssel_tga` bit width.
```
But actually, the very next line says almost the same thing: "Signal width is the number of slaves". This seems unnecessary.

coreblocks/socks/plic.py
```diff
@@ -0,0 +1,176 @@
...
+        context_treshold = Array([Signal(self.priority_bits) for _ in range(self.context_count)])
+        interrupt_priority = Array([Signal(self.priority_bits) for _ in range(self.interrupt_count)])
+
+        with m.If(is_perpiheral_request(self)):
+            m.d.comb += self.bus.err.eq(1)  # default - overwritten by memory mapped registers declaration
+            m.d.comb += self.bus.ack.eq(0)
+
+        # Step 1: Mask currently processed interrupts
+
+        # interrupt 0 is reserved and hard-wired to 0
```
qbojj, line 85: Maybe add an assert on ~self.interrupts[0]

coreblocks/socks/plic.py
```diff
@@ -0,0 +1,176 @@
...
+            with m.If(is_perpiheral_request(self) & (self.bus.adr == claim_addr)):
+                m.d.comb += assign(
+                    claim_complete_for_context,
+                    {
+                        "context": context,
+                        "claim": ~self.bus.we,
+                        "complete": self.bus.we,
+                    },
+                )
+                m.d.comb += claim_complete_for_context.complete_interrupt.eq(self.bus.dat_w)  # truncate
```
qbojj: this should not truncate the interrupt number, but ignore the values greater than the number of interrupts.
> The PLIC does not check whether the
completion ID is the same as the last claim ID for that target. If the completion ID does not match an
interrupt source that is currently enabled for the target, the completion is silently ignored.

As the not implemented bits are as-if they were not enabled (RO 0), the claims on them should be ignored

coreblocks/socks/plic.py
```diff
@@ -0,0 +1,176 @@
...
+            m.d.comb += self.bus.dat_r.eq(claim_response.value)
+            m.d.comb += self.bus.ack.eq(1)
+            m.d.comb += self.bus.err.eq(0)
+            m.d.sync += claim_response.set.eq(0)
+
+        with m.If(claim_complete_for_context.claim & ~claim_response.set):
+            claimed_interrupt = Signal(range(self.interrupt_count))
+            for interrupt in reversed(range(self.interrupt_count)):
+                with m.If(per_context_selected_priority_pendings[claim_complete_for_context.context][interrupt]):
+                    m.d.comb += claimed_interrupt.eq(interrupt)
```
qbojj: maybe something like `PriorityEncoder` would be better

coreblocks/socks/plic.py
```diff
@@ -0,0 +1,176 @@
...
+                m.d.comb += pending_to_disable.bit_select(claimed_interrupt, 1).eq(1)
+            m.d.comb += self.bus.ack.eq(0)
+            m.d.comb += self.bus.err.eq(0)
+
+        with m.If(claim_complete_for_context.complete):
+            # ignore completion for disabled interrupts in current context
+            with m.If(
+                enable[claim_complete_for_context.context].bit_select(claim_complete_for_context.complete_interrupt, 1)
+            ):
+                m.d.comb += processing_to_disable.bit_select(claim_complete_for_context.complete_interrupt, 1).eq(1)
```
qbojj:
```suggestion
            wants_claim = 1 << claim_complete_for_context.complete_interrupt
            m.d.comb += processing_to_disable.bit_select.eq(wants_claim & enable[claim_complete_for_context.context])
```

test/asm/plic.asm
```diff
@@ -0,0 +1,220 @@
...
+lw x1, 0x4(x16)
+bne x1, x0, fail
+
+# plic enable bit for irq 1, ctx 0
+li x1, 0b10
+li x16, 0xe2002000
+sw x1, (x16)
+
+li x16, 0xe2200004 # claim reg ctx 0
+lw x1, (x16)
```
qbojj: to be consistent with 19-21
```suggestion
li x16, 0xe2200000 # claim reg ctx 0
lw x1, 0x4(x16)
```

qbojj (changes requested):
minor issues, but looks good :3

coreblocks/socks/plic.py
```diff
@@ -0,0 +1,176 @@
...
+from transactron.utils import assign
+
+from coreblocks.peripherals.wishbone import WishboneInterface, WishboneParameters
+from coreblocks.socks.peripheral import (
+    SocksPeripheral,
+    convert_to_wishbone_addr,
+    gen_memory_mapped_register,
+    is_perpiheral_request,
+)
+
```
tilk, line 16: `__all__` missing.

coreblocks/socks/plic.py
awariac: this comment is about bit-truncation of index signal, because it cannot be assigned with `assign` (shape mismatch) and needs a separate statement. It doesn't do anything functional.
I can remove it if it's confusing

qbojj: I don't have a problem with the comment itself, but with the functionality.
Assuming we have `interrupt_count == 4` and someone request the completion of interrupt 5, current code will try to complete interrupt 1.

awariac: right, thanks

qbojj (approved):

awariac:
@tilk can we merge? this PR is a bit blocking

tilk:
I'll merge with bypass, ignoring the github runner timeout nonsense.

PR #950, opened 2026-05-11
[MMU] mmu integration

qbojj:
Based on #942, #947 and #948.

Merge of the MMU into the core, as well as enabling SV32 into the default configuration.


awariac:
have you tried verifying the critical paths with something like vivado?

qbojj:
Not yet. I will look into it

qbojj:
Totally forgot that the physical address range is greater when SV32 is enabled, so previous timings were all partially due to this.

awariac:
should phys range have such high impact on fmax?

qbojj:
It was not the only problem but deffinitely a part of it (critical paths were going via PMP)

qbojj:
The critical path of the basic config no longer goes through the MMU :3

coreblocks/priv/vmem/tlb.py
qbojj: NOTE:
changes to `FullyAssociativeTLB` were made to remove all combinatorial paths other than accept, requests from the L1. This makes the flush requests registered - no combinatorial path from any other path of the core than the address translator

coreblocks/priv/vmem/tlb.py
```diff
@@ -471,12 +490,13 @@ def _(vpn, is_store):
...
                     m.d.comb += rd_wr.addr.eq(refill_set_idx)
-                    m.d.comb += rd_wr.en.eq(1)
-
+                    m.d.comb += set_wr.addr.eq(refill_set_idx)
                     m.d.comb += set_wr.data.eq(set_rd.data)
                     m.d.comb += set_wr.data[cam.replace_candidate].eq(new_entry)
-                    m.d.comb += set_wr.addr.eq(refill_set_idx)
-                    m.d.comb += set_wr.en.eq(1)
+
+                    with m.If(refill_response.result == AddressTranslationLayouts.TLBResult.HIT):
```
qbojj, line 587: bugfix - make sure we only write valid entries - sometimes the test hit this problem

qbojj (commented):
Some comments to ease in review process

qbojj:
riscv-arch-test Sv32 status report:
[report.zip](https://github.com/user-attachments/files/29164739/report.zip)
[summary.log](https://github.com/user-attachments/files/29164740/summary.log)

TL;DR:
```
prinv/Sv/sv32_VA_all_ones_Smode.S
prinv/Sv/sv32_VA_all_zeros_Smode.S
prinv/Sv/sv32_global_pte_Smode.S
prinv/Sv/sv32_global_pte_Umode.S
prinv/Sv/sv32_mstatus_mprv_Smode.S
prinv/Sv/sv32_mstatus_mprv_Umode.S
prinv/Sv/sv32_pte_rsw_Smode.S
prinv/Sv/sv32_pte_rsw_Umode.S
prinv/Sv/sv32_satp_access_test.S
prinv/Sv/sv32_upage_mprv_set_sum_set_Smode.S
```
All pass.

`ExceptionsSv`, `SvPMP`, `Svade` timeout.
`ExceptionsSvZaamo`, `ExceptionsSvZalrsc` fail - same as `ExceptionsZaamo` and `ExceptionsZalrsc` respectively.
`priv/Sv/sv32_misaligned_page_Smode.S` and `priv/Sv/sv32_misaligned_page_Umode.S` fail by writing to unwriteable memory location `0x80000014` - to be checked.
The rest of the `Sv` tests timeout.

coreblocks/peripherals/bus_adapter.py
```diff
@@ -103,6 +108,45 @@ def __init__(self, bus_params: BusParametersInterface):
...
 
 
+class _WishboneSerializedPort:
+    def __init__(self, wb_params: BusParametersInterface, method_layouts: WishboneMasterMethodLayout):
+        self.wb_params = wb_params
+        self.request = Method(i=method_layouts.request_layout)
+        self.result = Method(o=method_layouts.result_layout)
+
+
+class WishboneMasterRequestResponseSerializer(Elaboratable):
```
tilk: We now have a serializer after a serializer. Maybe fold this functionality into the WB adapter? Design wise, I think this serialization should happen at the bus abstraction layer.

coreblocks/priv/vmem/tlb.py
```diff
@@ -128,6 +131,12 @@ def elaborate(self, platform):
 
         m.d.comb += self.next_replacement_rr_index.eq(mod_incr(self.replacement_rr_index, self.ways))
 
+        first_full_match_one_hot = Signal(self.ways)
+        m.d.comb += first_full_match_one_hot.eq(self.full_match & (~self.full_match + 1))
+        m.d.comb += self.matched_entry.eq(
+            or_value([Mux(first_full_match_one_hot[way], self.ways_data[way], 0) for way in range(self.ways)])
+        )
+
```
tilk: Should use new OneHotMux from Transactron 0.9?

coreblocks/priv/vmem/translation.py
```diff
@@ -2,7 +2,7 @@
 from amaranth.lib.enum import Enum, unique, auto
 
 from transactron import Method, TModule, def_method, Transaction
-from transactron.lib import Forwarder, condition
+from transactron.lib import condition, Forwarder
```
tilk: XD

tilk:
One more thing: I spoke with @lekcyjna123 and he thinks that L1 TLB should have multiple read ports for the purpose of superscalar address resolution. It seems to have more sense than a separate L1 TLB for each resolver, and it seems like that's what modern CPUs are doing.

coreblocks/priv/vmem/tlb.py
qbojj: yes

tilk (approved):
I think let's merge it, and fix issues as we go. The loss of IPC in machine mode is unfortunate, but maybe this can be reclaimed after LSU refactor.

tilk:
Also, I would like to merge @xThaid's PR first, he has limited time available so let's make things easier for him.

coreblocks/priv/vmem/tlb.py
```diff
@@ -183,109 +233,154 @@ def elaborate(self, platform):
...
-        def _():
-            self.perf_latency.stop(m)
-            return fwd.read(m)
+            return ask_backing, ret
 
-        @def_method(m, self.sfence_vma, ready=~request_in_flight)
-        def _(vaddr, asid, all_vaddrs, all_asids):
-            self.perf_flushes.incr(m)
+        for port in range(self.ports):
+            cam = TLBCAM(self.gen_params, self.entries)
```
lekcyjna123, line 263: Hmm... Should we create a separate TLBCAM for each port? I thought that 1 CAM for all ports should be used.

qbojj: there is no "memory" here - just a address/asid matching mechanism. I don't know how else should I call it.

If we want actually superscalar address translation this is needed.

lekcyjna123: ok, i see that this is a set not a whole TLB

lekcyjna123 (approved):

coreblocks/priv/vmem/tlb.py
tilk: I had the same doubt when reading this. Looks like something needs to be changed to make it clearer that the `TLBCAM` object does not actually own storage. Two things that caused my doubts:

- The name (M in CAM stands for `Memory`).
- It is not clear that `ways_data` is actually an input.

qbojj: I'm open for naming suggestions

qbojj:
I'm sorry for the changes besides the merge related ones, but I had found a nice mental model for `SFENCE.VMA`.

Currently adding `Svinval` will be free:
- `SFENCE.W.INVAL` only needs `side_fx_guard`
- `SINVAL.VMA` only needs to flush
- `SFENCE.INVAL.IR` only needs to stall the fetcher

From those only `SFENCE.INVAL.IR` and `SFENCE.VMA` need to stall the fetcher.

---

Some tangents:

We could also allow for concurrent flushing with refilling, but would need a bit (a single bit for each TLB instance - very funny) more book keeping, tiny bit more complex `SFENCE.W.INVAL` part of `SFENCE.VMA` (additional call), and resource management of the refill/flush port of both TLB kinds - trivial for L1 due to single-cycle flush, but more complex for L2.

As this is a possible performance improvement and not a soundness issue, I had left it for the future, though I think just implementing `Svinval` would make things much better for software (pipelined flushes and 

Tangent 2:

It would be quite easy to make L1 head-of-line non-blocking, but that would complicate the rest of the circuit - just a thought for the future.

tilk:
BTW, #979 starts to be more and more important - available resources on the ECP5 are rapidly dwindling for the full config...

tilk:
Wait a minute. Somehow, the multi port TLB change caused a massive increase in device usage. Can you explain this? This was supposed to be a refactor.

tilk:
Also, your VMA changes caused an IPC drop. I'm withdrawing my approval until these matters are sorted out.

tilk (changes requested):

qbojj:
> Also, your VMA changes caused an IPC drop.

It shouldn't have had. Maybe after merging the IPC drop from master got compounded.

tilk:
Please inform me when the PR stabilizes. It's probably better to merge this soon and then address shortcomings in further PRs, this is quickly becoming a mess.

qbojj:
Should be stable now.

tilk (approved):

PR #952, opened 2026-05-13
Fix mip.SEIP writeable

qbojj:
Proposed solution to SEIP semantics (see #945).

I believe this is least intrusive way to do this.

### DIFF
--- a/coreblocks/priv/csr/csr_register.py
+++ b/coreblocks/priv/csr/csr_register.py
@@ -201,7 +201,7 @@ class CSRRegister(CSRRegisterBase):
 
         self.fu_write_combine_read = MethodMap.create(
             self.fu_write_filter.method,
-            i_transform=(write_combined_layout, fu_write_combine_write),
+            i_transform=(self.csr_layouts.fu_write, fu_write_combine_write),
         )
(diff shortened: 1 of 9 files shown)


coreblocks/priv/csr/csr_register.py
```diff
@@ -177,24 +177,40 @@ def __init__(
...
-            i_transform=(self.csr_layouts.fu_write, lambda tm, ms: {"data": fu_write_filtermap(tm, ms["data"])[1]}),
+            i_transform=(write_combined_layout, lambda tm, ms: {"data": fu_write_filtermap(tm, ms["data"])[1]}),
         )
         self.fu_write_filter = MethodFilter.create(
             self.fu_write_map.method, lambda tm, ms: fu_write_filtermap(tm, ms["data"])[0]
         )
+
+        def fu_write_combine_write(m, data):
+            with m.Switch(data.op_type):
+                with m.Case(CSRRegisterLayouts.WriteOpType.CSR_WRITE):
```
qbojj: Oops

awariac (approved):
(for context) change based on our discussion.

Makes sense, turned out to be pretty clean.
Thanks for the improvement (and iterations here)!

PR #956, opened 2026-05-20
Efficient precommit

tilk:
There is no need to enable `precommit` just on the head of the ROB. When an instruction is preceded only by done, non-trapping insns, performing side effects early is safe.

Also, if any of the instructions in the done prefix is trapping, we already know that we will start flushing and no more side effects will need to be executed.

It also turns out that changing the semantics of `precommit` to only be enabled when not flushing (and therefore, side effects can be done) makes code simpler. Maybe the name should be changed to `can_side_fx` or something like that?

In the future, side effects could skip around instructions which are not done, but cannot trap nor perform side effects and are not dependent on side-effecting instructions (let's call them pure instructions?). CSRs would probably need to get special treatment as they can influence virtually freaking everything. <- ~this should get an issue~ #958

Unfortunately this patch breaks `rdinstret`. Fixing it requires working around the `DoubleCounterCSR` stuff: the length of the done prefix needs to be added to the counter value to get the proper `rdinstret` value.

This is a prerequisite for proper superscalar retirement.

Based on #957.

TODO:
- [x] Fix `rdinstret`
- [x] Fix tests

### DIFF
--- a/coreblocks/backend/retirement.py
+++ b/coreblocks/backend/retirement.py
@@ -86,7 +86,7 @@ class Retirement(Elaboratable):
         self.core_state = Method(o=self.gen_params.get(RetirementLayouts).core_state)
         self.dependency_manager.add_dependency(CoreStateKey(), self.core_state)
 
-        self.precommit = Method(i=layouts.precommit_in, o=layouts.precommit_out)
+        self.precommit = Method(i=layouts.precommit_in)
         self.dependency_manager.add_dependency(InstructionPrecommitKey(), self.precommit)
 
     def elaborate(self, platform):
@@ -98,8 +98,6 @@ class Retirement(Elaboratable):
         m_csr = csr_instances.m_mode
         s_csr = csr_instances.s_mode if self.gen_params.supervisor_mode else None
         m.submodules.instret_csr = self.instret_csr
-
-        side_fx = Signal(init=1)
 
         def free_phys_reg(i: int, rp_dst: Value):
             # mark reg in Register File as free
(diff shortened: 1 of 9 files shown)


coreblocks/func_blocks/csr/csr_unit.py
```diff
@@ -169,8 +167,7 @@ def elaborate(self, platform):
                         with m.If(priv_valid & csr_access_valid):
                             read_val = Signal(self.gen_params.isa.xlen)
                             with m.If(should_read_csr & ~done):
-                                with m.If(exe_side_fx):
-                                    m.d.av_comb += read_val.eq(csr._fu_read(m))
+                                m.d.av_comb += read_val.eq(csr._fu_read(m))
                                 m.d.sync += current_result.eq(read_val)
```
qbojj: I think now `current_result.eq(csr._fu_read(m))` is enough

coreblocks/backend/retirement.py
```diff
@@ -115,7 +119,7 @@ def retire_instr(rob_entry):
             # free old rp_dst from overwritten R-RAT mapping
             free_phys_reg(0, rat_out.old_rp_dst)
 
-            self.instret_csr.increment(m)
+            self.instret_csr.write(m, data=self.instret_csr.read(m).data + 1)
```
qbojj, line 122: I think extending the double counter to allow for multi increments would be better

qbojj (changes requested):

coreblocks/backend/retirement.py
tilk: But what kind of multi increments? Multiple increments using `Methods`, like in metric counters? Or by a value larger than 1 (from what range?). Or both? Both seem useful, but it's hard to know ahead of time which will actually be used in which place.

Also, a way to add the read map and possible other things to the internal register would be needed.

coreblocks/func_blocks/csr/csr_unit.py
tilk: More code removed, good. Done.

awariac (approved):

coreblocks/backend/retirement.py
qbojj: I think a single method would be enough, but I dont think this should block this pr

qbojj (approved):

Issue #937, opened 2026-05-02
writes to minstret have +1 too high value


awariac:
wee need to finally get rid of `DobuleCounterCSR` and add some rv32 *h aliasing to solve that cleany.
Either by tracking in write sidefx to CSR in retirement or mapping fu_write value to -1

PR #957, opened 2026-05-20
Double counter refactor

tilk:
This PR refactors the double counter, by introducing the following changes:

- The `ShadowCSR` class now has `width` and `offset` arguments, allowing to create a shadow of a part of a larger CSR.
- The internal counter is now always 64-bit. In RISC-V, all REG/REGH pairs are such that the base register is 64-bit, and the high register exists only on RV32.
- The arch-dependent logic is now internal, so that it does not need to be repeated at each use site.
- The dual-shadows part is now separate from the counter-part, so that the same logic can be used if the base register needs a different logic than a simple counter.

Required for #956.

### DIFF
--- a/coreblocks/priv/csr/shadow.py
+++ b/coreblocks/priv/csr/shadow.py
@@ -101,7 +101,7 @@ class ShadowCSR(CSRRegisterBase):
         def _(data: Value):
             self.shadowed.write(
                 m,
-                data=(data & write_mask) | (self.shadowed.read(m).data & ~write_mask),
+                data=((data & write_mask) << self.offset) | (self.shadowed.read(m).data & ~(write_mask << self.offset)),
             )
 
         @def_method(m, self.read, nonexclusive=True)
--- a/coreblocks/priv/csr/shadow.py
+++ b/coreblocks/priv/csr/shadow.py
@@ -22,7 +22,7 @@ log = logging.HardwareLogger("priv.csr.shadow")
 class ShadowCSR(CSRRegisterBase):
     """CSR shadow register.
 
-    Exposes an instruction-visible CSR number that reads/writes another CSR.
+    Creates a CSR which is backed by another CSR - reads and writes are forwarded to it.
     Optional bit masks can restrict visible read bits and writable bits for both instruction-visible access and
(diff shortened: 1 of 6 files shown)


coreblocks/priv/csr/shadow.py
```diff
@@ -33,13 +33,19 @@ def __init__(
...
         read_mask: Optional[ValueLike | Method] = None,
         write_mask: Optional[ValueLike | Method] = None,
         access_filter: Optional[Callable[[TModule, Value], ValueLike]] = None,
         src_loc: int | SrcLoc = 0,
     ):
-        super().__init__(gen_params, csr_number, width=shadowed.width, src_loc=get_src_loc(src_loc))
+        width = shadowed.width if width is None else width
+        self.offset = offset = 0 if offset is None else offset
+        assert self.offset < shadowed.width, "Offset larger than shadowed CSR width"
+        assert self.offset + width <= shadowed.width, "Shadowed wihdow outside of CSR"
```
awariac:
```suggestion
        assert self.offset + width <= shadowed.width, "Shadowed window outside of CSR"
```

coreblocks/priv/csr/shadow.py
```diff
@@ -33,13 +33,19 @@ def __init__(
...
+        super().__init__(gen_params, csr_number, width=width, src_loc=get_src_loc(src_loc))
```
awariac, line 60: There is a slight issue (from other PR) -

"Exposes an instruction-visible CSR number that reads/writes another CSR."
Instruction-visible CSRs must be xlen wide. That makes `width` unnecessary.
At the same time `csr_number` shouldn't be none - type should be int.

Or non-public CSRs are OK and docstring needs to be updated.

awariac (approved):
Thanks for refactoring that.
Makes #937 also possible to fix.

coreblocks/priv/csr/shadow.py
tilk: Don't see why non-public CSRs shouldn't be OK - flexibility is a good thing. Docstring it is.

qbojj (approved):

PR #966, opened 2026-05-23
increase riscv test CI timeout

qbojj:
It seems CI is failing lately due to timeout of "Run regression tests (riscv-tests)"


tilk (approved):

tilk:
Unfortunately riscv-arch-test started failing too.

PR #970, opened 2026-05-25
gen_verilog: add option to override core configs with env vars

awariac:
For easier integration with external toolchains like LiteX for custom configs injection (sometimes there would be no easy option to do that in the tool itself).
See instructions in https://github.com/litex-hub/pythondata-cpu-coreblocks.

### DIFF
--- a/coreblocks/gen_verilog.py
+++ b/coreblocks/gen_verilog.py
@@ -122,16 +122,16 @@ def main():
 
     os.environ["AMARANTH_verbose"] = "true" if args.verbose else "false"
 
-    if "COREBLOCKS_CORECONFIG_FILE" in os.environ:
-        config_file = SourceFileLoader("configfile", os.environ["COREBLOCKS_CORECONFIG_FILE"]).load_module()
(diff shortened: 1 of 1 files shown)


qbojj (approved):

tilk:
Kinda weird, typically it's the command line options which override the environment, not the other way around.

awariac:
@tilk  cli options would be always supplied by LiteX or other genration.
It is not really possible to inject dynamic variants into LiteX (it checks a static list before elaboration, from what i see there is no option to provide custom options to core).
This is made to override the whole toolchain on config file selection, while maintaing the old behavior in default cases.

As this may be not expected, there are print statements intserted. (and documented in pythondata_package, it should be doccumented on how to use docs page in future too).

I'm not sure of better way to implement this.
If I find some better way this can be removed.
It will also work if there is problem with some other toolchain integration, without modifiying its source

tilk:
Could you use "OVERRIDE" or some other word in the names of the env vars to make it absolutely clear that they have a weird behavior?

tilk (approved):

PR #973, opened 2026-05-28
test: fix pytest hook order so count runs after -k/-m (#753)

### LINKED ISSUE #753: Pytest option `--coreblocks-test-count` is not compatible with `-k`
The function `deselect_based_on_count` gets the full test list, instead of the filtered one. When using `-k`, the option `--coreblocks-test-count` can select only tests which are later filtered, and therefore no tests are run.

kambrch:
conftest'spytest_collection_modifyitems was registered after pytest's built-in _pytest.mark plugin, so by LIFO it ran first. This made --coreblocks-test-count slice the first N items from the full collection before -k/-m keyword/marker deselection, selecting the wrong tests (or none).

Add @pytest.hookimpl(trylast=True) so the name/count deselection runs after marker filtering, operating on the already-filtered set.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -90,6 +90,7 @@ def deselect_based_on_count(items: list[pytest.Item], config: pytest.Config) -> 
         items[:] = remaining
 
 
+@pytest.hookimpl(trylast=True)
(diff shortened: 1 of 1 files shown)


tilk (approved):
Thanks!

PR #983, opened 2026-06-18
fix pmp in fetcher

qbojj:


tilk:
Maybe I'm tired, but I'm looking at this thing for over 5 minutes and can't find the difference. Could you explain what happened here?

qbojj:
av_comb from the second condition was always active thus overwriting the access_fault. The tests (sometimes) worked due to implementation detail of ICache.

Effectively this was always
`m.d.av_comb += access_fault.eq(Mux(cache_resp.error, FetchLayouts.FaultFlag.ACCESS_FAULT, 0))`

tilk:
Damn it, `av_comb`. Changing to `comb` would fix this too, but your solution is better.

tilk (approved):

tilk:
Probably some information about this caveat (`av_comb` with `condition`) should be left in the docs (`condition` docstring?). It's all too easy to think of `condition` like it was `If` on steroids.

awariac:
yeah, or maybe the behavior should change for `condition`? No opinions for now

PR #985, opened 2026-06-22
Transactron 0.9

tilk:
This PR will update Transactron to 0.9, after its release. For now, it's to verify if CI passes after Transactron changes.


tilk:
Huh, IPC dropped?!?

Edit: looks like reduced conflicts caused the core to load more instructions, which reduced the performance. (#990)

qbojj (approved):

xThaid (approved):

PR #986, opened 2026-06-24
Pure insns

### LINKED ISSUE #958: Pure instructions
In superscalar retirement, side effects (e.g. loads/stores) could skip around instructions which are not done, but cannot trap nor perform side effects and are not dependent on side-effecting instructions (let's call them pure instructions?). CSRs would probably need to get special treatment as they can influence virtually freaking everything.

tilk:
This PR adds the concept of "pure instructions", which allows to run side effects quicker. The proposed implementation uses optypes to distinguish them.

Draft for now, because this can possibly be optimized with respect to combinational delays.

Closes #958.

Edit: huh,  no benchmark change, weird. A bug, maybe?
Edit: now there is a change, but surprisingly low.


coreblocks/func_blocks/fu/priv.py
```diff
@@ -119,7 +119,7 @@ def _(arg):
 
         with Transaction().body(m, ready=instr_valid & ~finished):
             precommit = self.dm.get_dependency(InstructionPrecommitKey())
-            precommit(m, instr_rob)
+            precommit(m, rob_id=instr_rob, require_done=0)
```
qbojj, line 122: I think FENCE.I and SFENCE.VMA need `require_done=1` - they require LSU writes to be globally visible

qbojj (approved):
Other than the thing with require_done for priv unit I don't see any issues.

EDIT: all good now

coreblocks/func_blocks/fu/priv.py
tilk: Why? LSU writes are not pure.

qbojj: It went over my head. My bad

PR #992, opened 2026-06-29
Use OneHotMux where appropriate

### LINKED ISSUE #975: Use OneHotMux where appropriate
In kuznia-rdzeni/transactron#180, there is an improved version of `OneHotMux` which has nice performance, and it can double as a priority mux thanks to a cool bit/arith trick. It should be used, to simplify code and/or improve performance, in:

- Superscalar retirement
- PMA
- FU decoder
- RS insertion
- Other places...?

tilk:
`OneHotMux` is currently more efficient than `OneHotSwitch*`, and it also implements nice FPGA-optimized priority muxing. This PR applies the new functionality where appropriate.

### DIFF
--- a/coreblocks/peripherals/wishbone.py
+++ b/coreblocks/peripherals/wishbone.py
@@ -7,7 +7,7 @@ import operator
 from transactron import Method, def_method, TModule
 from transactron.core import Transaction
 from transactron.lib import AdapterTrans, BasicFifo
-from transactron.utils import OneHotSwitchDynamic, assign, RoundRobin
+from transactron.utils import OneHotMux, assign, RoundRobin
 from transactron.utils.amaranth_ext.component_interface import ComponentInterface, CIn, COut
 from transactron.lib.connectors import Forwarder
 from transactron.utils.transactron_helpers import make_layout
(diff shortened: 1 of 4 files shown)


coreblocks/priv/pmp.py
```diff
@@ -116,20 +116,9 @@ def elaborate(self, platform) -> HasElaborate:
 
             entry_matches.append(entry_match)
 
-        matches = Cat(entry_matches)
-        one_hot = matches & (~matches + 1)
+        selected = OneHotMux.create(m, [(match, cfg) for match, cfg in zip(entry_matches, cfgs)], priority=True)
```
qbojj: isn't this just `list(zip(entry_matches, cfgs))`?

tilk: Even `list` is not necessary here.

Probably written this automatically because sometimes the matches are on a signal, and pyright doesn't like signals in iterable contexts- Signal doesn't have `__iter__`, but iteration works via `__getitem__`. I think Amaranth should provide `__getitem__` on signals. (Another omission I'm aware of is slicing on array views.)

coreblocks/peripherals/wishbone.py
```diff
@@ -373,11 +373,12 @@ def elaborate(self, platform):
...
-        for i in OneHotSwitchDynamic(m, self.sselTGA):
-            # mux S->M data
-            # workaround for the lack of selective connecting in wiring
-            for n in ["dat_r", "stall"]:
-                m.d.comb += getattr(self.master_wb, n).eq(getattr(self.slaves[i], n))
+        # mux S->M data
+        # workaround for the lack of selective connecting in wiring
+        for n in ["dat_r", "stall"]:
+            m.d.comb += getattr(self.master_wb, n).eq(
+                OneHotMux.create(m, [(i, getattr(self.slaves[i], n)) for i in range(len(self.slaves))])
```
qbojj:
```suggestion
                OneHotMux.create(m, [(self.sselTGA[i], getattr(self.slaves[i], n)) for i in range(len(self.slaves))])
```

qbojj (changes requested):

qbojj (approved):

Issue #990, opened 2026-06-29
Clean-the-core flush


tilk:
One thing I learned when implementing superscalarity: the current flushing implementation, where you have to wait patiently for each and every instruction to finish, is a huge performance burden. It makes the flush time proportional to the number of instructions in the core, which is a limiter on speculation depth. This disadvantage will be reduced with better prediction, but will never disappear completely.

The current implementation was born from the need for simplicity, to have something that works. Now, with most important features already here, it's probably time to revisit this old decision. The flush doesn't need to be single-cycle; it just needs to be constant time with respect to the number of instructions in flight.

PR #993, opened 2026-06-30
Towards fast flush - single cycle RAT overwrite

tilk:
This PR is a work in progress towards #990. Implemented changes:

- RRAT using `ArrayLayout` instead of a memory.
- Single-cycle RRAT->FRAT overwrite.

Clearing the free regs in single cycle requires Transactron changes.


qbojj (approved):
Seems simple enough

awariac:
The same thing was done in checkpoint integration - #846 so all the changes are necessary anyway.

awariac (approved):

PR #994, opened 2026-07-02
Add instruction pipeline visualization

### LINKED ISSUE #763: Instruction tracking debugging interface
Debugging assembly tests (incl. regression tests) on traces is hard.

It would be great to create tool that prints information about each executed instruction (and maybe inserted and flushed ones in verbose mode) with current register state.

Implementation proposition (there may be better ways to do that):
Collect instruction information via calling methods in a similar way that our hardware logging works at a few points:
1. Collect raw instruction at exit of fetch unit
2. Maybe assign some temopary id in pipeline and map it in ROB insertion when rob_id is known.
3. Report rob_ids of commited instruction in Retirement and current RRAT state.
4. Optional: Collect reported exceptions and OpTypes
(synthesis of extra features enabled in configuration)

Instructions from raw bits could be decoded and preety-printed externally.

It would be great if it worked both in python and faster verilator simulation, but I don't know if it is possible.

xThaid:
A few changes were required to achieve this:
- `ftq_offset` is now part of the instruction payload up to ROB allocation, so we can trace them
- new `coreblocks.telemetry` package with events tracing instructions through the whole pipeline
- a converter from captured event logs to the Kanata format, for viewing pipeline traces in Konata (https://github.com/shioyadan/Konata). The UI of the viewer is a bit wonky, but this is the best I could find and I think should be good enough for now. 

Depends on https://github.com/kuznia-rdzeni/transactron/pull/189

Usage:
1. run your favorite test (this _should_ also work on benchmarks, but haven't tested it yet - bechmark run times grew significantly in the past year hehe) 
```
$ pytest --coreblocks-backend=pysim --coreblocks-regression --coreblocks-evlog -k "rv32uc-rvc"
```
2. manually inspect events
```
$ python -m transactron.cmd.evlog test/__evlogs__/test.regression.rv32uc-rvc.jsonl | rg "rob_id=56"
 412 backend.scheduler     backend.rob_allocate     ftq_ptr=18 ftq_offset=2 rob_id=56
 415 backend.wakeup_select backend.fu_issue         rob_id=56 unit=alu
 416 backend.announcement  backend.exec_complete    rob_id=56
 417 backend.retirement    backend.rob_retire       rob_id=56
```
3. generate logs in Konata format
```
$ python -m coreblocks.telemetry.konata test/__evlogs__/test.regression.rv32uc-rvc.jsonl > /tmp/coreblocks.log
```
4. profit

Pics:
image
why there are two cycles between retiring instructions (7, 8) and (9, 10)? (this could be a bug in tracing)

This also shows that icache misses hurt (the big purple block).

image
Sequential execution of SYSTEM instructions.

image
misprediction + pipeline flush!

I think this fixes #763

### DIFF
--- a/coreblocks/telemetry/events.py
+++ b/coreblocks/telemetry/events.py
@@ -1,0 +1,133 @@
+import re
+
+from transactron.evlog import Event, Static, event
+
+__all__ = [
+    "FTQAlloc",
+    "FetchRequest",
+    "InstrFetched",
+    "InstrDecoded",
+    "FTQRollback",
+    "FTQCommit",
+    "SchedulerEnter",
+    "RobAllocate",
+    "RobRetire",
+    "RobFlush",
+    "FuIssue",
+    "ExecComplete",
+    "func_unit_kind",
+]
+
+
+def func_unit_kind(func_unit) -> str:
+    """A short name for a functional unit instance, derived from its class
+    name: ``JumpBranchFuncUnit`` -> ``jump_branch``."""
+    name = type(func_unit).__name__.removesuffix("FuncUnit").removesuffix("Unit")
+    # insert '_' at camelCase word boundaries, keeping acronym runs (e.g. LSU) together
+    return re.sub(r"(?<=[a-z0-9])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])", "_", name).lower()
+
+
+@event("frontend.ftq_alloc")
+class FTQAlloc(Event):
+    """An FTQ entry was allocated for the fetch block at `pc`."""
+
+    ftq_ptr: int
+    pc: int
+
+
+@event("frontend.fetch_request")
+class FetchRequest(Event):
+    """The FTQ issued a request to the instruction fetch unit for the fetch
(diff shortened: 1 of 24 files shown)


coreblocks/telemetry/events.py
```diff
@@ -0,0 +1,168 @@
...
+    "RobRetire",
+    "RobFlush",
+    "FuIssue",
+    "ExecComplete",
+    "FuncUnitKind",
+    "func_unit_kind",
+]
+
+
+class FuncUnitKind(enum.Enum):
```
tilk: Smells like unneeded redudancy. Is there a way to avoid it?

xThaid:
Zooming out on a trace from a benchmark also looks interesting
image
1. we start with a tight loop
2. then we have a misprediction when exitting the loop
3. gaps between the big blocks are caused by "ret" instructions - we can't currently predict them and the fetch unit stalls

awariac:
@camel-cdr you would be interested in this ;)

tilk (approved):
Please fix workflows.

PR #999, opened 2026-07-05
Isa extensions native

qbojj:
Closes #879

I don't know why `extensions_supported` did implied extension removal - could be removed now


tilk:
Huh, why is the test removed?

qbojj:
thought that it only checked string to extension set functionality of `ISA`, turns out it also tested the extension implications.

tilk (approved):

Issue #988, opened 2026-06-26
Exploration of ways towards better MMU


qbojj:
## PMP in TLB

PMP attributes can be handled via address translator and stored in the TLB. This would reduce the critical path.
The ISA acknowledges this possible behaviour by requiring `sfence.vma` after chaning the pmp attributes if any non-bare `satp` mode is implemented.
> The PMP settings for the resulting physical address may be checked (and possibly cached) at any point between the address translation and the explicit memory access.

This would require:
- 4kB PMP granule - making all small pages pmp-uniform and splitting non-pmp-uniform huge+ pages into smaller pmp-uniform pages (aka TLB atomization)
- special handling of non-uniform pages - having a `non-uniform` bit so pmps can be checked a cycle after the tlb hit.

The first options has precedent in SiFive MMU capable cores (e.g. U74) - at least that would make sense of the increase of PMP granularity.

Either option requires a change to the PMP checker to check for uniformity over an address range. (#1000)

Even more extreme version of this is storing all memory attributes in the TLB for both BARE and SV*, so:
- PMP in TLB
- PMA in TLB - already semi-mandated by Svpbmt
- BARE mode gets the attributes also from TLB (additional addressing mode in TLB), thus fully removing PMP/PMA from the main path of the core.

## Svnapot [RV64]

Adding real 64kB pages (only 64kB are currently added via Svnapot) only requires adding a new `size_class` in the tlb and changing calculation of the bits of specific `size_class`es.
Alternatively (what should be done if the first option is not selected) - the PTW can generate the TLB entry from NAPOT PTE by replacing the lower bits of PPN from PTE with lower bits for VPN - effectively making the extension noop.

## Svpbmt [RV64]

This is simply done by adding a two bit enum into the TLB structure - PMA/NC/IO flag for each entry - than the rest of the infrastructure can use those.

## Svinval

`sfence.vma` does three things:
1. `sfence.w.inval` - make core's writes visible to the PTW (next clear cache operation must wait on LSU)
1. `sinval.vma` - clear cache according to instruction's parameters - synchronizes with `sfence.w.inval`
1. `sfence.inval.ir` - make all next translations use new values (flush and stall fetcher) - sychronizes with `sinval.vma`

[...]

PR #1000, opened 2026-07-05
implement area checking pmp

qbojj:
Adds support for checking accesses of greater length than the PMP grain.

Useful for #988


coreblocks/priv/pmp.py
```diff
@@ -3,25 +3,172 @@
...
 from coreblocks.arch.isa_consts import PMPAFlagEncoding, PMPCfgLayout, PrivilegeLevel
 from coreblocks.interface.keys import CSRInstancesKey
 from coreblocks.params import *
 
 
 class PMPLayout(data.StructLayout):
+    r: Value
+    w: Value
+    x: Value
+
```
tilk: ?!?

qbojj: I had tried to make LSP see those values, but I guess either way this wouldn't work

tilk: These annotations are not valid on a `Layout`, as layouts don't have such attributes (and certainly not with values). And Pyright does not magically know to transfer them to a `View` using such a `Layout`.

tilk:
No tests.

coreblocks/priv/pmp.py
```diff
@@ -15,13 +15,151 @@ def __init__(self):
...
+                        "x": cfgs[i].X,
+                        "l": cfgs[i].L,
+                    },
+                },
+            )
+
+        default_input = Signal(layout)
+        m.d.comb += default_input.matches_all.eq(1)
+        m.d.comb += default_input.entry_data.eq(PMPLayoutFull().const({"r": 0, "w": 0, "x": 0, "l": 0}))
+
```
tilk, line 165: Can this be defined using `layout.const`?

coreblocks/priv/pmp.py
```diff
@@ -15,13 +15,151 @@ def __init__(self):
...
+        # entry fully covers the match if all entries before the fully matching one don't match any bytes.
+        # this is the same as checking the the first any-matching entry is also fully matching
+        selected = OneHotMux.create(
+            m, [(matches_any[i], all_data[i]) for i in range(n)], priority=True, default_input=default_input
+        )
```
tilk, line 170: What is the use case? Is the algorithm defined somewhere in the spec? If I understand correctly, this finds the first matching entry for a region. Why is this useful?

qbojj: >  Is the algorithm defined somewhere in the spec?

> PMP entries are statically prioritized. The lowest-numbered PMP entry that matches any byte of a memory operation determines whether that operation succeeds or fails. The matching PMP entry must match all bytes of a memory operation, or the operation fails, irrespective of the L, R, W, and X bits.

> Why is this useful?

To store pmp permissions in the TLB, the whole region must be uniform - setting the grain to 4kB is not enough, as there are bigger pages. When atomizing big pages, we want to select as big of a page type as possible not to trash the TLB. Also some operations may be bigger than the grain size.

tilk: > The matching PMP entry must match all bytes of a memory operation, or the operation fails, irrespective of the L, R, W, and X bits.

I think this is the bit I missed. So partial matches should be failures, according to spec.

coreblocks/priv/pmp.py
```diff
@@ -15,13 +15,151 @@ def __init__(self):
...
+
+
 @unique
 class PMPOperationMode(Enum):
     LSU = auto()
     INSTRUCTION_FETCH = auto()
     MMU = auto()
 
 
+class AreaPMPChecker(Elaboratable):
```
tilk: Another question: the `PMPChecker` implements the spec logic, while `AreaPMPChecker` basically only reports the first matching entry, plus `uniform` bit. The naming is confusing for me - the `AreaPMPChecker` is not doing the entriety of checking, just the initial part.

This also means that when `AreaPMPChecker` is used as is, spec conformity would need to be ensured via other means.

test/func_blocks/lsu/test_pmp.py
```diff
@@ -74,6 +75,44 @@ async def process(sim: TestbenchContext):
         with self.run_simulation(test_module) as sim:
             sim.add_testbench(process)
 
+    def run_dynamic_pmp_test(
```
tilk: This is not called from anywhere.

tilk (approved):

coreblocks/priv/pmp.py
```diff
@@ -15,41 +15,65 @@ def __init__(self):
...
+        """
+        mode: PMPOperationMode
+            Effective mode depends on `mode`:
+            - LSU: MPRV-aware (using MPP when MPRV=1 and current mode is M)
+            - INSTRUCTION_FETCH: uses only current privilege mode
+            - MMU: always behaves as supervisor mode
+        size_classes: list[int]
+            The log2 of power-of-two for which the checker can do the operations.
+            The size_class signal selects the value from these.
+        """
```
tilk, line 74: Use numpy docstring style. Move `PMPOperationMode` explanations to `PMPOperationMode`. Add similar docstring to `PMPChecker`.

tilk (approved):

PR #1004, opened 2026-07-06
Store FTQ pointers in RS and use them when redirecting the frontend

xThaid:
- Made the `FTQ::backend_redirect` function take a ftq_ptr so we can redirect the fetch unit before all active instructions are commited (useful for checkpointing)
- Passed the FTQ pointer of an instruction all the way to RSes. I think it is useful for functional units to know it: jumpbranch for reading the predicted target, csr & priv for redirecting the fetch unit.

### DIFF
--- a/coreblocks/backend/retirement.py
+++ b/coreblocks/backend/retirement.py
@@ -10,6 +10,7 @@ from coreblocks.interface.layouts import (
     RFLayouts,
     ROBLayouts,
     RetirementLayouts,
+    FTQPtr,
 )
 
 from transactron.core import Method, Methods, Transaction, TModule, def_method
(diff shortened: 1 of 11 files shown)


awariac:
I think you can implement `rollback_handler` directly inside `ftq.py` (proof of concept in #846).

xThaid:
sorry, I might be missing something, but can't we use the new `backend_redirect` function for rollbacks? The `rollback_handler` function in your PR seems to be exactly the same as the new `backend_redirect`. Also, FTQ is going to be pretty complex, so it'd be nice if it can remain oblivious to check-pointing (which seems likely).

awariac (approved):

qbojj (approved):

PR #1007, opened 2026-07-07
frontend: rename prediction-checker output fields, expose cfi_type

xThaid:
This is mostly to unify it with other modules that use `cfi_`-like naming. No functional changes expected.


tilk (approved):

tilk:
@xThaid Somehow, fmax dropped in basic config after #1004, could you investigate and fix?

PR #1008, opened 2026-07-08
frontend, backend: decouple predicted_taken from instructions

xThaid:
Currently, we're piggybacking on `funct7` to send information to the `jumpbranch` unit about branch prediction. We will no longer be able to do this once we start predicting jump targets.

As a permanent solution, I am introducing a new memory bank in the FTQ. This bank will contain all the information the jumpbranch unit needs to determine if a misprediction occurred. Since we only save information for taken CFIs, having one entry per FTQ entry is sufficient.

Depends on #1007


tilk:
Benchmarks broken, converting to draft.

tilk (approved):

PR #1009, opened 2026-07-08
Add Xilinx platforms to CI

### LINKED ISSUE #979: Add CI synthesis on a larger FPGA
The `full` configuration is steadily filling up the capacity of the ECP5 chip. We're going to need a bigger target to continue measuring it. NG-Ultra is a nice target because it is supported by nextpnr via Project Beyond and has 550K LUTs, more than enough for our uses. We're probably not going to have a physical instance of the chip, but who knows...?

Edit: openXC7 is probably a better target for now.

tilk:
This PR adds a Kintex-7 target to the CI, via openXC7. This should allow us to continue adding features which risk making the core too big for standard sized FPGAs.

Still, I'd like to still have a compact core config which can run Linux and fits on a ECP5. Changes to core config set are left to future PRs.

Closes #979.

TODO:
- [x] Docker image with required tools.
- [x] Workflow.


tilk:
New snags:

- `yosys-abc` started crashing for some reason
- block RAM don't seem to be inferred

awariac:
is RAM infering related to --enable-vivado-hacks maybe?

tilk:
> is RAM infering related to --enable-vivado-hacks maybe?

Found the issue - it turned out that Xray workflow passes `-nobram` to `synth_xilinx`, disabling block RAM entriely. It looks like the Xray workflow in Amaranth is pretty stale.

tilk:
Finally works, but the Fmax results are pretty weird. They are both lower than ECP5, and the result for the `full` config is especially bad. I wonder if I'm doing something wrong.

tilk:
Huh, this weird Fmax drop doesn't occur on the Kintex target. I guess this settles it.

tilk:
The resulting Fmax values are now weirdly close. I'm going to experiment with `nextpnr` target frequency values before merging this - I'm not certain how much it affects results.

Edit: looks like setting the target frequency doesn't matter much.

xThaid (approved):

qbojj (approved):

PR #1010, opened 2026-07-08
Stall ctrl OneHotMux

tilk:
Missed in #1004. PriorityEncoder + Array combination is pretty awful, as our previous PMP adventures have shown.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -26,7 +26,7 @@ jobs:
         )
       )
     runs-on: ubuntu-24.04
-    timeout-minutes: 45
+    timeout-minutes: 75
     container: ghcr.io/kuznia-rdzeni/amaranth-synth:ecp5-2025.01.31
     steps:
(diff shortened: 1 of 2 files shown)


qbojj:
I think there is still one pattern like that in `WakeupSelect`:
- `WakeupSelect` has PriorityEncoder and requests taking an entry via `RS.take` (via index)
- `RS.take` calculates `actual_rs_entry_id <- self.order[rs_entry_id]`
- `RS.take` returns `self.data[actual_rs_entry_id]`

As right now the RS size is 2 or 4, this doesn't hurt that much.

tilk:
Indeed! But this is baked into the method interfaces, unfortunately. It might make sense to change this, but it would be easier to do it after LSU refactor and removing the ordered RS abomination.

tilk:
Aaaand the `basic` Fmax is still bad, oh well. I wonder what happened.

qbojj (approved):
It may not fix the problem, but it is not worse

Issue #924, opened 2026-04-30
Interrupt tests run very slow


tilk:
The interrupt full core tests are much too slow. They run so long that I basically always cancel them when running full core tests in development. They also take a really long time in CI, making project management tasks slower. Summarizing, they are super inconvenient. I think they can be scaled back for now.

awariac:
maybe: extended tests can be run only on `master` commits, while defaults to scaled back

PR #1012, opened 2026-07-09
Run core tests first

tilk:
I noticed that the long running tests (especially interrupt tests, see #924) tend to linger to the end of the test run, which reduces the advantage of running tests on multiple cores. This change moves the costly tests to the beginning, making it more probable that the tests are distributed well.


tilk:
Looks like this is indeed faster by over 10 minutes.

qbojj (approved):

PR #1017, opened 2026-07-10
implement RVVI-TRACE

### LINKED ISSUE #984: Implement RVVI
[RVVI-TRACE](https://github.com/riscv-verification/RVVI/tree/master/RVVI-TRACE) is an core interface that exposes some information about the instructions getting retired.

Some tools like riscv-arch-test integrate it for coverage testing. It would also allow for better debugging.

It would be nice to have an option to generate the core with this interface exposed

qbojj:
Some part of `fibonacci.asm` with help of surfer with builtin RV32 data formatter.
Screenshot from 2026-07-11 01-47-22


awariac:
@kiryk

tilk:
Please fix formatting.

qbojj:
I don't have access to the computer untill Sunday.

coreblocks/frontend/fetch/fetch.py
```diff
@@ -253,6 +254,8 @@ def _(pc, ftq_ptr, fetch_gen):
             expanded_instr = [Signal(self.gen_params.isa.ilen) for _ in range(fetch_width)]
             is_rvc = Signal(fetch_width)
 
+            full_instrs = [Signal(self.gen_params.isa.ilen) for _ in range(fetch_width)]
```
tilk: I don't like this name, when I saw it I was wondering how is it different from `expanded_instr`. The name `full_instr` was used only in the context of the C extension, where it meant an instruction which was possibly reconstructed from two fetch blocks. Here, we just mean instructions as they are in the instruction stream, no need to concern ourselves with the minor detail of fetch blocks. Maybe just `instrs`?

coreblocks/telemetry/rvvi.py
```diff
@@ -0,0 +1,229 @@
...
+
+        self.mode = COut(2)  # optional
+        self.mode_virt = COut(1)  # optional
+        self.ixl = COut(2)  # optional
+
+
+class RVVIHartCollector(Component):
+    """A module that collects information required for RVVI-TRACE interface for a single hart.
+
+    Currently doesn't implement `pc_wdata`, `f_*`, `v_*`, `csr_*`.
```
tilk, line 59: Why is `pc_wdata` not implemented?

qbojj: I couldn't find how it should interact with async interrupts and I would prefer this to be correct

tilk:
The feature is not tested in CI. As experience shows (see e.g. #987), if a feature is not tested in CI, it inevitably breaks.

tilk (approved):

Issue #832, opened 2025-07-06
perf metrics from JumpBranchUnit are not exact


piotro888:
Perfs are incremented on every instruction that passes through `JumpBranchUnit`.
This does not account for ex. core being flushed.
What's worse, triggering a perf metric is basically a side effect, so (as this is not done on precommit), we don't know if the instruction will be not flushed in the future (or after checkpointing: on correct speculation path).

I'm not sure what would be the best approach to handle this and not overcomplicate the hw.
As this is a metric that will be accessible via CSR, I think the assumption is that it should be exact.

Maybe we should count the actual rollback events (or now: processed exceptions) as count of mispredictions? 
That's the best we can get - as this events executes some actual action in the hardware - when executing speculatively. Is this what we are interested in when checking for mispredicition counter in CSRs?

PR #1020, opened 2026-07-14
Implement HPM counters

xThaid:
Our existing metric `backend.fu.jumpbranch.mispredictions` is not entirely accurate, because it counts speculative (and flushed?) instructions (related issue #832).

Here I'm moving the metric to the retirement module. I also implemented HPM counters, so the misprediction count can be accessed from software.

Note that the RISC-V spec doesn't define what each HPM counter counts - it's platform-specific. I think the branch misprediction count is typically exposed through HPM counters (RISC-V or not), so it's reasonable to add it here.


.github/workflows/main.yml
```diff
@@ -191,8 +191,11 @@ jobs:
           submodules: recursive
 
       - if: ${{ steps.cache-arch-regression-tests.outputs.cache-hit != 'true' }}
```
xThaid, line 193: My change invalidated the build cache and I think uncovered a `mise` race. `make -j $(nproc)` below would implicitly run `mise install` concurrently and break stuff

tilk: @qbojj ?

tilk (approved):

PR #1021, opened 2026-07-16
Checkpointing part 1.5: Stall control, Exception Invalidation

awariac:
Part of rewrite of #846 

* Allows ExceptionInvalitationRegister information to be invalidated with tag changes
* Rework how exceptions and core interacts with fetch stall controller


tilk:
Can you explain the IPC loss?

awariac:
fetch flush was moved from 'on exception report' to 'on instruction with exception seen in retirement' to keep things as simple as possible for now.
I will change it next two part - excpetion invalidation needs to be preceded by a rollback anyway.

coreblocks/frontend/stall_controller.py
```diff
@@ -68,7 +77,19 @@ def _():
             pass
 
         with Transaction().body(m):
-            log.assertion(m, self.redirect_frontend.ready)
+            # Treat core flushing as a latched valid excpetion, all tags will be freed and invalidated during flush
+            exception = self.get_exception_information(m)
+
+            with m.If((exception.valid | core_state.flushing) & ~stalled_exception):
+                log.debug(m, True, "Stalling frontend - pending exception on speclation path")
```
qbojj: typo: speclation -> speculative?

awariac:
linux boot verified

coreblocks/priv/traps/exception.py
```diff
@@ -81,55 +53,56 @@ def call(m: TModule, **kwargs):
...
+            return {"data": self.data, "valid": self.valid}
 
         @def_method(m, self.clear)
         def _():
             m.d.sync += self.valid.eq(0)
-            for clear in self.clears:
+            for clear in self._clears:
                 clear(m)
-            del self.clears  # exception will be raised if new fifos are created later
+            del self._clears  # exception will be raised if new fifos are created later
```
qbojj, line 106: Not really comment for this pr, but this should be a key with unifier

qbojj (commented):
Seems ok, but I will take another look

coreblocks/frontend/stall_controller.py
```diff
@@ -42,24 +38,37 @@ class StallController(Elaboratable):
...
+    get_exception_information: Required[Method]
+    """Gets information from `ExceptionInformationRegister`."""
+
     def __init__(self, gen_params: GenParams):
         self.gen_params = gen_params
 
-        layouts = self.gen_params.get(FetchLayouts)
-
         self.stall_unsafe = Method()
         self.stall_exception = Method()
```
tilk: Remove.

coreblocks/frontend/stall_controller.py
```diff
@@ -68,7 +77,19 @@ def _():
             pass
 
         with Transaction().body(m):
-            log.assertion(m, self.redirect_frontend.ready)
+            # Treat core flushing as a latched valid excpetion, all tags will be freed and invalidated during flush
```
tilk:
```suggestion
            # Treat core flushing as a latched valid exception, all tags will be freed and invalidated during flush
```

tilk:
> fetch flush was moved from 'on exception report' to 'on instruction with exception seen in retirement' to keep things as simple as possible for now.

Any particular reason why keeping the original behavior would complicate things in this PR? It will be harder to confirm the regression was fixed in future PRs.

qbojj (commented):
I have looked it over again:
- FUs that have side effects are running even for inactive tags. Currently we fully flush the core on missprediction, but when we will be just invalidating the core - this needs to be handled. Do you handle this in the next batch of PRs?
- Once other comments are resolved I will approve

awariac:
> FUs that have side effects are running even for inactive tags. Currently we fully flush the core on missprediction, but when we will be just invalidating the core - this needs to be handled. Do you handle this in the next batch of PRs?

yes. There is a bit more left to implement, like rollback handling in frontend.

coreblocks/priv/traps/exception.py
tilk: Leave an issue?

awariac:
i'm not sure about the current loss

coreblocks/frontend/stall_controller.py
```diff
@@ -42,24 +37,40 @@ class StallController(Elaboratable):
...
 
-        DependencyContext.get().add_dependency(UnsafeInstructionResolvedKey(), self._resume_from_unsafe)
+        self.dm = DependencyContext.get()
+
+        self.dm.add_dependency(UnsafeInstructionResolvedKey(), self._resume_from_unsafe)
 
     def elaborate(self, platform):
         m = TModule()
 
+        with Transaction().body(m):
```
xThaid, line 69: I might be missing something, but why is this a separate transaction?

xThaid (commented):
overall looks good to me

qbojj (approved):

tilk:
Will merge after investigation of the IPC loss. #1029 is merged, Kanata logs will be availabie as an artifact.

xThaid:
I had a look at the instruction traces, and I think the difference comes down to frontend flush timing.

Before, ExceptionInformationRegister called fetch_stall_exception during the same cycle execution was reported. Now, the flush happens one cycle later, allowing extra instructions into the pipeline that the core has to wait on. This should become a non-issue with checkpointing, though

baseline
image

vs this PR

image

tilk:
Thanks for that!

awariac:
is this a reasonable decoupling or should we restore the previous behavior? (for exceptions)
I considered that reason but didn't expect the difference to be that much.

tilk:
If maintaining previous behavior for exceptions has significant costs, I think we can live with the current behavior. The checkpoint rollback path is more important.

PR #1022, opened 2026-07-17
Enum refactor

tilk:
IntEnums and IntFlags were used a lot before, even though the automatic int conversion isn't strictly necessary and might introduce bugs. Also, `shape` was often omitted, which disables Amaranth's runtime protections for the enum types. This PR changes some of the enums to stricter variants, and also removes one magic number use in favor of an enum.


coreblocks/priv/pmp.py
```diff
@@ -21,7 +21,7 @@ def __init__(self):
 
 
 @unique
-class PMPOperationMode(Enum):
+class PMPOperationMode(Enum, shape=2):
```
qbojj: This is a configuration enum - probably shouldn't have shape.

coreblocks/arch/isa_consts.py
```diff
@@ -219,22 +219,56 @@ class HPMEvent(IntEnum):
...
 @unique
-class XlenEncoding(IntEnum, shape=2):
+class XlenEncoding(Enum, shape=2):
     W32 = 1
     W64 = 2
     W128 = 3
 
+    @staticmethod
+    def from_xlen(xlen: int):
+        return XlenEncoding[f"W{xlen}"]
```
qbojj, line 229: this could also be used in csr_instances.py and rvvi.py

coreblocks/arch/isa_consts.py
qbojj, line 1: What about the rest of the enums? Shouldn't they also be pure `Enum`/`Flag` rather than `IntEnum`/`IntFlag`?

qbojj (changes requested):

coreblocks/priv/pmp.py
tilk: Fixed.

coreblocks/arch/isa_consts.py
tilk: Is there another place in csr_instances which I missed?

coreblocks/arch/isa_consts.py
tilk: Good question, probably they should, I will look into it later. Probably `Registers` should remain an `IntEnum`.

coreblocks/arch/isa_consts.py
qbojj: no. I am just blind

qbojj (approved):

coreblocks/arch/isa_consts.py
tilk: I would leave `Funct*` as intenums for now because of current hacky instruction decoding.

PR #1023, opened 2026-07-17
fetch unit: restructure fetch around the block exit point

xThaid:
The fetch unit now reads the block's prediction from the FTQ (an empty fall-through prediction for now). I also cleaned up the logic how the fetch unit computes the exit point per fetch block, from which redirect/stall decisions and the instruction mask are derived. 

Additionally, I fixed a bug in the prediction checker when a prediction is pointing at an invalid instruction slot (see the added test). It remained uncovered because currently we simply don't produce predictions.


tilk (approved):
Looks straightforward.

PR #1025, opened 2026-07-18
FTQ: prediction and training plumbing

xThaid:
The FTQ now:
- stores BPU predictions per entry (prediction_mem) and serves them to the fetch unit, 
- records branch resolutions from the jump-branch unit (train_mem), 
- and trains the BPU behind the commit pointer.

FTQ auxiliary memories are reworked into FTQReadQueue, which owns its read pointer and supports rollbacks.


coreblocks/frontend/ftq.py
```diff
@@ -49,24 +54,36 @@ def elaborate(self, platform):
         wrport = mem.write_port()
         rdport = mem.read_port(transparent_for=[wrport])
 
-        m.d.comb += rdport.addr.eq(self.read_ptr_next)
-        m.d.comb += self.read_data.eq(rdport.data)
+        read_ptr_next = FTQPtr(gen_params=self.gen_params)
```
qbojj, line 57: Why is there this dance with `read_ptr_next`? Just assign to `self.read_ptr`

coreblocks/frontend/ftq.py
```diff
@@ -49,24 +54,36 @@ def elaborate(self, platform):
...
+            return {"data": self.layout(rdport.data)}
 
         @def_method(m, self.write)
         def _(ftq_ptr, data):
             m.d.comb += wrport.addr.eq(ftq_ptr.ptr)
             m.d.comb += wrport.data.eq(data)
             m.d.comb += wrport.en.eq(1)
 
+        # Higher-indexed rollback ports override lower-indexed ones
+        @def_methods(m, self.rollback)
```
qbojj: Since you do manual arbitration anyway, `single_caller` seems reasonable (probably should do this in many more places, but that is not an issue with this pr)

qbojj (approved):

coreblocks/frontend/ftq.py
tilk: I believe the answer is three lines below. This way, the same signal both updates a register and drives a synchronous read port, guaranteeing consistency.

qbojj: Ok, too much dislexia

PR #1028, opened 2026-07-19
BPU: add micro-BTB predictor

xThaid:
Coreblocks' first branch predictor!

The micro-BTB (branch target buffer) maps a fetch block address to a predicted next-fetch PC. The lookup is fully associative and combinational over a request, so a prediction requested in one cycle is available in the next cycle.

It's "micro" because of its relatively small size - regular BTBs are on the order of thousands of entries and use SRAM.

Note that this predictor is currently hard-wired into the BPU - this is temporary, for simplicity. The end goal is to have the BPU configurable in the same way as our FUs.

Depends on #1025

### DIFF
--- a/coreblocks/cache/plru.py
+++ b/coreblocks/cache/plru.py
@@ -1,0 +1,83 @@
+from amaranth import *
+
+from transactron import Method, Methods, TModule, def_method, def_methods
+from transactron.utils.transactron_helpers import make_layout
+
+__all__ = ["TreePLRU"]
+
+
+class TreePLRU(Elaboratable):
+    """Tree-based pseudo-LRU replacement state over a power-of-two number of ways.
+
+    Tree-PLRU keeps one bit per internal node of a balanced binary tree. For a cache
+    with N ways, it requires exactly N - 1 bits. Each internal node indicates which
+    side was accessed less recently (0 - the left side).
+
+    To find the least recently used item, we traverse down the tree from the root node,
+    following the directions indicated by each bit.
+
+    When accessing an item, the bits along its specific path are flipped to point in
+    the opposite direction, so it is not selected for subsequent evictions.
+
+    Methods
+    -------
+    get_victim: Method (nonexclusive)
+        Returns the way to replace.
+    touch: Methods
+        Each method marks its `way` argument as most-recently-used. Touches are applied
+        in port order, so a higher-indexed port wins.
+    """
+
+    def __init__(self, ways: int, touch_ports: int = 1):
+        assert ways >= 2 and ways & (ways - 1) == 0, "TreePLRU ways must be a power of two"
+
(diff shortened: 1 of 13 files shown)


tilk:
New synthesis workflow confirms - there is definitely an Fmax regression here.

tilk:
I'm not sure #1054 helped or made it worse, but #1068 certainly helped.

Edit: think it's good enough for now, we should not stall any longer.

tilk (approved):

coreblocks/frontend/bpu/micro_btb.py
```diff
@@ -0,0 +1,168 @@
...
+        req_fb = Signal(self.tag_width)
+        req_valid = Signal()
+        m.d.sync += req_valid.eq(self.request.run)
+
+        def match_vec(fb: Value) -> list[Value]:
+            return [(entry.useful != 0) & (entry.tag == fb) for entry in entries]
+
+        def select_entry(matches: list[Value]):
+            entry = OneHotMux.create(m, list(zip(matches, entries)))
+            idx = OneHotMux.create(m, [(matches[i], C(i, range(self.num_entries))) for i in range(self.num_entries)])
```
qbojj, line 92: probably should be `Encoder`

coreblocks/cache/plru.py
```diff
@@ -0,0 +1,83 @@
...
+                direction = way[self.ways_log - 1 - level]
+                new[node] = Mux(valid & on_path, ~direction, cur[node])
+                walk(2 * node + 1, level + 1, on_path & ~direction)
+                walk(2 * node + 2, level + 1, on_path & direction)
+
+            walk(0, 0, C(1, 1))
+            return new
+
+        # Apply each touch in port order; a higher-indexed port wins on shared nodes
+        state = {node: tree[node] for node in range(self.ways - 1)}
```
qbojj: this is just a list

coreblocks/frontend/bpu/micro_btb.py
qbojj, line 1: All of this could be implemented without that many one-hot to/from binary round trips, if the interface with PLRU be on one-hot signals (victim and maybe also touch).

Here it probably wouldn't really matter, as tree plru operates internally on binary representations.

EDIT: This is more of a thought than anything else. You can close this if you think it is not worth the effort

qbojj (changes requested):

coreblocks/frontend/bpu/micro_btb.py
tilk: Current implementation of `Encoder` in Transactron is borrowed from Amaranth and is probably not great for larger bit widths.

tilk:
To not delay further, I'm merging without review status change.

PR #1029, opened 2026-07-20
benchmarks: upload Kanata pipeline logs as an artifact

xThaid:
Having pipeline logs generated for every benchmark run will be useful for answering questions like "why did our IPC drop by 0.003?". Konata can overlay two pipelines, so we will be able to display the trace from master together with the one from the branch and quickly find the point where they diverge. This won't tell us why exactly the regression happens, but it will at least let us narrow the issue down.

A compressed kanata file from a single benchmark is on the order of ~5MB, so I set an aggressive retention policy of 3 days.


tilk:
I downloaded the viewer from https://github.com/shioyadan/Konata and it doesn't see the files unless their names end with `.out`.

tilk (approved):

PR #1030, opened 2026-07-23
RS entries 2->4

tilk:
RS entries were reduced to 2 because superscalarity caused loading a lot of instructions into the core, which were then often flushed because of poor predictions and IPC dropped. Konata logs showed however that RS size 2 leads to weird slowdowns, which are caused by 2 cycle delay between inserting and removing instruction from RS. Here I'm experimentally changing the RS size to a weird number of 3, to see how that impacts IPC.

Edit: Well, looks like this should wait for checkpointing, predictors, or both.

Edit 2: Now it looks like we have 3 tests faster, 2 tests slower and 3 around the same, and the gains are pretty nice. I suppose let's finally merge, the losses can probably be reclaimed by better checkpointing or better predictors.


tilk:
Increasing RS size still hurts performance :(

awariac:
how it looks on K[a/o]nata?

tilk:
> how it looks on K[a/o]nata?

Exactly as expected. Artificial stalls because of slow RS turnaround are gone:

obraz

But misprediction recovery can be slower:

obraz

Edit: Huh, this looks like it's the fault of the fetcher?!?
Edit 2: Now I'm not so sure it's just the fetcher, here is the bigger picture for the same part:

obraz

If lined up to previous misprediction, instructions just after it linger in the core longer. This just might be some checkpointing limitation.

tilk:
@xThaid Could you comment on this, if you have the time?

xThaid (approved):
I think this can be merged as is, but let's open an issue to investigate the perf regression

PR #1032, opened 2026-07-25
change all FIFO to BasicFifo

qbojj:
removes all instances of `transactron.FIFO` in favor of `transactron.BasicFifo`.

### DIFF
--- a/coreblocks/func_blocks/fu/mul_unit.py
+++ b/coreblocks/func_blocks/fu/mul_unit.py
@@ -103,7 +103,7 @@ class MulUnit(FuncUnitBase[MulFn]):
     def elaborate(self, platform):
         m = super().elaborate(platform)
 
-        m.submodules.params_fifo = params_fifo = FIFO(
+        m.submodules.params_fifo = params_fifo = BasicFifo(
             [
                 ("rob_id", self.gen_params.rob_entries_bits),
                 ("rp_dst", self.gen_params.phys_regs_bits),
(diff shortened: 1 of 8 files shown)


coreblocks/func_blocks/fu/mul_unit.py
awariac, line 112: for other PR - maybe we should migrate to standard lib.data `StructLayout`s / via `make_layout`?

This was other temporary solution for old amaranth records, I would think about deprecating layoutlists in transactron data structures.

awariac (approved):

coreblocks/func_blocks/fu/mul_unit.py
qbojj: Do you mean only for `transactron.lib.*`, or for all places where `LayoutIterable` could have been used - aka all places which use `MethodLayout`, which includes `i` and `o` fields of `Method`?

awariac: for everything that goes out to transactron.
Ex. for `layouts.py` the current representation looks like a reasonable shortcut and is more readable, so it makes sense to use it internally

tilk: Layout lists in `i`/`o` are reasonable shorthands, maybe dict notation should be preferred because this is what Amaranth does. I need to think about this.

tilk:
Huh, BasicFifo uses slightly more FFs.

PR #1034, opened 2026-07-28
update embench march

qbojj:


tilk:
I believe the toolchain might be too old.

qbojj:
Yeah, works with riscv-toolchain:latest

PR #1037, opened 2026-07-28
use always_body whenever possible

qbojj:


tilk (approved):
Wow, that's a lot of them.

PR #1038, opened 2026-08-01
Fix wave tracing for cocotb

qbojj:
Traces with verilator according to [cocotb documentation](https://docs.cocotb.org/en/stable/simulator_support.html#sim-verilator-waveforms) for FST should include "--trace --trace-fst --trace-structs" to extra args rather than "--trace-fst --trace-structs".

This also makes wave tracing not require any changes in the verilog file

### DIFF
--- a/test/regression/cocotb/arch_test.Makefile
+++ b/test/regression/cocotb/arch_test.Makefile
@@ -1,25 +1,3 @@
-# Makefile for arch regression cocotb tests
-
-# defaults
-SIM ?= verilator
-TOPLEVEL_LANG ?= verilog
-SIM_BUILD ?= build/riscv-arch-test/$(SIM)
-
-# TOPLEVEL is the name of the toplevel module in your Verilog or VHDL file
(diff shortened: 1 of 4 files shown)


awariac:
is it? I think `--trace` was depracated way of `--trace-vcd` which is an alternative of `--trace-fst`.

awariac:
oh it is for cocotb. i see why the generated cpp issue was there now!

qbojj:
> is it? I think `--trace` was depracated way of `--trace-vcd` which is an alternative of `--trace-fst`.

Technically should be
```makefile
COMPILE_ARGS += --trace-fst --trace-structs
SIM_ARGS += --trace
```
but the docs said `EXTRA_ARGS += --trace --trace-fst --trace-structs`, so I did so

tilk (approved):

tilk:
Looks like CI is not broken, and it's worth it to merge this just for the makefile unification.

PR #1047, opened 2026-08-06
Update riscv arch test

qbojj:
riscv-arch-tests has seen quite some commits in the meantime, so it's time to update.

Things done here:
- update riscv-arch-tests
- update DUT config, sail.json, and link script based on current state of riscv-arch-tests
- make riscv-arch-tests responsible for test skipping, as before some tests that were ok were skipped because some other test in the same group
- rechecked the failing tests (there are quite a bit less now).


tilk:
Fix formatting.

test/regression/test_arch_regression.py
```diff
@@ -170,6 +170,8 @@ async def run_arch_elf(sim_backend, elf_path: str | Path, timeout_cycles: int =
         elf_path,
         sim_backend.stop,
         do_workarounds=False,
+        disable_write_protection=True,
+        force_executable=True,
     )
```
tilk, line 176: Do all (most?) tests do things that require looser protection?

qbojj: all* priv tests require executable data partition (due to T-SBI CSR emulation logic), while only (?) Zifencei requires text.* sections writeable.

tilk (approved):

tilk:
Huh? Stopped working after merging.

PR #1050, opened 2026-08-07
Checkpoint part 2: this time actually

awariac:
Introduces checkpointing to the core (actually this time :tada:), based on a previous rewrite #1021

Changes:
- Inactive instruction handling in Retirement
- Skips for side effects on inactive instructions in FUs
- Use rollbacks from JB unit

Some improvents are left for future PRs.
Linux boot is verified

### DIFF
--- a/coreblocks/backend/retirement.py
+++ b/coreblocks/backend/retirement.py
@@ -330,15 +330,15 @@ class Retirement(Elaboratable):
                             # hard flush instruction for trap handling
                             flush_instr(i, rob_entries.entries[i])
 
-                        # TODO: this is some approximation of misprediction events - looking for active -> inactive
-                        # changes in retured instructions.
-                        # More metadata needs to be stored for an accurate result, improve.
-                        active_mask_with_prev = Signal(tag_active_mask.shape().width + 1)
-                        m.d.comb += active_mask_with_prev.eq(Cat(last_retired_active, tag_active_mask & retiring_mask))
-                        change_mask = active_mask_with_prev & ((~active_mask_with_prev) >> 1)
-                        with m.If((change_mask & retiring_mask).any()):  # without last bit
-                            self.perf_mispredictions.incr(m)
-                            m_csr.hpm_event_report(m, events=1 << HPMEvent.BRANCH_MISPREDICTION)
+                    # TODO: this is some approximation of misprediction events - looking for active -> inactive
+                    # changes in retured instructions.
+                    # More metadata needs to be stored for an accurate result, improve.
+                    active_mask_with_prev = Signal(tag_active_mask.shape().width + 1)
+                    m.d.comb += active_mask_with_prev.eq(Cat(last_retired_active, tag_active_mask & retiring_mask))
+                    change_mask = active_mask_with_prev & ((~active_mask_with_prev) >> 1)
+                    with m.If((change_mask & retiring_mask).any()):  # without last bit
+                        self.perf_mispredictions.incr(m)
+                        m_csr.hpm_event_report(m, events=1 << HPMEvent.BRANCH_MISPREDICTION)
 
                     # Commit the FTQ entry for the last retired instruction this cycle.
                     with m.If(last_commit_ftq_ptr_v):
--- a/coreblocks/backend/retirement.py
+++ b/coreblocks/backend/retirement.py
@@ -421,9 +421,8 @@ class Retirement(Elaboratable):
             self.side_fx_guard,
             ready=~core_flushing,
             validate_arguments=lambda rob_id, tag, require_done: (rob_id == side_fx_rob_id)
-            & (
-                ~require_done | (self.pure_count == done_count) & active_tags[tag]
-            ),  # FUTURE-TODO: inactive instruction are pure
+            & (~require_done | (self.pure_count == done_count))
+            & active_tags[tag],  # FUTURE-TODO: inactive instruction are pure
             nonexclusive=True,
             combiner=lambda m, args, runs: {"rob_id": 0, "tag": 0, "require_done": 0},
         )
--- a/coreblocks/backend/retirement.py
+++ b/coreblocks/backend/retirement.py
@@ -1,6 +1,6 @@
 from amaranth import *
 from amaranth.lib.data import View
-from transactron.utils import count_trailing_zeros, or_value
+from transactron.utils import HardwareLogger, count_trailing_zeros, or_value
 from coreblocks.interface.layouts import (
     CoreInstructionCounterLayouts,
     ExceptionInformationRegisterLayouts,
@@ -24,6 +24,7 @@ from coreblocks.params.genparams import GenParams
 from coreblocks.arch import ExceptionCause, HPMEvent, PrivilegeLevel
 from coreblocks.arch.csr_address import CounterEnableFieldOffsets
 from coreblocks.interface.keys import (
+    ActiveTagsKey,
     CoreStateKey,
     CSRInstancesKey,
     RVVIHartCollectorKey,
@@ -39,14 +40,12 @@ from coreblocks.arch.isa_consts import TrapVectorMode
 __all__ = ["Retirement"]
 
 
+log = HardwareLogger("backend.retirement")
 evlog = EventSource("backend.retirement")
 
 
 class Retirement(Elaboratable):
-    def __init__(
-        self,
-        gen_params: GenParams,
-    ):
+    def __init__(self, gen_params: GenParams):
         self.gen_params = gen_params
         self.rob_peek = Method(o=gen_params.get(ROBLayouts).peek_layout)
         self.rob_retire = Method(i=gen_params.get(ROBLayouts).retire_layout)
@@ -72,7 +71,6 @@ class Retirement(Elaboratable):
         interrupt_controller_layouts = gen_params.get(InternalInterruptControllerLayouts)
         self.async_interrupt_cause = Method(o=interrupt_controller_layouts.interrupt_cause)
         self.checkpoint_tag_free = Method()
-        self.checkpoint_get_active_tags = Method(o=gen_params.get(RATLayouts).get_active_tags_out)
 
         self.pure_count = Signal(range(gen_params.retirement_superscalarity + 1))
         self.instret_csr = CSRRegister(None, gen_params, width=64, fu_read_map=lambda _, v: v + self.pure_count)
@@ -141,27 +139,51 @@ class Retirement(Elaboratable):
 
             self.perf_instr_ret.incr[i](m)
 
+            log.debug(
+                m,
+                True,
+                "retiring instruction rl{} -> rp{} freeing rp{}",
+                rob_entry.rob_data.rl_dst,
+                rob_entry.rob_data.rp_dst,
+                rat_out.old_rp_dst,
+            )
+
         def flush_instr(i: int, rob_entry: View):
             evlog.emit(m, RobFlush.hw(rob_id=rob_entry.rob_id))
 
-            # free the "new" instruction rp_dst - result is flushed
+            # free the "new" instruction rp_dst - result is discarded
             free_phys_reg(i, rob_entry.rob_data.rp_dst)
+
+            log.debug(
+                m,
+                True,
+                "flushing instruction freeing rp{}",
+                rob_entry.rob_data.rp_dst,
+            )
 
         retire_valid = Signal()
         exception = Signal()
-        continue_pc_override = Signal()
-        continue_pc = Signal(self.gen_params.isa.xlen)
-        core_flushing = Signal()
         trap_target_priv = Signal(PrivilegeLevel, init=PrivilegeLevel.MACHINE)
         ftq_commit_ptr = FTQPtr(gen_params=self.gen_params)
+
+        active_tags = Signal(self.gen_params.tag_count)
+        last_retired_tag = Signal(self.gen_params.tag_bits)
+        next_last_retired_tag = Signal.like(last_retired_tag)
 
         retire_count = Signal(range(self.gen_params.retirement_superscalarity + 1))
         no_trap_count = Signal.like(retire_count)
         done_count = Signal.like(retire_count)
         tag_incr_mask = Signal(self.gen_params.retirement_superscalarity)
-        safe_mask = Signal.like(tag_incr_mask)
+        retiring_mask = Signal.like(tag_incr_mask)
+        tag_active_mask_prefix = Signal.like(tag_incr_mask)
+        tag_active_mask_suffix = Signal.like(tag_incr_mask)
+        tag_active_mask = Signal.like(tag_incr_mask)
         done_mask = Signal.like(tag_incr_mask)
-        free_checkpoint = Signal()
+        first_tag_incr_pos = Signal.like(retire_count)
+        free_tag = Signal()
+
+        with Transaction().body(m):
+            active_tags = self.dependency_manager.get_dependency(ActiveTagsKey())(m).active_tags
 
         with Transaction().body(m):
             rob_entries = self.rob_peek(m)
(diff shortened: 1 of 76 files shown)


coreblocks/backend/retirement.py
```diff
@@ -141,27 +139,52 @@ def retire_instr(i: int, rob_entry: View):
...
+        tag_active_mask_prefix = Signal.like(tag_incr_mask)
+        tag_active_mask_suffix = Signal.like(tag_incr_mask)
+        tag_active_mask = Signal.like(tag_incr_mask)
         done_mask = Signal.like(tag_incr_mask)
-        free_checkpoint = Signal()
+        first_tag_incr_pos = Signal.like(retire_count)
+        free_tag = Signal()
+        last_retired_active = Signal()
+
+        with Transaction().body(m):
```
qbojj:
```suggestion
        with Transaction().always_body(m):
```

coreblocks/func_blocks/fu/jumpbranch.py
```diff
@@ -147,12 +151,21 @@ def elaborate(self, platform):
         )
         m.submodules.instr_fifo = instr_fifo = BasicFifo(instr_fifo_layout, 2)
 
+        with Transaction().body(m):
```
qbojj:
```suggestion
        with Transaction().always_body(m):
```

coreblocks/backend/retirement.py
```diff
@@ -376,20 +422,24 @@ def _():
...
+        # provided.
         @def_method(
             m,
             self.side_fx_guard,
             ready=~core_flushing,
-            validate_arguments=lambda rob_id, require_done: (rob_id == side_fx_rob_id)
-            & (~require_done | (self.pure_count == done_count)),
+            validate_arguments=lambda rob_id, tag, require_done: (rob_id == side_fx_rob_id)
+            & (~require_done | (self.pure_count == done_count))
+            & active_tags[tag],  # FUTURE-TODO: inactive instruction are pure
```
qbojj, line 438:
```suggestion
            & active_tags[tag],  # FUTURE-TODO: inactive instructions are pure
```

coreblocks/func_blocks/fu/jumpbranch.py
```diff
@@ -200,18 +213,15 @@ def elaborate(self, platform):
...
-                m.d.comb += exception.eq(1)
-                self.exception_report(
-                    m,
-                    cause=ExceptionCause._COREBLOCKS_MISPREDICTION,
-                    rob_id=instr.rob_id,
-                    tag=instr.tag,
-                    pc=jump_result,
-                    mtval=0,
-                )
+                with m.If(active_tags[instr.tag] & ~core_state.flushing):
```
qbojj, line 223: This seems to be recurring not only in this FU (e.g CSR, Priv). Maybe some function for this would be better

coreblocks/backend/retirement.py
```diff
@@ -171,13 +194,27 @@ def flush_instr(i: int, rob_entry: View):
...
+            m.d.av_comb += tag_active_mask.eq(tag_active_mask_suffix | tag_active_mask_prefix)
 
             exception_bits = Signal(self.gen_params.retirement_superscalarity)
-            m.d.av_comb += exception_bits.eq(Cat(rob_entry.exception for rob_entry in rob_entries.entries))
-            m.d.av_comb += no_trap_count.eq(count_trailing_zeros(exception_bits | safe_mask))
+            m.d.av_comb += exception_bits.eq(
+                Cat(rob_entry.exception for rob_entry in rob_entries.entries) & tag_active_mask
+            )
+            m.d.av_comb += no_trap_count.eq(count_trailing_zeros(exception_bits | ~retiring_mask))
             m.d.av_comb += exception.eq(no_trap_count != retire_count)
```
qbojj, line 218: maybe: `exception.eq((exception_bits & retiring_mask).any())`, as `retiring_mask` is a full 1-suffix.

qbojj (changes requested):
Looks good after a quick look. Thank you for your work. I will look at it again at later time, maybe than something will come up.

tilk:
Full review will come later. But for now, do you know where these holes in Kanata logs come from?

image

coreblocks/func_blocks/fu/jumpbranch.py
```diff
@@ -200,18 +213,15 @@ def elaborate(self, platform):
                     pc=jump_result,
                     mtval=0,
                 )
+            with m.Elif(is_jalr):
+                # JALR stalls the fetch (with unsafe reason) and doesn't create checkpoint.
```
xThaid, line 217: I think we shouldn't make this assumption. This won't hold once we have even a very simple branch predictor (soon!).

coreblocks/frontend/frontend.py
```diff
@@ -171,12 +171,21 @@ def elaborate(self, platform):
...
+
         @def_method(m, self.redirect)
         def _(ftq_ptr, pc):
             flush(m)
             self.ftq.backend_redirect(m, ftq_ptr=ftq_ptr, pc=pc)
             self.stall_ctrl.on_redirect_frontend(m)
 
+        @def_method(m, rollback)
+        def _(tag, pc, ftq_ptr):
+            flush(m)
```
xThaid, line 185: could we call `redirect` here?

coreblocks/backend/retirement.py
```diff
@@ -273,29 +299,49 @@ def set_trap_csrs(cause_reg, epc_reg, tval_reg):
...
+                            # hard flush instruction for trap handling
+                            flush_instr(i, rob_entries.entries[i])
+
+                    # TODO: this is some approximation of misprediction events - looking for active -> inactive
+                    # changes in retured instructions.
+                    # More metadata needs to be stored for an accurate result, improve.
+                    active_mask_with_prev = Signal(tag_active_mask.shape().width + 1)
+                    m.d.comb += active_mask_with_prev.eq(Cat(last_retired_active, tag_active_mask & retiring_mask))
+                    change_mask = active_mask_with_prev & ((~active_mask_with_prev) >> 1)
+                    with m.If((change_mask & retiring_mask).any()):  # without last bit
```
xThaid, line 339: I don't fully understand this yet, but yeah, the misprediction count is quite off. For example crc32 has now 0.048 MPKI vs 25.08 baseline

coreblocks/backend/retirement.py
qbojj, line 298: instret should only count active instructions

coreblocks/backend/retirement.py
```diff
@@ -72,7 +71,6 @@ def __init__(
         interrupt_controller_layouts = gen_params.get(InternalInterruptControllerLayouts)
         self.async_interrupt_cause = Method(o=interrupt_controller_layouts.interrupt_cause)
         self.checkpoint_tag_free = Method()
-        self.checkpoint_get_active_tags = Method(o=gen_params.get(RATLayouts).get_active_tags_out)
 
         self.pure_count = Signal(range(gen_params.retirement_superscalarity + 1))
         self.instret_csr = CSRRegister(None, gen_params, width=64, fu_read_map=lambda _, v: v + self.pure_count)
```
qbojj, line 76: instret should only count active instructions

qbojj (changes requested):

qbojj:
I think that is all from me - will approve once comments get resolved.

coreblocks/func_blocks/fu/lsu/dummyLsu.py
```diff
@@ -104,11 +100,16 @@ def _(arg):
...
+            core_state = self.dependency_manager.get_dependency(CoreStateKey())(m)
+            active_tags = self.dependency_manager.get_dependency(ActiveTagsKey())(m).active_tags
+
         # Issues load/store requests when the instruction is known, is a LOAD/STORE, and just before commit.
         # Memory loads can be issued speculatively.
         pmas = pma_checker.result
         can_reorder = is_load & ~pmas["mmio"]
-        want_issue = rob_id_match | can_reorder
+        want_issue = request_side_fx | can_reorder
+        flush = core_state.flushing | ~active_tags[request_tag]
```
tilk, line 112: Flush is used in two places, de-Signal-ing it was a wrong choice.

coreblocks/backend/retirement.py
```diff
@@ -141,27 +139,52 @@ def retire_instr(i: int, rob_entry: View):
...
+        tag_active_mask_suffix = Signal.like(tag_incr_mask)
+        tag_active_mask = Signal.like(tag_incr_mask)
         done_mask = Signal.like(tag_incr_mask)
-        free_checkpoint = Signal()
+        first_tag_incr_pos = Signal.like(retire_count)
+        free_tag = Signal()
+        last_retired_active = Signal()
+
+        with Transaction().always_body(m):
+            active_tags = self.dependency_manager.get_dependency(ActiveTagsKey())(m).active_tags
```
tilk, line 187: The `active_tags` local variable is reassigned here.  Also, why another `always_body`?

test/frontend/test_frontend.py
```diff
@@ -201,6 +201,8 @@ async def ftq_commit_process(self, sim):
 
             await self.ftq_commit.call(sim, ftq_ptr=self.ftq_commit_queue.pop())
 
+    # TODO: Improve to test rollbakcs and RolbackTagger. This unfortunatelly requires to simulate full misprediction
```
tilk, line 204:
```suggestion
    # TODO: Improve to test rollbacks and RollbackTagger. This unfortunately requires to simulate full misprediction
```

coreblocks/backend/retirement.py
tilk: If possible, add a test that triggers the issue before fixing. Precise instret is pretty tricky.

coreblocks/backend/retirement.py
```diff
@@ -273,29 +299,49 @@ def set_trap_csrs(cause_reg, epc_reg, tval_reg):
...
-                                rvvi.finalize_retire[i](
-                                    m,
-                                    rob_id=entry.rob_id,
-                                    rl_dst=entry.rob_data.rl_dst,
-                                    rp_dst=entry.rob_data.rp_dst,
-                                    trap=entry.exception & arch_trap,
-                                    interrupt=entry.exception
-                                    & (cause_register.cause == ExceptionCause._COREBLOCKS_ASYNC_INTERRUPT),
-                                )
+                            with m.If(tag_active_mask.bit_select(i, 1)):
```
tilk, line 307: `i` is constant:
```suggestion
                            with m.If(tag_active_mask[i]):
```

coreblocks/backend/retirement.py
```diff
@@ -141,27 +139,52 @@ def retire_instr(i: int, rob_entry: View):
...
+        next_last_retired_tag = Signal.like(last_retired_tag)
+
         retire_count = Signal(range(self.gen_params.retirement_superscalarity + 1))
         no_trap_count = Signal.like(retire_count)
         done_count = Signal.like(retire_count)
         tag_incr_mask = Signal(self.gen_params.retirement_superscalarity)
-        safe_mask = Signal.like(tag_incr_mask)
+        retiring_mask = Signal.like(tag_incr_mask)
+        tag_active_mask_prefix = Signal.like(tag_incr_mask)
+        tag_active_mask_suffix = Signal.like(tag_incr_mask)
```
tilk, line 179: These two are only used to construct `tag_active_mask`. They probably don't need to be `Signals`.

coreblocks/backend/retirement.py
tilk: I believe that this is a consequence of JALRs no longer being counted, as they are here just a kind of an unsafe instruction and do not trigger tag inactivation.

coreblocks/backend/retirement.py
```diff
@@ -273,29 +299,49 @@ def set_trap_csrs(cause_reg, epc_reg, tval_reg):
...
-                            flush_instr(i, entry)
+                            # hard flush instruction for trap handling
+                            flush_instr(i, rob_entries.entries[i])
+
+                    # TODO: this is some approximation of misprediction events - looking for active -> inactive
+                    # changes in retured instructions.
+                    # More metadata needs to be stored for an accurate result, improve.
+                    active_mask_with_prev = Signal(tag_active_mask.shape().width + 1)
+                    m.d.comb += active_mask_with_prev.eq(Cat(last_retired_active, tag_active_mask & retiring_mask))
+                    change_mask = active_mask_with_prev & ((~active_mask_with_prev) >> 1)
```
tilk, line 338: Given the use, this looks like a contrived way to write:

```suggestion
                    m.d.comb += active_mask_with_prev.eq(Cat(last_retired_active, tag_active_mask & retiring_mask))
                    change_mask = active_mask_with_prev & ~tag_active_mask
```

coreblocks/backend/retirement.py
```diff
@@ -273,29 +299,49 @@ def set_trap_csrs(cause_reg, epc_reg, tval_reg):
...
+                                flush_instr(i, rob_entries.entries[i])
+                                m.d.sync += last_retired_active.eq(0)
 
                         with m.Elif(i < retire_count):
-                            flush_instr(i, entry)
+                            # hard flush instruction for trap handling
+                            flush_instr(i, rob_entries.entries[i])
+
+                    # TODO: this is some approximation of misprediction events - looking for active -> inactive
+                    # changes in retured instructions.
```
tilk, line 334:
```suggestion
                    # changes in retired instructions.
```

coreblocks/backend/retirement.py
```diff
@@ -171,13 +194,27 @@ def flush_instr(i: int, rob_entry: View):
             m.d.av_comb += tag_incr_mask.eq(Cat(entry.rob_data.tag_increment for entry in rob_entries.entries))
             m.d.av_comb += done_mask.eq(Cat(entry.done for entry in rob_entries.entries))
             m.d.av_comb += done_count.eq(count_trailing_zeros(~done_mask))
-            m.d.av_comb += safe_mask.eq(tag_incr_mask & (tag_incr_mask - 1) | (-1 << done_count))
-            m.d.av_comb += retire_count.eq(count_trailing_zeros(safe_mask))
-            m.d.av_comb += free_checkpoint.eq(safe_mask != (tag_incr_mask | (-1 << done_count))[: len(safe_mask)])
+            limiting_instruction_mask = (tag_incr_mask & (tag_incr_mask - 1)) | (-1 << done_count)
```
tilk, line 197: Make it a signal again, this is an argument to `count_trailing_zeros` which uses its argument multiple times.

coreblocks/backend/retirement.py
```diff
@@ -171,13 +194,27 @@ def flush_instr(i: int, rob_entry: View):
...
+
+            m.d.av_comb += retire_count.eq(count_trailing_zeros(limiting_instruction_mask))
+            m.d.av_comb += retiring_mask.eq(~(-1 << retire_count))
+            m.d.av_comb += free_tag.eq((tag_incr_mask & retiring_mask).any())
+
+            m.d.av_comb += first_tag_incr_pos.eq(count_trailing_zeros(tag_incr_mask | (-1 << done_count)))
```
tilk, line 203: This didn't impact Fmax (probably because the superscalarity level is just 2 currently), but the round-trip bit mask -> count -> bit mask again worries me. Maybe using the classic bit-smearing trick `b | -b` would work better in the long run for the same reason why the `OneHotMux` trick works.

tilk (commented):
Overall, looks like a great PR, congratulations! Some minor comments below.

coreblocks/backend/retirement.py
awariac: I also thought this is the case, but I want to verify that.
Wasn't crc32 the test with simple loop with easily predictable branches? JALRs are technically not mispredictions currently so the count should be OK (for now).

coreblocks/func_blocks/fu/jumpbranch.py
awariac: This assumption is also made in different places, like RollbackTagger. I would prefer to change how it works in a separate PR.

coreblocks/backend/retirement.py
awariac: Thanks for catching that.
Triggering misprediction is easy for now, but it becomes less obvious when predictors will be introduced. What would you think of introducing explicit internal "trigger misprediction" instruction for testing instead of relying on internal predictor assumptions? Forward jumps are used in some other tests as well.

coreblocks/func_blocks/fu/jumpbranch.py
awariac: moved to `active_tags` as discussed

tilk:
Something is wrong, the IPC benchmark is massively down.

awariac:
> Something is wrong, the IPC benchmark is massively down.

ahhh, it all makes sense now.
It was because the `instret` count was wrong and it was used to compute IPC.

aha-mont64 has a huge factor, because it counted a lot of mispredictions. Cycle count between changes look about the same.

awariac:
> But for now, do you know where these holes in Kanata logs come from?

Isn't that a regular cache miss? Similar gaps are present in non-checkpointed version.
It's because flush of inactive instruciton in retirement is not tracked - maybe we can change that in future, but currently it reproduces the previous behaviour.

Btw the konata waveforms are so great!!
example of checkpointing in action:
image

qbojj (commented):
Seems ok, just need to fix the tests. I'm not sure about the riscv-arch-test though. It fails on PMP tests with Sv32 - could be unrelated and just this PR exposed the issue.

awariac:
@qbojj I'm afraid this might be the case.

Do you have a time to maybe help me debug it?
Trace from S-mode PMP test https://nc.kuznia-rdzeni.org/s/KCLiLoxwoqXLqAW

What I found for now:
1. There is a valid exception that flushes the core, and resuming the fetch to trap handler from retirement is unsuccessful
2. `fetch_address_unit` provides correct addresses, `FTQ_Alloc` is the last point that runs.
3. `FTQ_Send_Fetch_Requests` doesn't run, but it should, `stall_guard` is ready. Requests doesn't reach `Fetch`.

qbojj:
jet another TOCTOU :(
address translator decides from which channel (either TLB or forwarder) to get data from based on current priv mode. But current priv mode can change between request and accept.

qbojj:
#1062 ~should do~ does the trick.

test/scheduler/test_checkpointing.py
```diff
@@ -160,7 +160,7 @@ async def rob_retire_process(sim):
                 if entry["tag_increment"]:
                     await dut.free_tag.call(sim)
 
-        @def_method_mock(lambda: dut.core_state, nonexclusive=True)
+        @def_method_mock(lambda: dut.core_state)
```
tilk, line 163: Huh?

tilk (approved):

qbojj (approved):

PR #1054, opened 2026-08-08
Timing optimizations in the prediction checker

xThaid:
Currently, a big chunk of our critical path is in the prediction checker (this is further aggravated by the upcoming branch prediction changes). I attempted a few micro optimizations - mostly replace arrays with OneHotMux.

This should be a no-op functionally.

### DIFF
--- a/coreblocks/frontend/fetch/fetch.py
+++ b/coreblocks/frontend/fetch/fetch.py
@@ -669,6 +669,7 @@ class PredictionChecker(Elaboratable):
         m = TModule()
 
         params = self.gen_params.get(FrontendParams)
+        layouts = self.gen_params.get(FetchLayouts)
 
         m.submodules += [
             self.perf_mispredicted_cfi_type,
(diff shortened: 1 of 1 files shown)


xThaid:
OK, it seems that this change didn't affect Fmax meaningfully, but nevertheless, I think it's an improvement

coreblocks/frontend/fetch/fetch.py
```diff
@@ -694,26 +696,49 @@ def _(fb_addr, starts_mid_instr, instr_valid, predecoded, prediction):
...
+            # m.d.comb
+            m.submodules.pd_redirect_mux = pd_redirect_mux = OneHotMux(
+                layouts.predecoded_instr, self.gen_params.fetch_width, priority=True, has_default=False
+            )
+            m.d.av_comb += pd_redirect_mux.select.eq(pd_redirect_mask)
+            for i in range(self.gen_params.fetch_width):
+                m.d.av_comb += Value.cast(pd_redirect_mux.inputs[i]).eq(Value.cast(predecoded[i]))
+
+            pd_redirect_instr = Signal(layouts.predecoded_instr)
+            m.d.av_comb += Value.cast(pd_redirect_instr).eq(Value.cast(pd_redirect_mux.output))
```
qbojj: I think plain `.eq` would be enough, as there are no plain `Value`s
```suggestion
            for i in range(self.gen_params.fetch_width):
                m.d.av_comb += pd_redirect_mux.inputs[i].eq(predecoded[i])

            pd_redirect_instr = Signal(layouts.predecoded_instr)
            m.d.av_comb += pd_redirect_instr.eq(pd_redirect_mux.output)
```

or even
```suggestion
            pd_redirect_instr = Signal(layouts.predecoded_instr)
            m.d.av_comb += pd_redirect_mux.inputs.eq(predecoded)
            m.d.av_comb += pd_redirect_instr.eq(pd_redirect_mux.output)
```

needs to be tested though

coreblocks/frontend/fetch/fetch.py
```diff
@@ -737,27 +762,26 @@ def get_decoded_target_for(idx: Value) -> Value:
...
                         "cfi_target": decoded_target_for_decoded_cfi,
                     },
                 )
             with m.Elif(mispredicted_cfi_type):
                 self.perf_mispredicted_cfi_type.incr(m, prediction.cfi_type)
-                # If no decoded instruction redirects (pd_redirection_enc.n), this is a
-                # fall-through resteer: cfi_type is INVALID and cfi_idx/cfi_target are
-                # meaningless.
+                # If no decoded instruction redirects, this is a fall-through resteer: the
+                # one-hot mux outputs CfiType.INVALID and cfi_idx/cfi_target are meaningless.
```
qbojj, line 790: this relies on implementation detail of `one_hot_mux` - you can use explicit default instead.

qbojj (changes requested):

coreblocks/frontend/fetch/fetch.py
xThaid: I think this is semi-documented
`OneHotMux.create`:
```
        default_input: ValueLike, optional
            Default input. If not provided, the multiplexer will set the output to 0 vector when all select bits are 0.
```

I changed it to use the default value explicitly anyway.

qbojj (approved):

PR #1057, opened 2026-08-10
Handle wake-up deps

### LINKED ISSUE #1056: Add wake-up dependency tracking to Kanata graphs
The Kanata format supports wake-up dependencies. Adding them would allow to better judge the effect of instruction latencies and how well out-of-order execution goes around them.

tilk:
An approach to adding register dependencies to Kanata logs. Work in progress.

Fixes #1056.

### DIFF
--- a/coreblocks/telemetry/konata.py
+++ b/coreblocks/telemetry/konata.py
@@ -61,7 +61,7 @@ class KonataParser(EventConsumer):
     - "Rn" (rename) from entering the scheduler,
     - "Ds" (dispatch) from ROB allocation (which is where the
       instruction becomes identified by its ROB id),
-    - "Is" (issue) from being issued to a functional unit,
+    - "IX" (issue) from being issued to a functional unit,
     - "Cm" once execution completes.
 
     The instruction terminates by retiring (`RobRetire`) or being squashed.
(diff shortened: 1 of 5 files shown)


tilk:
This still generates errors. I think I'll take a more direct approach instead and do precise wake-up tracking.

tilk:
And now we can see cases where out-of-order superscalar execution allows to parallelize and go around latencies.

obraz

qbojj (approved):

coreblocks/telemetry/konata.py
```diff
@@ -205,6 +225,9 @@ def on_rob_flush(self, rec: DecodedEvent):
             return
         self._command(rec.cycle, "R", insn_id, 0, 1)
         self.terminated.add(insn_id)
+        if insn_id in self.rp_dst:
+            del self.by_rp_dst[self.rp_dst[insn_id]]
+            del self.rp_dst[insn_id]
```
awariac, line 230: i might miss something but this is not freed in case of `R` by `FTQRollback`

tilk: I believe there is nothing to free there. These follow physical registers, which are only deallocated in the retirement, together with `RobFlush` and `RobRetire` events.

PR #1058, opened 2026-08-12
make SSI edge triggered, as defined in ISA

qbojj:
> "norm:sip_ssip_sie_ssie": "Bits sip.SSIP and sie.SSIE are the interrupt-pending and interrupt-enable bits for supervisor-level software interrupts. If implemented, SSIP is writable in sip and may also be set to 1 by a platform-specific interrupt controller.",

This PR sets the `SSI` to be edge triggered, as defined by spec - this naturally makes `SSIP` writeable in `sip`

### DIFF
--- a/coreblocks/priv/traps/interrupt_controller.py
+++ b/coreblocks/priv/traps/interrupt_controller.py
@@ -96,7 +96,7 @@ class InternalInterruptController(Component):
         # mip_meip_rdonly
         # mip_mtip_rdonly
         # mip_msip_rdonly
-        self.mip_writeable = self.edge_reported_mask & custom_interrupt_mask
+        self.mip_writeable = self.mie_writeable & self.edge_reported_mask
 
         if gen_params.supervisor_mode:
             # mip_stip_no_stimecmp_acc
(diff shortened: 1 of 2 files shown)


awariac:
> writable in sip and may also be set to 1 by a platform-specific interrupt controller

huh, this is a very non-obvious way to word that.
If anyone else is wondering this is in contrast spec about MIP "mip, and is set and cleared by a platform-specific interrupt controller."

coreblocks/priv/traps/interrupt_controller.py
```diff
@@ -81,17 +84,19 @@ def __init__(self, gen_params: GenParams):
...
             | (1 << InterruptCauseNumber.MEI)
-            | (((1 << gen_params.interrupt_custom_count) - 1) << 16)
+            | custom_interrupt_mask
         )
 
         # mip_meip_rdonly
         # mip_mtip_rdonly
         # mip_msip_rdonly
-        self.mip_writeable = self.edge_reported_mask
+        self.mip_writeable = self.edge_reported_mask & custom_interrupt_mask
```
qbojj: mirroring `sip_writeable`
```suggestion
        self.mip_writeable = self.mie_writeable & self.edge_reported_mask
```

awariac (approved):

PR #1059, opened 2026-08-12
Add more HardwareLogs

awariac:
Add more debug logs for places that I often look when debugging regression tests in traces or add them every time again.

@qbojj iirc did you say you had more places to put logs in?


qbojj:
> @qbojj iirc did you say you had more places to put logs in?

I think you have found all of the places I have been using.

qbojj (approved):

PR #1062, opened 2026-08-16
fix addr translation lockup

qbojj:
Fixes the TOCTOU on `priv_mode` in addr translator and tlb permission miss incorrectly being applied when entry was not hit (e.g on access faults)


awariac (approved):
Thank you very much for help!
Looks like a clean fix

PR #1065, opened 2026-08-17
great cocotb unification

qbojj:
make all `test/regression` tests use the same verilated module.

All tests have now:
- memory from 0x80000000
- socs peripherials from 0xE0000000
- test peripherials from 0xF0000000

1. Moved benchmark and riscv-tests to-host to 0xF0000000 (either way those should be in MMIO region).
1. Start pc is now 0x80000000.
1. Removed the "SKIP" logic by adding a common skip module
1. Made cocotb configuration in python modules rather than in the makefiles

### DIFF
--- a/test/regression/cocotb.py
+++ b/test/regression/cocotb.py
@@ -370,6 +370,7 @@ def run_cocotb_entrypoint(
 
     env = os.environ.copy()
     env["PATH"] = str(TEST_ROOT.resolve()) + os.pathsep + env["PATH"]
+    env["SIM"] = os.environ.get("SIM", "verilator")
     if additional_env is not None:
         env.update(additional_env)
 
--- a/test/regression/cocotb.py
+++ b/test/regression/cocotb.py
@@ -11,6 +11,7 @@ import subprocess
 import tempfile
 import sys
 import xml.etree.ElementTree as eT
+import argparse
 
 import cocotb
 from cocotb.clock import Clock, Timer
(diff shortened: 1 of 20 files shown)


qbojj:
I think we need to migrate to [cocotb runners](https://docs.cocotb.org/en/stable/runner.html) if we want artifact that can be run locally or in different CI stages, as it looks like the makefiles really want to recompile the verilated module (probably the file times are incorrect or something).

tilk:
> I think we need to migrate to [cocotb runners](https://docs.cocotb.org/en/stable/runner.html) if we want artifact that can be run locally or in different CI stages, as it looks like the makefiles really want to recompile the verilated module (probably the file times are incorrect or something).

Does this happen if you copy the entire build directory, or just the exec? If just the exec, then yes, the makefile will try to rebuild.

test/regression/cocotb.py
```diff
@@ -5,6 +5,13 @@
 from typing import Any
 from collections.abc import Coroutine
 from dataclasses import dataclass
+from pathlib import Path
+from filelock import FileLock
```
tilk, line 9: This is an external dependency, needs to be in pyproject.toml. Currently installed by accident as it's a dependency of `virtualenv`.

tilk (changes requested):

qbojj:
> Does this happen if you copy the entire build directory, or just the exec? If just the exec, then yes, the makefile will try to rebuild.

Also happens with whole directory

test/regression/cocotb.py
qbojj: Which one? `filelock` is explicitly in `dev` group, and the rest of the packages here are builtin

tilk: Huh, it actually is there. Nevermind.

tilk:
> > Does this happen if you copy the entire build directory, or just the exec? If just the exec, then yes, the makefile will try to rebuild.
> 
> Also happens with whole directory

Did you try `make -d` to look for the reason? Weird.

tilk (approved):

PR #1067, opened 2026-08-17
frontend: decouple checkpoint creation from instruction type

xThaid:
Checkpoints are currently created only for OpType.BRANCH, which works because fetch always stalls on JALR. This will no longer be the case once the branch predictor starts predicting JALR targets.

### DIFF
--- a/coreblocks/frontend/decoder/decode_stage.py
+++ b/coreblocks/frontend/decoder/decode_stage.py
@@ -35,7 +35,7 @@ class Decode(Elaboratable):
         m = TModule()
 
         @def_method(m, self.decode)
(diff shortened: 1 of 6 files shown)


qbojj (approved):
It may also be worth considering if we want to handle known unpredictable branches as unsafe.

xThaid:
I haven't done any measurements, but my intuition is that this wouldn't help much. Now that mispredictions are not much more expensive than stalling+resuming the core, I think making a static branch prediction (backwards branches taken) should be on average better. This is substantially different than JALRs, for which we can't do any kind of static prediction.

awariac:
can't i even get 24h for a requested review :( ?

tilk:
My bad, missed the request :(

PR #1068, opened 2026-08-18
fetch: cut the critical path in WideFifo

xThaid:
`validate_arguments` in `WideFifo.write` is a function of its arguments, so the entire fetch stage-2 datapath - predecode, prediction check, exit/unsafe resolution, aligner - feeds the grant for Fetch_Stage2, which then feeds readiness of Fetch_Stage1.

This is an attempt to cut this path by passing an upper bound (which itself is not on the critical path) of the actual count to validate_arguments.

Depends on https://github.com/kuznia-rdzeni/transactron/pull/229

### DIFF
--- a/coreblocks/frontend/fetch/fetch.py
+++ b/coreblocks/frontend/fetch/fetch.py
@@ -123,6 +123,7 @@ class FetchUnit(Elaboratable):
             depth=serializer_depth,
             read_width=self.gen_params.frontend_superscalarity,
             write_width=fetch_width,
+            write_max_count=True,
         )
 
         with Transaction(name="cont").body(m):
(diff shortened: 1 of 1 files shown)


coreblocks/frontend/fetch/fetch.py
```diff
@@ -538,7 +539,7 @@ def _(pc, ftq_ptr, fetch_gen):
 
                 # Make sure this is called only once to avoid a huge mux on arguments
                 m.d.av_comb += [aligner.valids.eq(fetch_mask), aligner.inputs.eq(raw_instrs)]
-                serializer.write(m, data=aligner.outputs, count=aligner.output_cnt)
+                serializer.write(m, data=aligner.outputs, count=aligner.output_cnt, max_count=popcount(instr_valid))
```
xThaid, line 556: The actual count is smaller than `popcount(instr_valid)` only in cases where we have a redirecting control flow instruction somewhere inside the fetch block.

pyproject.toml
```diff
@@ -20,8 +20,9 @@ dependencies = [
     "amaranth-stubs~=0.5.0.3",
     "amaranth-yosys==0.40.0.0.post100",
     "dataclasses-json==0.6.3",
-    "transactron~=0.12.0",
+    "transactron @ git+https://github.com/xThaid/transactron.git@d90d3a3264f2f4a7845c5a1f20b637d1e2f43d25",
```
tilk: Merge Transactron 0.13 first.

tilk (approved):

PR #1069, opened 2026-08-19
Fix konata wakeup deps

xThaid:
The operand wakeup event can arrive after its producer is already flushed (https://github.com/kuznia-rdzeni/coreblocks/actions/runs/32077746048/job/95534681965), so let's make sure we handle this in konata.py.

I also renamed `Update` event to `OperandWakeup`. Arguably, it was not the most descriptive name.


tilk (approved):

tilk:
Looks good, but I'll run benchmarks and look at konata files to be sure.

PR #1049, opened 2026-08-06
test: replace parameterized with pytest.mark.parametrize

slegarraga:
## Summary

Replaces `parameterized` and `parameterized_class` usage in the test suite with `pytest.mark.parametrize` and removes the `parameterized` dev dependency.

## Changes

- Converts class-level parameterization to pytest markers with autouse fixtures so setup state is preserved.
- Replaces `parameterized.expand` on FPU tests with pytest parametrize markers.
- Removes `parameterized==0.8.1` from `pyproject.toml`.

## Verification

- Formatting checks pass (`scripts/lint.sh check_format_flake8` and `check_format_black`).
- Pyright passes.
- 76 tests pass across the changed files: rvc, FPU, functional units, scheduler, prediction checker, and representative ICache parameterizations.
- `test_core.py` is collected successfully; full execution requires a RISC-V assembler that is not available in this environment.

Closes #1043

PR #1070, opened 2026-08-19
Replace parameterized with pytest.mark.parametrize

tilk:
As it looks like #1049 is abandoned, I'll massage it into something mergeable.

TODO:
- [x] Fix FU tests
- [ ] Restore test names where useful.


slegarraga:
Sorry again for going quiet on #1049 — that one is on me.

For the remaining TODO (restoring test names), I pushed a commit on top of your current tip (8b0d6c4b) that adds the pytest-native `ids=` argument everywhere the old `parameterized_class` names carried information:

```
git fetch https://github.com/slegarraga/coreblocks pr1070-test-names
git cherry-pick 213dda3e
```

It restores IDs like `TestCoreBasicAsm::test_asm_source[fibonacci_mem_tiny]`, `TestDivisionUnit::test_fu[ipc5]`, `TestSimpleCommonBusCacheRefiller::test[line32B_block8B_rv64i]`, `TestMultiplierUnit::test_fu[recursive_multiplier]`, `TestScheduler::test_randomized[Three-RS]`, etc. Verified with `pytest --collect-only` and a real sim run (`TestZbcUnit::test_fu[iterative]` passes). I left the FPU tests untouched since their original `parameterized.expand` cases never had meaningful names.

Happy to rework it however you prefer (different ID style, subset of files, or just ignore it).

PR #1073, opened 2026-08-22
more always_body

qbojj:
add missing always_body


coreblocks/scheduler/scheduler.py
```diff
@@ -435,10 +435,9 @@ def elaborate(self, platform):
 
                 # RS insertion guarantees RF response for present instructions
                 # Nested transaction used to avoid locking
-                with (tr := Transaction()).body(m):
+                with Transaction().always_body(m, ready=(i < instrs.count)):
```
tilk: On one hand, this looks nice. On second, this loses half of the assertion.

tilk (approved):

PR #1074, opened 2026-08-22
restore RVVI functionality

qbojj:
retiring ports stopped being contiguous thus breaking `order` calculations.

Additionally fixes following issues:
- `valid` signal not being set when `trap` was set
- `coreblocks --with-rvvi` not being operational (kuznia-rdzeni/transactron#238) - `ReadPort.data` was used directly in `log.info`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,7 +43,7 @@ jobs:
           TRANSACTRON_VERBOSE: 1
         run: |
           . venv/bin/activate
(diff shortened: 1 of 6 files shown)


tilk:
More CI checks are needed, so that this will not break again in the future.

tilk (approved):
Was a little hard to read because refactoring (e.g. moved port creation) was mixed with functionality changes.

PR #1075, opened 2026-08-23
ftq: count unpredicted JALRs as mispredictions

xThaid:
Currently, `HPMEvent.BRANCH_MISPREDICTION` is only an approximation of the number of mispredictions. Notably, it doesn't count unpredicted JALRs as mispredictions, while arguably they should be counted towards the total. It is important to tell the difference between a predictor that doesn't predict indirect targets and a predictor that predicts perfectly - the current implementation can't capture this.

I'm moving the counter to the FTQ entirely. It happens to have enough information to accurately track mispredictions. The only drawback is that the counter will be delayed by a few cycles compared to the previous approach.

### DIFF
--- a/coreblocks/backend/retirement.py
+++ b/coreblocks/backend/retirement.py
@@ -20,7 +20,7 @@ from transactron.lib.metrics import *
 from coreblocks.telemetry import RobFlush, RobRetire
 
 from coreblocks.params.genparams import GenParams
-from coreblocks.arch import ExceptionCause, HPMEvent, PrivilegeLevel
+from coreblocks.arch import ExceptionCause, PrivilegeLevel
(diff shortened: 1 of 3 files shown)


xThaid:
MPKI
```
Testbench name 	aha-mont64 	crc32 	minver 	nettle-sha256 	nsichneu 	slre 	statemate 	ud

this PR
Mispredicts/1k instr 	25.7208 	25.0839 	61.5207 	3.78449 	31.4815 	74.7678 	46.696 	23.5693

baseline
Mispredicts/1k instr 	25.0803 	0.0486121 	49.0654 	1.92665 	31.3985 	53.8958 	34.0989 	20.7867
```

qbojj (approved):

PR #1076, opened 2026-08-25
frontend: clean up unsafe stall logic

xThaid:
The stall controller changed a bit recently, and the logic for stalling the frontend on unsafe instructions now looks somewhat redundant to me. This is my attempt to clean it up.

Conceptually, once we hit an unsafe instruction we drop the next PC from the Fetch Address Unit, and the frontend stops producing new fetch blocks until it gets redirected. I believe exception stalls could move to the same approach (no next PC -> stall), but that would probably take more massaging.


tilk (approved):
Less code - better!

qbojj (approved):
~I think with this the jump branch unit can just unify the unpredicted and misspredicted paths (could not be true though)~

PR #1077, opened 2026-08-27
rvvi everywhere in testing for better debugging

qbojj:


qbojj:
regression tests didn't suffer, pysim tests got 20min more on this run.

EDIT: nvm, I just can't read

tilk (approved):
Sure, why not. As the performance impact seems negligible, making this switchable is probably an overkill.

PR #1078, opened 2026-08-27
More double-csrs

qbojj:
final (?) double-csr migration.

Fixed/added:
- double shadow being able to be specified without writeable shadows
- fixed `CSRRegister().write` and `CSRRegister()._fu_write` stripping upper bits (higher than XLEN) - this was effecting all counters, but other than that every other CSR only had (currently) constant bits in the upper 32 bits. Technically this was OK, as hpm counters are not required to have full 64 bit width
- fixed address calculation of CLINT fields for multi-hart systems


coreblocks/priv/csr/csr_instances.py
```diff
@@ -593,7 +598,7 @@ def elaborate(self, platform):
             m.d.sync += time_source.eq(time_source + 1)
 
         with Transaction().always_body(m):
-            self.time.write(m, data=time_source)
+            self.time.write(m, time_source)
```
tilk, line 601: Huh?

tilk (approved):

coreblocks/priv/csr/csr_instances.py
qbojj: I was getting desperate

PR #1079, opened 2026-08-27
implement Sstc

qbojj:
implements Sstc extension (supervisor counter compare)

It allows for better OS support (no M-mode in the hot-path of the OS). It also is nicely supported in the riscv-arch-test.

Contains a bonus fix for:
```asm
li x1, -1
csrs mip, x1
```
writing to the read only bits, because of the `read_comb`+`write` combo used to set edge triggered bits didn't account for `read_comb` not being masked by `ro_bits`

### DIFF
--- a/coreblocks/params/core_configuration.py
+++ b/coreblocks/params/core_configuration.py
@@ -117,6 +117,8 @@ class _CoreConfigurationDataClass:
         Enable User Mode.
     supervisor_mode: bool
         Enable Supervisor Mode.
+    sstc: bool
+        Enable Supervisor Mode Timer Counter (SSTC) extension. (only apllicable if supervisor_mode is enabled)
     asidlen: int
         Number of writable ASID bits in SATP.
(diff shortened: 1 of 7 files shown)


coreblocks/params/core_configuration.py
```diff
@@ -199,6 +201,9 @@ def __post_init__(self):
     asidlen: int | None = None
     supported_vm_schemes: Collection[SatpMode] = (SatpMode.BARE, SatpMode.SV32)
     phys_addr_bits: int | None = None
+
+    sstc: bool = True
+
```
tilk: Any reasoning for putting it here? I thought moving it just under `supervisor_mode` makes more sense.

qbojj: I was thinking about previous block as supervisor mode address space configuraiton, but now that I think about it it may be more general "address space configuration" due to `phys_addr_bits`

tilk (approved):

PR #1081, opened 2026-08-30
fix misa CSR value calculation

qbojj:
`misa` register had value 0x0, as `list(MisaExtension)` doesn't iterate names, but values

### DIFF
--- a/coreblocks/priv/csr/csr_instances.py
+++ b/coreblocks/priv/csr/csr_instances.py
@@ -413,7 +413,7 @@ class MachineModeCSRRegisters(Elaboratable):
         misa_ext = MisaExtension(0)
 
         for ext in gen_params.isa.extensions:
-            if ext.name in MisaExtension:
+            if ext.name in dir(MisaExtension):
                 misa_ext |= MisaExtension[ext.name]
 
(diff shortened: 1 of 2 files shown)


coreblocks/priv/csr/csr_instances.py
```diff
@@ -413,7 +413,7 @@ def _misa_value(self, gen_params):
         misa_ext = MisaExtension(0)
 
         for ext in gen_params.isa.extensions:
-            if ext.name in MisaExtension:
+            if ext.name in dir(MisaExtension):
```
tilk:
```suggestion
            if ext.name in MisaExtension.__members__:
```
Technically `dir` has more than just names. Unlikely that this would cause a problem here, but...

tilk:
How about adding a simple test in `csr_mmode` which would catch this? Just checking that `I` is set and some reserved bit is not set is enough.

qbojj:
> How about adding a simple test in csr_mmode which would catch this? Just checking that I is set and some reserved bit is not set is enough.

all bits of `mise` are checked (for set bits and writeablility) on the HEAD of riscv-arch-test. I could add something though

tilk (approved):

PR #1082, opened 2026-09-01
sim: a simulation backend written in C++

xThaid:
Adds a third backend for running programs on the core, next to pysim and cocotb.

cocotb crosses into Python on every clock edge and every bus transfer, even though a whole benchmark only touches MMIO a handful of time. This backend keeps the clock loop, the Wishbone buses and the RAM in C++ next to the verilated core, and calls into Python only for MMIO.

It doesn't support traces, logs, profiles or event logs for now - cocotb remains the backend for those.

CI now runs the regression and arch tests on it (in addition to a benchmark run under it too).


.github/workflows/main.yml
```diff
@@ -262,15 +266,118 @@ jobs:
...
+    container: ghcr.io/kuznia-rdzeni/riscv-toolchain:2024.12.07
+    timeout-minutes: 20
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          submodules: recursive
+
+      - name: Build embench
+        run: cd test/external/embench && make
```
qbojj:
```suggestion
        working-directory: test/external/embench
        run: make
```

scripts/run_benchmarks.py
```diff
@@ -116,7 +121,19 @@ def make_backend(benchmark_name: str) -> SimulationBackend:
     return run_benchmarks_with_backend(benchmarks, make_backend, jobs)
 
 
-def run_benchmarks(benchmarks: list[str], backend: Literal["pysim", "cocotb"], traces: bool, jobs: int) -> bool:
+def run_benchmarks_with_cxxsim(benchmarks: list[str], traces: bool, jobs: int) -> bool:
+    if traces:
+        print("The cxxsim backend does not support traces")
+        sys.exit(1)
```
qbojj:
```suggestion
    if traces:
        raise RuntimeError("The cxxsim backend does not support traces")
```

test/sim/cxx_build.py
```diff
@@ -0,0 +1,183 @@
...
+    "-Wno-CASEOVERLAP",
+    "-Wno-WIDTHEXPAND",
+    "-Wno-WIDTHTRUNC",
+    "-Wno-UNSIGNED",
+    "-Wno-CMPCONST",
+    "-Wno-LITENDIAN",
+    "-Wno-UNOPTFLAT",
+]
+
+OPTIMIZATION_FLAGS = ["-O3", "-march=native", "-DNDEBUG", "-std=c++17"]
```
qbojj: 1. `-std=c++17` is not an optimization flag. Also we probably shouldn't be passing this to verilator
2. maybe skip `-march=native`, to be able to use the artifacts on different machines (also it is not reproducible). If this will impact the performance too much we can consider `-march=x86-64-v3`, as probably everyone has that.
3. `-DNDEBUG` is not really needed, as verilator doesn't use `assert`, it has it's own debugging solution that is guarded by `VL_DEBUG`
4. `-O3` - verilator has special `OPT_FAST` to only optimize the critical code path (`-MAKEFLAGS OPT_FAST="-O3"`). Also probably we should check if `-Os` (or `-Os -fstrict-aliasing`) doesn't give better results, as verilator tends to have problems with cache busting. (the default for `OPT_FAST` is `-Os` either way)

test/sim/cxx_build.py
```diff
@@ -0,0 +1,183 @@
...
+        "--top-module",
+        "top",
+        "--prefix",
+        "Vtop",
+        "-Mdir",
+        str(BUILD_DIR),
+        "--build",
+        "-j",
+        "0",
+        "-O3",
```
qbojj, line 83: maybe also consider `–x-initial fast`

test/sim/cxx/memory.h
```diff
@@ -0,0 +1,96 @@
...
+
+class MemorySegment {
+  public:
+    MemorySegment(uint64_t start, uint64_t end, uint32_t flags) : start_(start), end_(end), flags_(flags) {}
+    virtual ~MemorySegment() = default;
+
+    bool contains(uint64_t addr) const { return addr >= start_ && addr < end_; }
+
+    uint64_t start() const { return start_; }
+    uint64_t end() const { return end_; }
```
qbojj: nit: `end` has special meaning in C++, maybe something else?

test/sim/cxx/module.cpp
```diff
@@ -0,0 +1,38 @@
...
+        .value("EXECUTABLE", SEGMENT_EXECUTABLE);
+
+    py::enum_<ReplyStatus>(m, "ReplyStatus")
+        .value("OK", ReplyStatus::Ok)
+        .value("ERROR", ReplyStatus::Error)
+        .value("RETRY", ReplyStatus::Retry);
+
+    py::enum_<FinishReason>(m, "FinishReason")
+        .value("STOPPED", FinishReason::Stopped)
+        .value("TIMEOUT", FinishReason::Timeout);
```
qbojj: prefer `py::native_enum_` over `py::enum_`

test/sim/cxx/memory.h
```diff
@@ -0,0 +1,96 @@
...
+    uint32_t flags() const { return flags_; }
+    bool has_flag(SegmentFlags flag) const { return (flags_ & flag) != 0; }
+
+    virtual ReadResult read(uint32_t addr, uint8_t byte_count, uint8_t byte_sel, bool exec) = 0;
+    virtual ReplyStatus write(uint32_t addr, uint32_t data, uint8_t byte_count, uint8_t byte_sel) = 0;
```
qbojj: addr is `uint32_t` while start and end are `uint64_t`. please choose the address space width (best to just typedef it)

qbojj (changes requested):

tilk:
Damn, that's a lot of new code to maintain. The sim performance advantage is obvious, but the non-cooperation with logging etc. is a downside, and fixing this will probably add even more code. Will merge if others agree, but would prefer something lighter.

test/sim/cxx/simulation.cpp
```diff
@@ -0,0 +1,92 @@
...
+void Simulation::add_ram(uint64_t start, uint64_t end, uint32_t flags, const py::bytes& initial_data) {
+    const std::string_view contents = std::string_view(initial_data);
+    memory_.add_segment(std::make_unique<RamSegment>(
+        start, end, flags, std::vector<uint8_t>(contents.begin(), contents.end())));
+}
+
+void Simulation::add_mmio(uint64_t start, uint64_t end, uint32_t flags, py::object on_read, py::object on_write) {
+    memory_.add_segment(
+        std::make_unique<CallbackSegment>(start, end, flags, std::move(on_read), std::move(on_write)));
+}
```
qbojj: prefer explicit typing with a typedef. eg:
```cpp
// in memory.h
using read_callback_t = std::function<ReadResult(uint32_t,uint8_t,uint8_t,bool)>;

// in CallbackSegment
read_callback_t on_read_;

// here
void Simulation::add_mmio(uint64_t start, uint64_t end, uint32_t flags, read_callback_t on_read, write_callback_t on_write) {
    memory_.add_segment(
        std::make_unique<CallbackSegment>(start, end, flags, std::move(on_read), std::move(on_write)));
}
```

This will also automatically acquire GIL when needed and will fully remove pybind11 from memory.cpp and memory.h. From what I remember it also handles convertions to/from tuple likes (aka `ReadResponse`)

test/sim/cxx/simulation.cpp
```diff
@@ -0,0 +1,92 @@
...
+        instr_bus.on_falling_edge();
+        data_bus.on_falling_edge();
+        top.interrupts = interrupts_.load(std::memory_order_relaxed);
+
+        top.clk = 0;
+        top.eval();
+    };
+
+    // The callbacks take the GIL back for the duration of a call.
+    py::gil_scoped_release gil;
```
qbojj: GIL can be released even before this function. maybe add `py::call_guard<py::gil_scoped_release>()` on this method in module.cpp

test/sim/cxx/simulation.cpp
```diff
@@ -0,0 +1,92 @@
...
+void Simulation::add_mmio(uint64_t start, uint64_t end, uint32_t flags, py::object on_read, py::object on_write) {
+    memory_.add_segment(
+        std::make_unique<CallbackSegment>(start, end, flags, std::move(on_read), std::move(on_write)));
+}
+
+RunResult Simulation::run() {
+    if (has_run_) {
+        throw std::runtime_error("A simulation can only be run once");
+    }
+    has_run_ = true;
```
qbojj: I don't see why that would need to be the case, but since all other places here are already made to be thread safe, maybe make `has_run_` atomic and do:
```suggestion
    if (has_run_.exchange(true)) {
        throw std::runtime_error("A simulation can only be run once");
    }
```

test/sim/cxx/simulation.cpp
```diff
@@ -0,0 +1,92 @@
...
+    : context_(std::make_unique<VerilatedContext>()),
+      memory_(fail_on_undefined_read, fail_on_undefined_write),
+      timeout_cycles_(timeout_cycles) {
+    context_->traceEverOn(false);
+    top_ = std::make_unique<Vtop>(context_.get(), "top");
+}
+
+Simulation::~Simulation() = default;
+
+void Simulation::add_ram(uint64_t start, uint64_t end, uint32_t flags, const py::bytes& initial_data) {
```
qbojj: I think just making `initial_data` a `std::string_view` would work too

test/sim/cxx/module.cpp
```diff
@@ -0,0 +1,38 @@
+#include <pybind11/pybind11.h>
+
+#include "memory.h"
+#include "simulation.h"
+
+namespace py = pybind11;
+
+PYBIND11_MODULE(coreblocks_cxxsim, m) {
```
qbojj:
```suggestion
PYBIND11_MODULE(coreblocks_cxxsim, m, py::mod_gil_not_used()) {
```

or even

```suggestion
PYBIND11_MODULE(coreblocks_cxxsim, m, py::mod_gil_not_used(), py::multiple_interpreters::per_interpreter_gil()) {
```

test/sim/cxx_build.py
```diff
@@ -0,0 +1,183 @@
...
+    """Links the verilated core and the simulator into a Python extension module."""
+    try:
+        import pybind11
+    except ImportError:
+        raise RuntimeError("Building the cxxsim backend requires pybind11")
+
+    verilator_root = _verilator_root()
+    includes = {sysconfig.get_paths()["include"], sysconfig.get_paths()["platinclude"], pybind11.get_include()}
+
+    command = ["g++", "-shared", "-fPIC", "-fvisibility=hidden"] + OPTIMIZATION_FLAGS
```
qbojj: maybe add some warning options for sanity?

qbojj (changes requested):

qbojj:
> the non-cooperation with logging etc.

I think I can make this work with logging and evlog with some codegen. I think if we get to the point where those work we could just remove cocotb support

tilk:
> I think I can make this work with logging and evlog with some codegen. I think if we get to the point where those work we could just remove cocotb support

The main reason cocotb was selected was that it was relatively simple to implement, and low maintenance. We can be reasonably sure cocotb won't break randomly while updating some dependency, and even if it does, it's cocotb's problem. I agree that the performance is disappointing, though.

I'm not completely against an in-house solution, but it would be nice if someone was committed to fixing problems if something breaks.

qbojj:
> We can be reasonably sure cocotb won't break randomly while updating some dependency, and even if it does, it's cocotb's problem.

that would be true if we only used plain cocotb, but we also use cocotb-bus which is no longer maintaned

test/sim/cxx/memory.h
xThaid: I presume you're referring to conventional terminology in the standard library, but I think I'll keep `end` - I can't find any good alternatives

test/sim/cxx_build.py
xThaid: nice! seems to be around 2x faster now on my laptop

xThaid: nevermind, on github seems to be the same

test/sim/cxx_build.py
qbojj: build still spits -Wno-sign-compare warnings.

Verilator's makefile adds `-Wno-bool-operation -Wno-int-in-bool-context -Wno-shadow -Wno-sign-compare -Wno-subobject-linkage -Wno-tautological-compare -Wno-uninitialized -Wno-unused-but-set-parameter -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-unused-variable` by default.

Also I'm not sure if I like that we are pretending to be a build system - maybe just use verilator's makefiles.

test/sim/cxx/memory.h
qbojj: that's ok. For the record I was referring to the iterator api (what allows for range based `for` loops).

test/sim/cxx_build.py
```diff
@@ -0,0 +1,206 @@
...
+    command += [
+        f"-I{BUILD_DIR}",
+        "-isystem",
+        f"{verilator_root}/include",
+        "-isystem",
+        f"{verilator_root}/include/vltstd",
+    ]
+    command += [str(CXX_ROOT / source) for source in SOURCES]
+    # The Verilator runtime, which is not part of the generated library.
+    command += [str(verilator_root / "include" / name) for name in ["verilated.cpp", "verilated_threads.cpp"]]
```
qbojj: verilator creates `libverilated.a` in the verilated dir. Link it instead of compiling those files again (it will also help with warnings)

qbojj: I think this would be enough (with -Wno-* removed from `MODULE_CXX_FLAGS`)
```py
    command = ["g++", "-shared"] + MODULE_CXX_FLAGS
    command += [f"-I{include}" for include in sorted(includes)]
    command += [
        f"-I{BUILD_DIR}",
        "-isystem",
        f"{verilator_root}/include",
        "-isystem",
        f"{verilator_root}/include/vltstd",
    ]
    command += [str(CXX_ROOT / source) for source in SOURCES]
    command += [f"-L{BUILD_DIR}", "-lVtop", "-lverilated"]
    command += ["-o", str(MODULE_PATH)]
```

xThaid: nice, I didn't have these two libs, but then I just realized that I'm running 5.008 version, which is actually the same as on our github workflows currently (updating this as well).

qbojj:
I think right now we lost the transactron errors and assertions in CI, as those are not real verilog assertions, but soft assertions checked by the logger.

qbojj:
It also made the build location aware (it will rebuild when e.g copying between computers), as the full path is used in the stamp calculation

xThaid:
> I think right now we lost the transactron errors and assertions in CI, as those are not real verilog assertions, but soft assertions checked by the logger.

I re-added regression tests with the cocotb backend to the CI workflow.

> It also made the build location aware (it will rebuild when e.g copying between computers), as the full path is used in the stamp calculation

I actually removed all the fingerprinting logic. I realized that the behavior was inconsistent and a bit misleading. We can’t detect changes to the Python HDL, so after making a change, we would have to run the build with the --clean flag anyway.

qbojj (approved):
Let's merge this for now. And after checking how it compares to new cocotb (possibly with verilator control files or just wait for lazy vpi public to be merged), than we will decide which path should stay

.github/workflows/main.yml
```diff
@@ -262,15 +266,113 @@ jobs:
...
+          . venv/bin/activate
+          python3 -m test.sim.cxx_build
+
+      - name: Upload the simulator
+        uses: actions/upload-artifact@v7
+        with:
+          name: cxxsim
+          path: test/sim/build/cxxsim/*.so
+
+  build-embench:
```
tilk, line 317: Embench is now built twice, in `main.yml` and `benchmark.yml`. Why?

xThaid: Mainly because of simplicity - it’s quite annoying to share artifacts between different workflows. We could technically use github caches, but then we’d have to introduce an ordering between the workflows. It takes an epsilon to build the embench binaries, so I thought I would just build it twice

tilk:
So let's merge it, but I'd like to hear more from @qbojj on the faster cocotb front.

PR #1083, opened 2026-09-02
regression: separate memory segment descriptions from their emulation

xThaid:
Separate memory segment descriptions from their Python emulation, allowing non-Python backends to consume the memory map without unused simulation state.

### DIFF
--- a/test/regression/memory.py
+++ b/test/regression/memory.py
@@ -1,10 +1,9 @@
 from abc import ABC, abstractmethod
-from collections.abc import Callable
 from enum import Enum, IntFlag, auto
-from typing import Optional, TypeVar
-from dataclasses import dataclass, replace
+from dataclasses import dataclass
 from elftools.elf.constants import P_FLAGS
 from elftools.elf.elffile import ELFFile, Segment
+
 from coreblocks.params.configurations import CoreConfiguration
 from transactron.utils import align_to_power_of_two, align_down_to_power_of_two
 
@@ -15,13 +14,10 @@ __all__ = [
     "ReadReply",
     "WriteRequest",
     "WriteReply",
-    "MemoryEmulation",
     "MemorySegment",
     "RandomAccessMemory",
-    "RandomAccessMemoryEmulation",
     "MMIOSegment",
     "CoreMemoryModel",
-    "CoreMemoryEmulation",
     "load_segment",
     "load_segments_from_elf",
 ]
@@ -66,21 +62,6 @@ class WriteReply:
     status: ReplyStatus = ReplyStatus.OK
 
 
-class MemoryEmulation(ABC):
-    """Emulates the behaviour of a memory segment in Python.
-
-    Request addresses are relative to the start of the segment.
-    """
-
-    @abstractmethod
-    def read(self, req: ReadRequest) -> ReadReply:
-        raise NotImplementedError
-
-    @abstractmethod
-    def write(self, req: WriteRequest) -> WriteReply:
-        raise NotImplementedError
-
-
 @dataclass
 class MemorySegment:
     """Describes a range of the address space."""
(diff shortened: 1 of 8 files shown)


test/regression/memory.py
```diff
@@ -73,17 +81,36 @@ def write(self, req: WriteRequest) -> WriteReply:
...
+        if len(self.contents) != len(self.address_range):
             raise ValueError("Data length must be equal to the length of the address range")
 
+
+class MMIOSegment(MemorySegment, MemoryEmulation):
+    """A segment whose accesses are handled by Python code."""
+
+
+class RandomAccessMemoryEmulation(MemoryEmulation):
+    """Byte-addressable storage."""
```
tilk: What is the rationale behind this design? Sometimes things are both a `Segment` and an `Emulation`, sometimes just one of them, it's seriously weird.

test/regression/memory.py
```diff
@@ -156,6 +205,9 @@ def load_segment(
     seg_start = paddr
     seg_end = paddr + memsz
 
+    # The bus is word-addressed, so a segment cannot end in the middle of a word.
+    seg_end = align_to_power_of_two(seg_end, 2)
```
qbojj: probably should assert rather than quietly increase the size. Is it even a problem in any of the current tests?
`seg_start`'s alignment should also be checked if we are pedantic.

qbojj (approved):

test/regression/memory.py
qbojj: also looking at the final result, you do handle the partially out of bounds reads/writes, so there is no need to do this here (no need for this or assert)

xThaid: yeah, `rv32ui-sb` happens to have a write that crosses a segment boundary.
```
Section Headers:
  [Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            00000000 000000 000000 00      0   0  0
  [ 1] .text.init        PROGBITS        80000000 001000 00063c 00  AX  0   0 64
  [ 2] .data             PROGBITS        90000000 002000 00000a 00  WA  0   0 16
  [ 3] .riscv.attributes RISCV_ATTRIBUTE 00000000 00200a 000033 00      0   0  1
  [ 4] .symtab           SYMTAB          00000000 002040 0002f0 10      5  45  4
  [ 5] .strtab           STRTAB          00000000 002330 000166 00      0   0  1
  [ 6] .shstrtab         STRTAB          00000000 002496 00003e 00      0   0  1
```

.data is 10 bytes long and the test tries to write to 0x90000008
```
RuntimeError: Read crosses a segment boundary: 90000008
```
which is technically fine because byte_sel is set accordingly, but I thought that it would be saner if we instead ensured that segments cannot end in the middle of a machine word.

the python memory emulator silently ignores the out-of-bounds write:
```
>>> foobar = bytearray(b"hello")
>>> foobar[4:10] = b", world"
>>> foobar
bytearray(b'hell, world')
```

test/regression/memory.py
xThaid: My idea behind this is:
MemorySegment - the backend-agnostic representation of the memory map.
MemoryEmulation - a python emulator of a memory segment.

This distinction is quite simple for plain RAM-like memory: the abstraction describes the address range, permissions, and initial contents, while the python emulator realizes it by maintaining the actual storage.

MMIOSegment is, by definition, a memory segment that handles requests in Python (regardless of the backend), so its backend-agnostic abstract definition contains read/write methods. Coincidentally or not, its python emulation follows exactly the same protocol (read and write methods).

Let me refactor this to make the distinction between backend-agnostic definitions and emulations clearer...

test/regression/memory.py
qbojj: I think if this is the route we take, a full memory map checker is needed to assert we don't create aliased memory regions

tilk: If the problem is just riscv-tests, I think a better idea than doing this kind of a correction (which might lead to overlapping segments, as noticed) would be to fix this in the linker script.

PR #1084, opened 2026-09-02
 regression: one core configuration  & benchmark improvements

xThaid:
- use a single core configuration across all simulation backends
- benchmarks now run in parallel, report simulation speed, and can rebuild generated artifacts with --clean

Depends on #1083

### DIFF
--- a/scripts/run_benchmarks.py
+++ b/scripts/run_benchmarks.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 
-import unittest
 import asyncio
 import argparse
 import json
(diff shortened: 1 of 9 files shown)


scripts/run_benchmarks.py
```diff
@@ -142,6 +180,18 @@ def main():
     parser.add_argument("-p", "--profile", action="store_true", help="Write execution profiles")
     parser.add_argument("--evlog", action="store_true", help="Write captured event logs")
     parser.add_argument("-b", "--backend", default="cocotb", choices=["cocotb", "pysim"], help="Simulation backend")
+    parser.add_argument(
+        "--clean",
+        action="store_true",
+        help="Discard the generated Verilog and the built testbench",
+    )
+    parser.add_argument(
```
qbojj, line 188: many tools taking `-j` argument also accept `0` as a auto thread count flag. Maybe we should do the same

qbojj (approved):

PR #1087, opened 2026-09-03
fix mip for real this time

qbojj:
the `csrs` and `csrc` instructions behaved like they were writing to all writable bits - not only those in the mask


tilk:
Clearly, a test is needed which:
- Fails with buggy version,
- Succeeds with corrected version,
- Is clearly testing for spec-compliant behavior.

Edit: is this still #938, should it be reopened?

qbojj:
It's unrelated to 938

qbojj:
OK, I think I cannot read... again.
This is literally equivalent and just misread everything :sob:

Issue #805, opened 2025-03-06
Add benchmark cycle run time to graphs


piotro888:
IPC is not always the best measurement. Ex. Toolchain updates may replace instructions for shorter/longer ones that execute faster but executes less instructions.
 
This data can be overlaid on the same graphs as IPC

PR #1090, opened 2026-09-06
Align regression ELF segments to 64B

xThaid:
It seems reasonable to assume that ELF segment boundaries should be aligned to the cache line size. This way, we can remove the ad hoc alignment adjustments used to work around icache issues.

### DIFF
--- a/test/external/riscv-tests/environment/custom/link.ld
+++ b/test/external/riscv-tests/environment/custom/link.ld
@@ -7,20 +7,11 @@ MEMORY
     data (wxa!ri) : ORIGIN = 0x90000000, LENGTH = 128K
 }
 
-PHDRS
-{
-  text PT_LOAD;
-  data_init PT_LOAD;
-  data PT_NULL;
-}
-
 SECTIONS
 {
-  .text.init : { *(.text.init) } >text AT>text :text
-  . = ALIGN(0x1000);
-  .text : { *(.text) } >text AT>text :text
-  . = ALIGN(0x1000);
-  .data : { *(.data) } >data AT>data :data_init
-  .bss : { *(.bss) } >data AT>data :data
+  .text.init : { *(.text.init) } >text
+  .text : { *(.text .text.*) . = ALIGN(0x40); } >text
+  .data : { *(.data .data.*) } >data
+  .bss : { *(.bss .bss.*) . = ALIGN(0x40); } >data
   _end = .;
 }
--- a/test/external/embench/common/link.ld
+++ b/test/external/embench/common/link.ld
@@ -17,10 +17,12 @@ SECTIONS
 
   .text : {
     *(.text .text.*)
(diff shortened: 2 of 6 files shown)


tilk:
This somehow improved performance on one of the tests, huh?

tilk:
The `statemate` jump is kind of funny. Because of the memory layout change, code became slightly larger - more `auipc` instructions were preserved by the linker. But at the same time, the code size change didn't impact the performance that much, probably thanks to superscalarity. End result? Higher IPC. More instructions, only slightly larger number of cycles.

tilk:
Still, everything works, so I'd consider this ready for review.

tilk:
I think that there is a little too much alignment added to the scripts. Sections don't really need the extra alignment; segments do.

Edit: Separation between `.rodata` and `.data` is fine because of different permissions. Adding alignment between `.data` and `.bss` looks superfluous.

awariac:
> The `statemate` jump is kind of funny. Because of the memory layout change, code became slightly larger - more `auipc` instructions were preserved by the linker. But at the same time, the code size change didn't impact the performance that much, probably thanks to superscalarity. End result? Higher IPC. More instructions, only slightly larger number of cycles.

That's why I opened an issue long time ago to overlay a plot with total number of cycles on IPC one

there was a PR open for it, what did happen with it?

tilk:
> That's why I opened an issue long time ago to overlay a plot with total number of cycles on IPC one
> 
> there was a PR open for it, what did happen with it?

Did you mean #805?

Anyway, this would not do anything useful here - the plots are for already committed PRs.

awariac:
#875 
yes, I'm only saying it's a useful data to plot

test/external/riscv-tests/environment/custom/link.ld
```diff
@@ -7,20 +7,11 @@ MEMORY
     data (wxa!ri) : ORIGIN = 0x90000000, LENGTH = 128K
 }
 
-PHDRS
```
xThaid, line 10: these were in fact unnecessary and made .bss land in a separate segment.

new layout:
```
Program Headers:
  Type           Offset   VirtAddr   PhysAddr   FileSiz MemSiz  Flg Align
  LOPROC+0x3     0x00200a 0x00000000 0x00000000 0x00033 0x00000 R   0x1
  LOAD           0x001000 0x80000000 0x80000000 0x00640 0x00640 R E 0x1000
  LOAD           0x002000 0x90000000 0x90000000 0x0000a 0x00040 RW  0x1000

 Section to Segment mapping:
  Segment Sections...
   00     .riscv.attributes
   01     .text.init .text
   02     .data .bss
```

tilk: Thanks for cleaning this up.

qbojj (approved):

PR #1096, opened 2026-09-08
arch: narrow CfiType to the CFI type alone & add return-address-stack action

xThaid:
`CfiType` currently packs two orthogonal facts into one enum: what kind of CFI an instruction is, and whether it acts on the return address stack. This PR narrows `CfiType` to the four real CFI types and moves the stack effect into a separate `RasAction` field.

### DIFF
--- a/coreblocks/arch/optypes.py
+++ b/coreblocks/arch/optypes.py
@@ -9,6 +9,7 @@ __all__ = [
     "OpType",
     "impure_optypes",
     "CfiType",
+    "RasAction",
 ]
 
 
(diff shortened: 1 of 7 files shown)


coreblocks/arch/optypes.py
```diff
@@ -66,43 +67,35 @@ class OpType(IntEnum):
...
-    RET = 0b110  # Return from a function (JALR with rs1 equal to x1 or x5)
 
-    JAL = 0b011  # Jump and Link
-    CALL = 0b111  # Call a function (JAL with rd equal to x1 or x5))
-
-    @staticmethod
-    def valid(val: ValueLike) -> Value:
-        return Value.cast(val)[0:2] != CfiType.INVALID
+@unique
+class RasAction(Enum, shape=2):
```
qbojj: maybe just a `Flag` with `PUSH` and `POP`?

qbojj (approved):

PR #1097, opened 2026-09-08
frontend/bpu: add the return address stack

xThaid:
This PR adds a return address stack (RAS) to predict return targets during fetch, with checkpoints for recovery after redirects. We only restore the top entry, so speculative execution can still corrupt deeper entries in the stack.

Two follow-ups are planned: make the RAS resilient to these corruptions and move it into the BPU.

Depends on #1096


coreblocks/frontend/bpu/ras.py
```diff
@@ -0,0 +1,94 @@
...
+        self.peek = Method(o=self.peek_layout)
+        self.update = Method(i=self.update_layout, o=self.state_layout)
+        self.recover = Method(i=self.state_layout)
+
+    def elaborate(self, platform):
+        m = TModule()
+
+        xlen = self.gen_params.isa.xlen
+
+        stack = Array(Signal(xlen, name=f"ras_entry_{i}") for i in range(self.depth))
```
tilk: I believe a synchronous RAM could be easily used, this would reduce resource usage.

tilk:
Probably soon it's time to move the `full` config out of ECP5 (and maybe add some intermediate config), but for this PR I believe the resource usage can be optimized.

tilk (approved):

qbojj (approved):

PR #1100, opened 2026-09-11
Verilator control file generation

qbojj:
create verilator control file generator.

Further steps - either:
- implement logging/evlog in cxxsim
- remove `--public-flat-rw` from the cocotb makefile and add this file to the build (this is very simple if we migrated to runners - just add a new `Runner` based on `class Verilator(Runner)`

### DIFF
--- a/test/sim/verilog.py
+++ b/test/sim/verilog.py
@@ -1,5 +1,6 @@
 import shutil
 from pathlib import Path
+import json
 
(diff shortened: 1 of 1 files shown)


awariac:
ehh, another case when patch based review system would be useful

tilk (approved):
Untested in the PR, but will get used soon.

PR #1101, opened 2026-09-12
migrate to cocotb 2.1

qbojj:
1. update cocotb to 2.1
1. remove cocotb_bus
1. fix `sel`-less bus handling in cocotb
1. since cocotb 2.0 clock is driven natively without interaction with python
1. for now still makes all signals public


tilk:
What's missing for this PR to be ready for review?

tilk (approved):
Cocotb-bus dependency removed, fresh cocotb - nice.

PR #1105, opened 2026-09-14
frontend/bpu: introduce a uniform predictor component protocol

xThaid:
This sets things up for adding CFI and direction predictors and passing their metadata back for training.

I tried to find a balance between a fully customizable pipeline and a fixed setup, keeping latency and circuit size in mind. Letting predictors be swapped out while keeping their pipeline stages fixed feels like a good tradeoff.

The plan is to split the BPU into three roles:

- Fast predictor: gives an early CFI and target prediction in S1 to keep fetching moving.
- CFI predictor: identifies possible branches and jumps in S1, then provides their targets in S2.
- Direction predictor: uses those hints to predict which branches are taken in S2.


coreblocks/frontend/bpu/component.py
```diff
@@ -0,0 +1,138 @@
...
+    "FastPredictor",
+    "CfiPredictor",
+    "DirectionPredictor",
+]
+
+
+class BPUComponentLayouts:
+    """Shared predictor layouts."""
+
+    def __init__(self, gen_params: GenParams):
```
qbojj: `meta_width` should probably be an argument here.
Also `fetch_width` is in `gen_params` - no need to pass it to `direction_prediction`.

As for `candidate_count` idk. Either leave it as is, or just create a second class, as fast and normal BPU components have different interfaces either way

qbojj (changes requested):

qbojj (approved):

PR #1107, opened 2026-09-15
Increase synthesis benchmark time limit

tilk:
The full-xc7k480t benchmark is flaky, probably because of synthesis time. Timeout increase hopefully should fix the problem.


qbojj (approved):

awariac:
we should also lower synthesis frequency target if applicable

tilk:
I did try to change the frequency target when working on the CI workflow, and after a few tries concluded this changes exactly nothing.

PR #1108, opened 2026-09-15
embench: add build-fast and run more benchmarks by default

xThaid:
The only reason we run a subset of embench is that we currently have `cpu_mhz = 0.01`, which makes the repeat count (LOCAL_SCALE_FACTOR * CPU_MHZ, cut down to a whole number) zero for most benchmarks. Now that we have cxxsim, which runs significantly faster, we can afford `cpu_mhz = 0.1` by default and run 16 benchmarks (technically we could run 17, but `huffbench` fails and I didn't debug it yet). 

The old setup stays as `make build-fast` / `--fast`, and benchmark.yml uses it.


tilk (approved):
Benchmarks now are run in two ways in CI, but that's justified given the current state of the simulators. Clean up left for later.

tilk:
Also, it's pretty cool we now have a benchmark where IPC>1.

PR #1109, opened 2026-09-16
Retirement names refactor

tilk:
This PR replaces arcane bit tricks with function calls. Also, a bug was found, where a mask was truncated because of wrong signal size.

The fix even improved IPC - probably effects are fired earlier.


coreblocks/backend/retirement.py
```diff
@@ -427,7 +436,9 @@ def _():
             current_tag_expr += entry.rob_data.tag_increment
             current_tag = Signal(self.gen_params.tag_bits)
             m.d.comb += current_tag.eq(current_tag_expr)
-            m.d.comb += pure_inactive_offset[i].eq(~active_tags[current_tag] & (~(impure_mask | -impure_mask))[i])
+            m.d.comb += pure_inactive_offset[i].eq(
+                ~active_tags[current_tag] & mask_before_first_set_bit(impure_mask)[i]
```
qbojj: we should separate `mask_before_first_set_bit` to its own assignment

qbojj (approved):

PR #1110, opened 2026-09-16
make cocotb fast

qbojj:
- fix verilator control files
- remove all signals being public
- query handles only at startup


qbojj:
Tests from clean build (need to build verilog first)
cocotb - `375,38s user 8,89s system 203% cpu 3:08,95 total`
cxxsim - `317,92s user 10,25s system 181% cpu 3:00,82 total`

Performance:
1. riscv-arch-tests ExceptionsSm-00.elf - 36167 cycles
    - cocotb - 4,96s user 0,13s system 99% cpu 5,116 total (9146.23 cycles/s + run overhead)
    - cxxsim - 0,82s user 0,06s system 99% cpu 0,886 total

1. riscv-arch-tests I-add-00.elf - 13532 cycles
    - cocotb - 2,90s user 0,12s system 99% cpu 3,028 total (7128.86 cycles/s + run overhead)
    - cxxsim - 0,54s user 0,05s system 100% cpu 0,588 total

Please remember that those do not have feature parity. with that some observations
1. There is about 1s overhead on running cocotb.

tilk:
I think that extending the set of benchmarks could be done in another PR.

test/sim/cocotb.py
```diff
@@ -175,19 +175,27 @@ def get_cocotb_handle(self, path_components: list[str]) -> LogicObject:
         return obj
 
     async def profile_handler(self, clock: LogicObject, profile: Profile):
```
tilk, line 177: What is the reason for these changes?

qbojj: 1. It errors out at the initialisation stage rather than when it tries to log a signal that doesn't exist (helped debugging)
1. Querying VPI is quite slow - we should hold onto the handles

tilk (approved):

